PR #31, opened 2016-07-03
Fix the link to SUMMARY.md

Javix:
Use the relative path to be able to navigate to `parent` chapters


Javix:
It seems like `master`has already this implemented (strange,  how could I miss that, it didn't work properly on my Mac ...

PR #39, opened 2016-07-13
Cucumberdocs modified as per suggestions

gmk1217:
As per Aslak suggestions, I modified the documents.


aslakhellesoy:
We can't use copyrighted images, ref c1a73792fe0e06cb3b693631ef0f0afa64c70c3e

gmk1217:
Cant we use PC below the image...or at the end of the page

aslakhellesoy:
Please don't group several document changes into the same PR. See 783928efa64e52e64cd0fa218a4e86b6b91eeb6a and c4a842c83fa0b3216ce010d6507e17a629d52a19.

Once you have created a pull request you can continue to push new commits to the pull request's branch whenever you're addressing feedback comments. Don't create a new PR to addres them.

In this case I suggest you start over with new branches and pull requests since it's a little messy now.

    git checkout master
    git checkout -b docs-selenium
    git add [selenium related docs]
    git commit
    git push
    [create a PR]

Repeat the process for the other (unrelated docs).

Whenever you need to fix comments from us, update your existing PR:

    git checkout docs-selenium
    git add [the files you changed]
    git commit
    git push

gmk1217:
My bad, i was following the wrong way.
Thnx, will follow these steps

aslakhellesoy:
> Cant we use PC below the image...or at the end of the page

Sorry, what is "PC"?

All creations are copyright by default, even without an explicit © sign, unless it's made explicit in writing that copying is allowed. Copying by anyone else than the copyright holder is illegal. See http://www.templetons.com/brad/copymyths.html

We should avoid using 3rd party images unless they have an explicit license (like creative commons) that permits copying. 

Even when it's allowed we should avoid it since we want to have a consistent look and feel for the docs.

gmk1217:
PC: picture courtesy

aslakhellesoy:
No, we'd still be infringing copyright (unless the pic has a license permitting copying).

gmk1217:
ok will delete those

PR #34, opened 2016-07-06
Add tags

belgoros:
Adds a tags topic.

PR #42, opened 2016-07-14
Add step definitions

belgoros:
Adds step definitions sections extracted as suggested in https://github.com/cucumber/cucumber/pull/34

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -66,7 +66,7 @@
 * [Gherkin](docs/gherkin.md)
   * [Feature](docs/gherkin.md#feature)
   * [Scenario](docs/gherkin.md#feature)
-  * [Given/When/Then (Steps)](docs/gherkin.md#steps)
+  * [Given/When/Then (Steps)](docs/step_definitions.md#steps)
   * [Tags](docs/tags.md)
   * [Background](docs/gherkin.md#background)
   * [Scenario Outline](docs/gherkin.md#background)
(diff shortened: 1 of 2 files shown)


jbpros:
Thanks @Javix. I find this PR must easier to parse as it's focussed on a single topic and the updated table of content. Please keep doing that, that's really good :)

I think there is more to say about steps (what they represent, what do `Given`, `When` and `Then` mean, `And` and `But`, the sequential nature of steps, etc.) before introducing step definitions. That's why I think it would be better to keep the steps docs in the gherkin section. If you don't know what else to write about steps, I'd be happy to help you with that.

Could you move the steps section to `gherkin.md` and make a separate PR for the step definitions?

Javix:
@jbpros Thank you. As @aslakhellesoy already mentioned it in #34 
>Step definitions don't belong in this document. Please move to a separate step-definitions.md doc, in a separate PR.


I had to move the **Step definitions** section to a separate PR #42 . May be I misunderstood what exactly I had to move: **Step definitions** only or the entire **Steps** section including the **Step definitions** ? If so, I'll move **Steps** to back `gherkinh.md`, add the link to **Step definitions**  which is already in a `step_definitions.md`and in a separate PR #42 . Thank you.

Javix:
I added more description to `Given`, `When`, `Then` steps, added the content for `And` and `But` steps as well as asterisk character. It the content is OK, I could move Steps section (only) to `gherkin.md` and keep only **step definitions** content here.

gavvvr:
The info about asterisk-notation for step-definition shoud be moved to `gherkin.md` asap. As a Cucumber user I had to ask on Stack Overflow to discover how else i can define steps besides using a set of reserved keywords.

PR #67, opened 2016-09-01
Add tag expressions for ruby

nodo:
## Summary

Implements tag expressions for ruby.

## Motivation and Context

Related issues:
- https://github.com/cucumber/cucumber/issues/60
- https://github.com/cucumber/cucumber-ruby/issues/1003

## How Has This Been Tested?

The changes implements rspec unit tests and coverage reporting. The tests are similar to tag-expressions for javascript and java versions.

## Types of changes
-  New feature (non-breaking change which adds functionality).

## Checklist:


nodo:
@aslakhellesoy I have implemented tag-expressions-ruby. It is working and almost ready to be mirrored as a gem.

I have a doubt regarding the naming convention. At the moment, the gem is called "tag-expressions-ruby". However, it will be required by the code as "tag_expressions_ruby" (with underscores). Besides, the main module is called `TagExpressions`. 

Do you have a convention to follow in cucumber? Or do you have a suggestion regarding this point?

Looking forward to see your comments and improve the code.

mattwynne:
Awesome work @nodo.

Regarding names, I suggest we put this into a `Cucumber` namespace, so the require statements become `cucumber/tag_expressions`.

The idiomatic way to name the gem itself would then be `cucumber-tag_expressions`. That might look a bit awkward, but I think it's fine.

WDYT? @aslakhellesoy?

tag-expressions/ruby/lib/tag_expressions_ruby/expressions.rb
```diff
@@ -0,0 +1,63 @@
+module TagExpressions
```
mattwynne: I suggest this file moves to `lib/cucumber/tag_expressions`, and we add the `module Cucumber` around this `TagExpressions` module.

nodo:
Thanks @mattwynne !

I totally agree with your suggestion. I have changed the naming, but I am happy to discuss further.

tag-expressions/ruby/lib/cucumber/tag_expressions/parser.rb
```diff
@@ -0,0 +1,120 @@
+require 'cucumber/tag_expressions/expressions.rb'
+
+# Array monkey-patch
+class Array
```
mattwynne: We load Cucumber into the same Ruby process as other people's code, so we try to avoid adding monkey patches (there is some legacy code that does, but it's something we don't do in new code).

I suggest creating a decorator using [SimpleDelegator](https://hashrocket.com/blog/posts/using-simpledelegator-for-your-decorators) that you can use to create Arrays decorated with this method. You could call it `PoppableArray` or something.

aslakhellesoy:
Rebased and merged in f64f7a5fb7bfba0c0b44027e3bb3af2adb3f4e8c. I took the liberty to remove the monkey patch in 55092fe1fde0b023719c3221fa59a41b17c2e34e.

Excellent work @nodo - I like how it follows the original implementation while adopting a few rubyisms to make it more idiomatic. Enjoy the commit bit!

This is a tiny library, so straying from the structure of the other libraries is not a problem. For bigger libraries (like Cucumber Expressions) I think it's better to keep exactly the same structure, even if it means non-idiomatic code. That makes maintenance easier for people who aren't fluent in every language.

I'll see if I can push a gem :-)

aslakhellesoy:
I forgot to mention:

* https://travis-ci.org/cucumber/tag-expressions-ruby
* github.com/cucumber/tag-expressions-ruby (read-only mirror)

nodo:
Thanks @aslakhellesoy and @mattwynne . Really happy you like it!

Just a minor point, there is a typo in the description of the repository:
```
[READ-ONLY] Tag Expression Parser for Java - subtree of monorepo https://github.com/cucumber/cucumber
```
Java -> Ruby

aslakhellesoy:
Thanks @nodo - fixed the description

PR #71, opened 2016-09-14
[html-formatter] Use ESLint

aslakhellesoy:
This PR sets up [ESLint](http://eslint.org/) for html-formatter

### DIFF
--- a/html-formatter/nodejs/.eslintrc.yml
+++ b/html-formatter/nodejs/.eslintrc.yml
@@ -1,0 +1,12 @@
+root: true
+
+parserOptions:
+  sourceType: module
+
+plugins:
+  - node
+  - react
+extends:
+  - eslint:recommended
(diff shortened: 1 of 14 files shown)


mattwynne:
@aslakhellesoy are we using labels or issue name prefixes to separate these issues out? When I get an email notification I can't see tags, so I'd like to stick to using prefixes if that's OK with the rest of us?

aslakhellesoy:
I think we should use both, because:

* prefix, because what you said
* label, because it makes triage and filtering using the github UI a lot easier

New contributors are likely to not set any of them, so contributors should go in and add prefixes and labels when they are missing.

aslakhellesoy:
I also use prefixes in commit messages (except when I forget!)

mattwynne:
Sounds good.

PR #87, opened 2016-10-16
[cucumber-expressions] handle optional groups in regexp.

brasmusson:
## Summary

To be able to handle optional groups in regexp, null should be transformed to null.

## Details

[Java] When performing a constructor transform on the value null, the transformed object should also be null.

## Motivation and Context

To be able to use cucumber-expressions in Cucumber-JVM optional groups in regexps ([like this](https://github.com/cucumber/cucumber-jvm/blob/master/jruby/src/test/resources/cucumber/runtime/jrubytest/stepdefs.rb#L33-L35)) need to be handled. Currently the Java implementation throws a "Failed to invoke `new String("null")`" instead.

## How Has This Been Tested?

- [X] Tests are included
- [X] Has been tested with Cucumber-JVM (the stepdef linked above)

## Types of changes

## Checklist:

- [X] My change does not requires a change to the documentation.

### DIFF
--- a/cucumber-expressions/examples.txt
+++ b/cucumber-expressions/examples.txt
@@ -13,3 +13,7 @@ I have 22 cukes in my belly now
 /I have (-?\d+) cukes? in my (.*) now/
 I have 1 cuke in my belly now
 [1,"belly"]
+---
+/^Something( with an optional argument)?$/
+Something
+[null]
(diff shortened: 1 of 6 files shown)


aslakhellesoy:
LGTM!

PR #99, opened 2016-12-18
tag-expressions: (Java) Make the jar a bundle to support osgi.

brasmusson:
## Summary

Build the tag-expressions jar as a OSGi bundle.

## Details

* Declare the packaging type as `bundle`
* Use the `maven-bundle-plugin` to create the manifest according to the OSGi requirement.
 
## Motivation and Context

cucumber/cucumber-jvm#873 added support to execute Cucumber-JVM in OSGi containers. To not lose that ability when using the tag-expression library in Cucumber-JVM, the tag-expressions jar need to be built as an OSGi bundle.

## How Has This Been Tested?

I have locally run the [pax-exam example of Cucumber-JVM](https://github.com/cucumber/cucumber-jvm/tree/master/examples/pax-exam), which exemplifies testing OSGi bundles, successfully with these changes.

## Types of changes

### DIFF
--- a/tag-expressions/java/pom.xml
+++ b/tag-expressions/java/pom.xml
@@ -4,7 +4,7 @@
     <groupId>io.cucumber</groupId>
     <artifactId>tag-expressions</artifactId>
     <version>1.0.0</version>
-    <packaging>jar</packaging>
+    <packaging>bundle</packaging>
     <name>Cucumber Tag Expressions</name>
     <description>Parses boolean infix expressions</description>
     <url>https://github.com/cucumber/tag-expressions-java</url>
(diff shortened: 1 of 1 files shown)


mattwynne:
@aslakhellesoy @brasmusson how do we get this released?

aslakhellesoy:
@mattwynne please always update the History/CHANGELOG when you merge a PR - otherwise consumers have a hard time knowing what's new in a given release.

In this case, tag-expressions doesn't even have a change log, so that needs to be added too.

mattwynne:
@aslakhellesoy noted. Are you able to make a release? Is seems a lot of other stuff depends on this.

aslakhellesoy:
I'll make a release when I find some spare cycles.

PR #102, opened 2017-01-16
[cucumber-expressions] Handle multiple capture group regexps for matching

gpichot:
## Summary

When creating a new transform with multiple capture group regexps, cucumber fail to match expressions that are not using the first capturing regexp.

For instance, a transform with the captureGroupRegexps with the value ['red', 'yellow'] would fail to match an expression containing yellow.

## Details
The matching expression is now a disjonction of all the capturing regexps using non capturing group (with `(?:...)`).

## Motivation and Context

I had an issue when using multiple regexps for the same transform as it was only matching the first one. This commit fix it.

## How Has This Been Tested?

I added one test that use the second regexp of the `captureGroupRegexps` option of a transform (and that is not matching a subset of the first regexp).

## Types of changes

## Checklist:

### DIFF
--- a/cucumber-expressions/javascript/src/cucumber_expression.js
+++ b/cucumber-expressions/javascript/src/cucumber_expression.js
@@ -41,7 +41,7 @@ class CucumberExpression {
       this._transforms.push(transform)
 
       const text = expression.slice(matchOffset, match.index)
-      const captureRegexp = `(${transform.captureGroupRegexps[0]})`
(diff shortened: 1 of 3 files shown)


aslakhellesoy:
Thanks for this! I'll merge it as after I've implemented a Java and Ruby implementation too.

aslakhellesoy:
What a great patch @gpichot - really well done.

gpichot:
Thanks for the merge @aslakhellesoy ! :)

Issue #114, opened 2017-02-10
Rename Transform to Parameter


aslakhellesoy:
## Summary

The term *parameter* is more descriptive than *transform*.

First off, parameters don't necessarily transform anything - their transform functions may be an identity function (leaving the string unchanged). Some parameters only apply a more restrictive pattern than `.+`, for example `[A-Z]\w+`.

Moreover, *transform* describes *what*, i.e. a custom parameter type, with a specific name. Using the term *transform* says more about the implementation of this feature. 
## Current Behavior

~~~
~~~

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Version used:
* Operating System and version:
* Link to your project:

aslakhellesoy:
Fixed by 5feacc737bc756c436c137699a4da0616e97c25b

mattwynne:
@aslakhellesoy can I ask you to try and prefix your issue names with the sub-project, as well as adding the label? It makes it much easier to parse my email notifications and know which ones I want to open and read.

aslakhellesoy:
I'll do that for future ones. Can I ask you to rename the existing ones?

mattwynne:
Done!

PR #117, opened 2017-02-17
Deprecate {name:type} syntax in favoud of just {type}

aslakhellesoy:
## Summary

After having used Cucumber expressions daily for a few weeks I have realised that:
* The `{name:type}` syntax is usually redundant for both built-in and custom parameters (previously called transforms, see #114)
*The `{name:type}` syntax is never needed, not even for expressions with multiple parameters of the same type. It is sufficient to give different names to the function/method parameters of the stepdef. 
* The `{name:type}` syntax is ugly and causes more visual/cognitive strain

So let's get rid of it! This PR prints a deprecation warning with a recommendation to `STDERR` if the `{name:type}` syntax is used. In a future release we can remove support for it altogether.

## Types of changes

## Checklist:

### DIFF
--- a/cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/CucumberExpression.java
+++ b/cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/CucumberExpression.java
@@ -25,19 +25,23 @@ public class CucumberExpression implements Expression {
         regexp.append("^");
         int typeIndex = 0;
         while (matcher.find()) {
-            Type type = types.size() <= typeIndex ? null : types.get(typeIndex++);
             String parameterName = matcher.group(1);
             String typeName = matcher.group(3);
+            if (typeName != null) {
+                System.err.println(String.format("Cucumber expression parameter syntax {%s:%s} is deprecated. Please use {%s} instead.", parameterName, typeName, typeName));
+            }
+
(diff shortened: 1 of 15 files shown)


jenisys:
@aslakhellesoy 
Just a comment, the syntax ``{name:type}`` is necessary for dynamic languages.
Actually, there are three schemas (relevant in this case):
* ``{name:type}`` when you have a named parameter with a type transformation
* ``{name}`` when you have a parameter that is the string (and the ``:type`` is a ``string`` implicitly)
* ``{:type}`` when only want to specify the type (transformation) and use the position for the arg.

When you replace it with ``{type}`` (instead of ``{:type}``) you are breaking it (at least for me).
And I use named parameters at lot. In addition, they make the documentation of the step definitions much more readable.

SEE ALSO:
* https://github.com/r1chardj0n3s/parse (where the syntax came from, at least in ``behave``)

aslakhellesoy:
> the syntax {name:type} is necessary for dynamic languages

I'm not sure why you think so. We've been using Cucumber.js 2.0.0-rc.7 for several weaks, using the new transform feature and custom transforms (recently renamed to parameters).

Have you looked into the [custom parameter](https://docs.cucumber.io/cucumber-expressions/#custom-parameters) feature?

For dynamic languages, processing of `{name}` is as follows:
1) If there is a registered parameter type for `name`, use it
2) Otherwise, don't transform the match

PR #124, opened 2017-03-03
event-protocol: Add pickle event

mattwynne:
## Summary

Add a `pickle` event to the event protocol validation schemas.

## Motivation and Context

This event will allow us to broadcast the list of test cases about to be executed by a Cucumber. We can use it to build a progress bar, for example, to be updated as test results arrive.

## How Has This Been Tested?

I've added an example of a valid event, and used `make` in the project root.

## Types of changes

## Checklist:


mattwynne:
@aslakhellesoy as I try implementing this in the consumer, it's quite a pain to have to keep going up the tree for the `uri`. I would be easier to work with if we duplicated the `uri` down the tree in each of the locations.

How would you feel about that?

aslakhellesoy:
@mattwynne it was added in gherkin/gherkin@21873b0db02af797abc2f8b98f49f603de74abb8 (as `path`), then removed again in gherkin/gherkin@26ed643cd5238b277ff297179b8b20d7748186f7. As you can see I did a poor job in the commit message explaining why :-(

My vague recollection is that it "felt" better. Less duplication.

Do you have some code I can see that illustrates the pain? Instead of going up the tree to get the `uri`, couldn't you pass the `uri` when you walk down the tree?

PR #128, opened 2017-03-06
event-protocol: Remove series and timestamp

aslakhellesoy:
## Summary

Remove `series` and `timestamp` properties from all events.

## Details

Modified the schema

## Motivation and Context

This aligns Gherkin generated events (`source` and `pickle`) with the event-protocol validator.
    
Gherkin-created pickle events currently do not have those properties. I don't think we have any consumers that require them at the moment.

We can add them back when/if a consumer needs them. I'd rather not add them prematurely, Updating Gherkin in 8 languages is a lot of work. 

## How Has This Been Tested?

    ../../gherkin/javascript/bin/gherkin ../../gherkin/testdata/good/*.feature --no-ast | ./validate.js 

## Types of changes

## Checklist:


mattwynne:
The GUi spike uses the timestamp from the start message but not for the rest. The series isn't used yet, but I have immediate plans to use it. I think it's a good idea. Let's just make it optional.

aslakhellesoy:
If a property is optional, what would you do if the property is absent? I think optional properties should be avoided in general.

I think `series` is superfluous because there are other mechanisms available to group related events. The most obvious one being a session, or a stream: Every time someone connects and pumps events, they can be considered related.

For `timestamp` - could you just generate a timestamp when you receive the event?

mattwynne:
Agreed about `series` - I'll try that out in the GUI spike.

I think that for the `start` event, it makes sense to allow the sender to put out extra metadata in the event if they want to. Especially while we figure out the finer details of the protocol. The listeners just need to be robust enough to handle those optional parameters being absent.

`timestamp` would be useful I think, for example if we allow results to be cached locally while the client is off the internet. But if it's awkward to send from some senders, we can make it optional, and use the receipt timestamp as a fallback.

Another example is `workingDirectory`. It will certainly be useful to have this for the GUI, but I'm not sure we need to make it mandatory.

mattwynne:
I'm gonna merge this, and we can talk about adding the timestamp and other optional params back in separately.

aslakhellesoy:
\o/

PR #129, opened 2017-03-06
Event protocol gherkin document schema

aslakhellesoy:
## Summary

Validate `gherkin-document` events (AST events)

## Details

Added `gherkin-document.json` JSON Schema

## Motivation and Context

With #128 we have a validator that can validate all of Gherkin CLI's output.

## How Has This Been Tested?

```
../../gherkin/javascript/bin/gherkin ../../gherkin/testdata/good/tags.feature --no-source --no-pickles | ./validate.js
make clean && make
```

## Types of changes

## Checklist:


event-protocol/schemas/defs.json
```diff
@@ -38,6 +38,49 @@
...
+        "minimum": 1
+      }
+    },
+    "required": [
+      "line",
+      "column"
+    ],
+    "additionalProperties": false
+  },
+  "ast-tag": {
```
mattwynne: I don't understand why we need this extra def. The only difference from the other `tag` def I can see is the extra `type` attribute which is always set to `Tag`. Can't we just get rid of that in the source data and then we can use one tag structure for pickles and ast elements?

event-protocol/schemas/gherkin-document.json
```diff
@@ -0,0 +1,59 @@
...
+              "type": "string"
+            },
+            "keyword": {
+              "type": "string"
+            },
+            "name": {
+              "type": "string"
+            },
+            "children": {
+              "type": "array"
```
mattwynne: Can we say here what type the things in the array will be?

mattwynne:
Do we have any consumers that use this event at the moment?

mattwynne:
Don't forget to update the README with details of the new event.

event-protocol/schemas/defs.json
aslakhellesoy: The JSON representation of an AST Tag and a Pickle Tag are slightly different. The Java and .NET implementations of Gherkin even have two different classes (the generated JSON depends on the fields in those classes):

* [Tag](https://github.com/cucumber/gherkin/blob/master/java/src/main/java/gherkin/ast/Tag.java) which inherits fields from [Node](https://github.com/cucumber/gherkin/blob/master/java/src/main/java/gherkin/ast/Node.java)
* [PickleTag](https://github.com/cucumber/gherkin/blob/master/java/src/main/java/gherkin/pickles/PickleTag.java)

This `type` attribute on AST nodes is useful in some corner cases, where a collection can have different types. For example, the `Feature.children` collection can have 3 different types of elements - `Background`, `Scenario` and `ScenarioOutline`. The `type` attribute makes it easy to know what type it is.

For pickles we don't have this problem, because it's a flatter, simpler structure where we don't have this polymorphism.

Since `Tag` != `PickleTag` I thought it made more sense to have distinct JSON schemas for them, even if they look similar. It's less confusing, and also allows us to change one of them without having to change the other.

event-protocol/schemas/gherkin-document.json
aslakhellesoy: Ideally we should, yes. As mentioned above, the elements can be one of 3 types, so I think we need to use something like [oneOf](http://json-schema.org/latest/json-schema-validation.html#rfc.section.5.24). I was lazy and decided that super strict validation of `gherkin-document` is less important than all of the other schemas, because `gherkin-document` objects will practically only be generated by the Gherkin library, and that's *really* thoroughly tested.

That said, I think it would be nice to have a full schema - a schema is useful for more than validation - it can also be used to generate documentation. See http://json-schema.org/implementations.html under "Documentation generation". Especially [Doca](https://github.com/cloudflare/doca/) looks interesting.

aslakhellesoy:
> Do we have any consumers that use this event at the moment?

[Gherkin-Lint](https://github.com/cucumber/cucumber/tree/gherkin-lint-stream/gherkin-lint) uses it.

event-protocol/schemas/gherkin-document.json
mattwynne: I'd like to have a full schema. I think it sets a good example for the rest of the protocol, and could also come in useful later if we want to send bits of AST around in other messages.

I'm quite enjoying the process of modelling the schema, so happy to help.

event-protocol/schemas/defs.json
mattwynne: Understood.

event-protocol/schemas/gherkin-document.json
aslakhellesoy: Ok sounds good. You can expand the schema on this branch.

PR #130, opened 2017-03-09
event-protocol: add some optional metadata in `start` event

mattwynne:
This allows the GUI to:

1) Identify (and display) the project that the results it's receiving
are coming from.
2) Display an accurate start time for the test run.

I've specified these attributes in the schema, but made them
non-required. I think this is better than just turning on
`additionalProperties` which would be too flexible.

I'd like to work towards a point where all emitters send out these
attributes (and we can then make them mandatory), but this flexibility
allows for some growth and development within the protocol.

An alternative would be to have a grab-bag `metadata` property of type
`object` and not specify the shape of that object.

### DIFF
--- a/event-protocol/schemas/start.json
+++ b/event-protocol/schemas/start.json
@@ -6,6 +6,12 @@
   "properties": {
     "type": {
       "type": "string"
+    },
+    "workingDirectory": {
+      "type": "string"
+    },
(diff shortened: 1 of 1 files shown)


mattwynne:
We've decided to put this info into a new `test-case-started` event instead.

PR #160, opened 2017-03-18
WIP: Cucumber redux

aslakhellesoy:
## Summary

Add a standalone [Redux](http://redux.js.org/) package for the Cucumber event protocol.

## Details

Builds an [Immutable.js](https://facebook.github.io/immutable-js/) state tree of Gherkin documents and attachments, using event protocol events as Redux actions.

## Motivation and Context

React based consumers of the Event Protocol need to render Gherkin Documents with additional information such as results, stack traces and other attachments such as screenshots.

The reducer extracted in the new cucumber-redux package builds a state tree suitable for React rendering.

## How Has This Been Tested?

Using mocha

## Types of changes

## Checklist:


mattwynne:
I like the pattern / design here, but I worry that this separate library might be premature - that we should build this within the GUI (monolith-first), then extract it out. WDYT?

aslakhellesoy:
This was extracted after an initial prototype was built (cucumber-html, which I think we should retire in favour of a better electron-based implementation).

I don't see much risk in this being premature. It was a little cumbersome to extract it, and in my experience it just becomes harder to do as the code evolves.

It's easy to hack on the various components without having to build or publish anything. Just `make` each individual project (`cucumber-redux`, `cucumber-react`, `cucumber-html`, `cucumber-gui`) and they will be linked with `yarn link`.

PR #161, opened 2017-03-18
event-protocol: Add test-run started event

mattwynne:
## Summary

Add a `test-run-started` event. This contains contextual information about the test run.

## Details

More PRs coming for test results events

## Types of changes

## Checklist:


aslakhellesoy:
We probably need this.

Going forward I would like to see additions or changes to the event protocol to be done in the same commit as the consumer(s) requiring that change. This means that consumers driving the change to the protocol will have to be added to this repo.

mattwynne:
I'll do this differently I think, and propose all the test result events together so we can review the whole lot at the same time.

mattwynne:
I agree about having the consumer in the same PR in general, but I'm worried it will make for a very big PR if we have the GUI and the protocol changes all in one. I'm planning to send one for the GUI and another for the protocol, building on that one.

PR #163, opened 2017-03-21
Build (almost) the entire monorepo in Docker

aslakhellesoy:
## Summary

This PR adds a Docker image with software installed to build most of the libraries. It also sets up a Travis build. 

## Details

The docker image is based on Alpine Linux. The base image is `mhart/alpine-node` rather than `alpine` because the latter only has a nodejs package version 6 (we need 7).

Some libraires are not built (or only partially built):

* `gherkin/c` - compiled, but not tested - there is a failing test (bug)
* `gherkin/dotnet` - there is no mono package for alpine yet
* `gherkin/go` - there is no go package for alpine yet
* `gherkin/objective-c` - there is no `xcodebuild` for Linux

There libraries are fully built in their subrepos' Travis builds, which do not run in Docker and therefore have more tooling available.

## Motivation and Context

The main reason for this PR is to have CI builds for contributor PRs.

The other reason is to make it easier for contributors to run tests locally without having to install a lot of software (just Docker).

## How Has This Been Tested?

Travis runs the build!

## Checklist:


aslakhellesoy:
I've opened a separate ticket about documenting this: #166

PR #164, opened 2017-03-21
[gherkin-ruby] Add ruby-head as a allow failures in .travis.yml.

junaruga:
## Motivation and Context

I want to add  ruby-head to Travis as a allow failures for `gherkin-ruby`.
The motivation and the detail is same with cucumber/cucumber-ruby#1087

## How Has This Been Tested?

When I send this PR here, below Travis will test my `.travis.yml` right?
https://travis-ci.org/cucumber/gherkin-ruby

## Types of changes

## Checklist:

### DIFF
--- a/gherkin/ruby/.travis.yml
+++ b/gherkin/ruby/.travis.yml
@@ -1,10 +1,16 @@
 language: ruby
 
 rvm:
+  - ruby-head
   - 2.4.0
   - 2.3.3
   - 2.2.6
   - jruby-9.1.7.0
+
+matrix:
+  allow_failures:
+    - rvm: ruby-head
+  fast_finish: true
 
 script: make
 


junaruga:
My PR work flow is correct?
How can Travis run ` gherkin/ruby/.travis.yml ` before the PR is merged?

aslakhellesoy:
Thanks @junaruga. Only the top level `.travis.yml` gets built by travis, so this won't actually get built by Travis until I sync to https://github.com/cucumber/gherkin-ruby which I'll do shortly.

aslakhellesoy:
The build is passing on `ruby-head`: https://travis-ci.org/cucumber/gherkin-ruby

I wonder - wouldn't it be a good thing to disallow failures on `ruby-head` so it can be brought to our attention early and fix any issues? I wouldn't expect it to be *that* unstable.

junaruga:
@aslakhellesoy thanks. ok I understood it.

I think this logic is not so useful until the new version Ruby's actual preview release.
But this is still better than current situation.
And from my experience I contributed to `cucumber-ruby` for Ruby 2.4.0, I thought it was useful.

When I tried to do it, `cucumber` contributes did not have Ruby 2.4.0 environment in local.
It was hard to explain them about the Ruby 2.4.0 issue.

junaruga:
@aslakhellesoy I am working on Fedora Project's Ruby team.
Fedora Project likes to use latest version's package as their RPM package.
It is kind of experimental distribution.

So, I sometimes discover the issue for new version Ruby, as faster than the upstream project members.
One of the merit for this PR is to reduce my work :)

PR #166, opened 2017-03-21
event-protocol: Make binary command executable and add hashbang

mattwynne:
I might have missed something, but I couldn't work out to run it (from
another directory) anymore since it moved into the bin directory in
3dbafa2e4a327a4251038a97063b7de5e0da3f88


aslakhellesoy:
Yeah weird - I probably ran it with `node`.

I stole your `readline` code from your `cucumber-gui-spike` BTW :-)

PR #173, opened 2017-03-26
event-protocol: Add ruby formatter

mattwynne:
## Summary

Add a plugin for cucumber-ruby that produces events according to the protocol defined in #172

## Details

This plugin emits the necessary NDJSON events for driving the prototype GUI (see #171) and satisfies the protocol defined in #172.

## Motivation and Context

This provides the first version of a producer of the new test-result events proposed in #172. It can act as a reference implementation for other Cucumbers to follow when implementing their own.

Keeping the plugin here in the monorepo rather than in the cucumber-ruby codebase will enable other implementors to re-use tests (or perhaps we can even see how to develop a shared test suite).

Once this gem has been published, we'll use it as a dependency in Cucumber-Ruby 3.0, and make it one of the standard command-line options.

## How Has This Been Tested?

The set of examples in `/event-protocol/ruby/examples` each contain an `expected-events.ndjson`. The RSpec tests for the project run Cucumber (with the plugin enabled) for each of the examples, and compare the output from the plugin with that file.

Because of some non-deterministic data in the events, the data is somewhat normalised before comparison. Read the code for details.

## Types of changes

## Checklist:


mattwynne:
@aslakhellesoy the build is broken as there's no Makefile in the `event-protocol` directory anymore. Should I add one that builds both the Ruby plugin and runs the validator schema tests?

aslakhellesoy:
Yes please do. Each `Makefile` should have a `default` and `clean` target. Example parent Makefile here: https://github.com/cucumber/cucumber/blob/master/cucumber-expressions/Makefile

event-protocol/ruby/Gemfile
```diff
@@ -0,0 +1,4 @@
+source 'https://rubygems.org'
+gemspec
+
+gem "cucumber", path: "../../../cucumber-ruby"
```
aslakhellesoy: This might work on your machine, but it [breaks the build](https://travis-ci.org/cucumber/cucumber/builds/216494566#L285).

I suggest depending on the git repo (with a particular branch) instead.

event-protocol/ruby/LICENSE.txt
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2017 Cucumber Limited and contributors
```
aslakhellesoy: Never add year to Copyright. It's not required and nobody remembers to update it.

event-protocol/ruby/Gemfile
mattwynne: Oh yeah I should have taken that out. Thanks!

Issue #131, opened 2017-03-10
Cucumber Expressions: Non Capture Group


KyleFairns:
## Summary

I'm linking to the issue that I created over in the cucumber-js repo [here](https://github.com/cucumber/cucumber-js/issues/784) in order to trace back the whole conversation.

There are many cases in our step definitions where we have _a varity of repeated groups that don't change the steps functionality, but does do quite a bit for readablilty_ in the feature files:

* "must"/"should"
* optional words like "the", "I" etc that would match `Given I fill in "Foo"` & `And fill in "Bar"`
* lists of interchangable words

Say that you are using these alternatives across a large set of step definitions and you wish to change them all. This is where the problem (and the solution) both appear.

## Current way of completing the task

At current, we can do alternatives like so:

```javascript
Given('{pronoun} have/has {int} child(ren)', (pronoun, childCount) => {
  // Code that does not require the "pronoun" parameter at all
});
```

However, this does mean that we have a parameter that is never actually used in the step definition that we have and this is the only way of storing the list to be used in many differing steps.

## Possible Solution

Perhaps square brackets could be used for things that don't need to be captured, but still require the transformation.

```javascript
Given('[pronoun] have/has {int} child(ren)', (childCount) => {
   // Code
});

When('[pronoun] picks up {stringInDoubleQuotes} from {stringInDoubleQuotes}', (child, place) => {
   // Code
});

Then('[pronoun] should have {int} child(ren) in [pronoun] care', (childCount) => {
  // Code
});

```
Another plus side, is that if the parameter needs to be captured, the only difference would be the brackets that you use in that step definition:

```javascript
  Then('{pronoun} does something', (pronoun) => {
     if(pronoun == "Jeff"){
        // Do something different for Jeff
     } else {
       // Do your thing
     }   
  });
```

## Context & Motivation

We will be working with cucumber on a wide variety of projects in our company and I can guarantee that many of them will require a large number of declaratively written step definitions that require many alternatives that won't be captured for use within the step definitions.

This would make it easier for us to keep track of all of the alternatives that we will be using, and will mean that we wont have to search through the many step definition files for all of the uses..

And I am certain that others in the cucumber community will be in the same situation as I.

mattwynne:
Couldn't you just use regular string interpolation here?

i.e.
```
const pronoun = "He/She"
Given(`${pronoun} have/has {int} child(ren)`, (pronoun, childCount) => {
  // Code that does not require the "pronoun" parameter at all
});
```

[...]

PR #179, opened 2017-04-08
cucumber-expressions: #131 Cucumber Expressions: Non Capture Group

KyleFairns:
## Summary
I'm linking to the issue that I created over in the cucumber-js repo [here](https://github.com/cucumber/cucumber-js/issues/784) and the change that this relates to in the cucumber parent project repo [here](https://github.com/cucumber/cucumber/issues/131) in order to trace back the whole conversation.

There are many cases in our step definitions where we have _a varity of repeated groups that don't change the steps functionality, but does do quite a bit for readablilty_ in the feature files:

* "must"/"should"
* optional words like "the", "I" etc that would match `Given I fill in "Foo"` & `And fill in "Bar"`
* lists of interchangable words

Say that you are using these alternatives across a large set of step definitions and you wish to change them all. This is where the problem (and the solution) both appear.

## Details

Changed the regex that matches parameter groups to include square brackets as well as curly braces.
Later on, a line to differentiate between the two types of braces - naming curly as capture and square as non-capture.
Later still, an edit to the line that adds to the Regex, to make sure curly has capture group regex, and square has non-capture group regex.

Removed 3 tests that check that cucumber escapes square brackets - it now doesn't, as these are now special characters in cucumber-expressions. 

## Motivation and Context

We will be working with cucumber on a wide variety of projects in our company and I can guarantee that many of them will require a large number of declaratively written step definitions that require many alternatives that won't be captured for use within the step definitions.

This would make it easier for us to keep track of all of the alternatives that we will be using, and will mean that we wont have to search through the many step definition files for all of the uses..

And I am certain that others in the cucumber community will be in the same situation as I.

## How Has This Been Tested?

I've done a set of tests for this, by checking the number of parameters needed in an example cucumber suite (I can load this to a repository on request). The square brackets now match with the transforms that we set, but do not pass a parameter, as they are optional.

cucumber_code

Note: The two failures are intentional, to show that the code actually works
cucumber_running

## Types of changes

## Checklist:

### DIFF
--- a/cucumber-expressions/javascript/examples.txt
+++ b/cucumber-expressions/javascript/examples.txt
@@ -6,10 +6,6 @@ I have {int} cuke(s) in my {bodypart} now
 I have 1 cuke in my belly now
 [1,"belly"]
 ---
-I have {int} cuke(s) and some \[]^$.|?*+ {something}
-I have 1 cuke and some \[]^$.|?*+ characters
-[1,"characters"]
----
 /I have (\d+) cukes? in my (.+) now/
 I have 22 cukes in my belly now
 [22,"belly"]
--- a/cucumber-expressions/javascript/src/cucumber_expression.js
+++ b/cucumber-expressions/javascript/src/cucumber_expression.js
@@ -1,87 +1,106 @@
-const matchPattern = require('./build_arguments')
+"use strict";
 
-class CucumberExpression {
-  /**
-   * @param expression
-   * @param types Array of type name (String) or types (function). Functions can be a regular function or a constructor
-   * @param parameterTypeRegistry
-   */
-  constructor (expression, types, parameterTypeRegistry) {
-    const PARAMETER_REGEXP = /\{([^}:]+)(:([^}]+))?}/g
-    const OPTIONAL_REGEXP = /\(([^)]+)\)/g
-    const ALTERNATIVE_WORD_REGEXP = /(\w+)((\/\w+)+)/g
+var _createClass = function () { function defineProperties(target, props) { for (var i = 0; i < props.length; i++) { var descriptor = props[i]; descriptor.enumerable = descriptor.enumerable || false; descriptor.configurable = true; if ("value" in descriptor) descriptor.writable = true; Object.defineProperty(target, descriptor.key, descriptor); } } return function (Constructor, protoProps, staticProps) { if (protoProps) defineProperties(Constructor.prototype, protoProps); if (staticProps) defineProperties(Constructor, staticProps); return Constructor; }; }();
 
-    this._expression = expression
-    this._parameterTypes = []
-    let regexp = "^"
-    let typeIndex = 0
-    let match
-    let matchOffset = 0
+function _classCallCheck(instance, Constructor) { if (!(instance instanceof Constructor)) { throw new TypeError("Cannot call a class as a function"); } }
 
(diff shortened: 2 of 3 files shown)


aslakhellesoy:
Hi @KyleFairns - thanks for this very detailed pull request!

You're proposing to add new functionality here, and this needs both new tests and documentation. Please add tests to `cucumber_expression_regexp_test.js`. I recommend one using `[int]` and another one using `[float]`. This should cover the two cases where a parameter type has 1 regexp (`float`) or two (`int`).

Please also update the documentation in [README.md](https://github.com/cucumber/cucumber/blob/master/cucumber-expressions/README.md). As you'll see, it's possible to link to examples in the tests, which will show up in the [documentation](https://docs.cucumber.io/cucumber-expressions/).

I have a few more comments which I'll leave inline.

cucumber-expressions/javascript/examples.txt
```diff
@@ -6,10 +6,6 @@ I have {int} cuke(s) in my {bodypart} now
 I have 1 cuke in my belly now
 [1,"belly"]
 ---
-I have {int} cuke(s) and some \[]^$.|?*+ {something}
```
aslakhellesoy: Rather than removing the entire example, please just remove the `[]`.

cucumber-expressions/javascript/src/cucumber_expression.js
```diff
@@ -50,10 +50,11 @@ class CucumberExpression {
       if (!parameter) {
         parameter = parameterTypeRegistry.createAnonymousLookup(s => s)
       }
-      this._parameterTypes.push(parameter)
+        parameter.matchType = match[0].split("")[0] === "{" ? "capture" : "non-capture";
```
aslakhellesoy: Here you're mutating state in an object that really ought to be immutable. I think it would be better to simply make `matchType` a local variable.

cucumber-expressions/javascript/src/cucumber_expression.js
```diff
@@ -7,7 +7,7 @@ class CucumberExpression {
    * @param parameterTypeRegistry
    */
   constructor (expression, types, parameterTypeRegistry) {
-    const PARAMETER_REGEXP = /\{([^}:]+)(:([^}]+))?}/g
+      var PARAMETER_REGEXP = /[{\[]([^}\]:]+)(:([^}\]]+))?[}\]]/g;
```
aslakhellesoy: I think it would be better to make the opening `[{\[]` and closing `[}\]]` capture groups. That would simplify the code below where `matchType` is assigned.

Formatting:
* Indent with 2 spaces
* Use `const`, not `var`
* No semicolons

cucumber-expressions/javascript/src/cucumber_expression.js
```diff
@@ -72,9 +73,9 @@ class CucumberExpression {
   }
 }
 
-function getCaptureRegexp (regexps) {
+function getCaptureRegexp (regexps, matchType) {
   if (regexps.length === 1) {
-    return `(${regexps[0]})`
+      return matchType === "capture" ? "(" + regexps[0] + ")" : "(?:" + regexps[0] + ")";
```
aslakhellesoy: 2 space indentation, no semicolon

aslakhellesoy:
Closed in favour of #220. See #211 for details. Thanks!

PR #186, opened 2017-04-19
cucumber-expressions: Allow duplicate regexps in parameter types

aslakhellesoy:
## Summary

Fix for #132

## Details

It's now OK to define multiple parameter types with the same regexp.

For undefined steps, multiple snippets may be generated, using all combinations of candidate parameter types.

Consider this step (without keyword)

    I bought a blue ball on 28/03/1983

With `blue` matching parameter types `{color}` and `{csscolor}`, and `28/03/1983` matching `date`, `datetime` and `timestamp` parameter types, the following expressions would be generated and used in snippets:

    I bought a {color} ball on {date}
    I bought a {color} ball on {datetime}
    I bought a {color} ball on {timestamp}
    I bought a {csscolor} ball on {date}
    I bought a {csscolor} ball on {datetime}
    I bought a {csscolor} ball on {timestamp}

Clearly there are many combinations here, but a very large number of combinations is unlikely:
* Few projects are likely to have more than 3 parameter types matching the same text
* Few steps are likely to match more than 3 parameters.

So in most cases there will only be 1-4 variants, and rarely above 9.

## Motivation and Context

That's explained in #132

## How Has This Been Tested?

Lots of unit tests

## Types of changes

## Checklist:

## Language implementations

All of these must be completed before merging

- [x] Java
- [ ] JavaScript
- [ ] Ruby


aslakhellesoy:
@jbpros you think you could port this to JavaScript - and update Cucumber.js to use it?

aslakhellesoy:
@jbpros @charlierudolph I've started on the JavaScript port, so let's not dupe efforts.

PR #188, opened 2017-04-21
Add "Dado que" to Spanish translation for "Given"

nilp0inter:
"Dado que" is a better translation in Spanish for the "Given" word than "Dado" is. Also, the use of the current translation, "Dado", in Spanish generates a lot of duplication in steps.

In the following example:

Scenario: eat 5 out of 12
  Given there are 12 cucumbers
  When I eat 5 cucumbers
  Then I should have 7 cucumbers

The Spanish translation with the current translation is the following:

Escenario: me como 5 de 12
  Dado que hay 12 pepinos
  Cuando me como 5 pepinos
  Entonces debería tener 7 pepinos

Initially you may think "Dado" is the right choice for the translation, but is not the case. In English you can swap any of the steps without any problem, but in Spanish you can't.

Given there are 12 cucumbers == When there are 12 cucumbers == Then there are 12 cucumbers
Given I eat 5 cucumbers == When I eat 5 cucumbers == Then I eat 5 cucumbers
Given I should have 7 cucumbers == When I should have 7 cucumbers == Then I should have 7 cucumbers

But in Spanish this is not the case, we can't swap the steps without making new steps:

Dado que hay 12 pepinos != Cuando que hay 12 pepinos != Entonces que hay 12 pepinos (This makes no sense)
Dado me como 5 pepinos (The right way is "Dado que me com...") != Cuando me como 5 pepinos == Entonces me como 5 pepinos
Dado debería tener 7 pepinos (The right way is "Dado que deb...") != Cuando debería tener 7 pepinos == Entonces debería tener 7 pepinos

Given that "Dado que" works most of the time, but "Dado" works sometimes as well, my proposal is to add "Dado que" and maintain "Dado" in the translations list.

Thank you.

Best regards,
Roberto

## Types of changes

## Checklist:

### DIFF
--- a/gherkin/gherkin-languages.json
+++ b/gherkin/gherkin-languages.json
@@ -945,6 +945,7 @@
     "given": [
       "* ",
       "Dado ",
+      "Dado que ",
       "Dada ",
       "Dados ",
       "Dadas "


aslakhellesoy:
Thanks @nilp0inter 

Makes sense to me, but shouldn't the other conjugations of *dar* also get a synoym? *dada que* etc.

nilp0inter:
No, _Dada que_ is not well formed.

*Dado que* is like a shorter form of *Dado el hecho de que* (*Given the fact that...*); *the fact* in Spanish is _male_, so the sorter form must be _male_ as well. Yeah, Spanish is tricky.

Let me clarify that:

- *Dado que* hay 12 pepinos (Given there are 12 cucumbers)
- *Dado que* hay 12 zanahorias (Given there are 12 carrots)

In both examples the correct translation is *Dado que* even though the genders are different (_pepino_ is male, _zanahoria_ is female). This is because you can express the same in this (cumbersome) way:

- *Dado* el hecho de que hay 12 pepinos
- *Dado* el hecho de que hay 12 zanahorias

Hope it makes sense.

aslakhellesoy:
Unfortunately, this breaks backwards compatibility, so we won't merge this. See #273 for details.

PR #201, opened 2017-05-08
Gherkin-Java: Added female, singular given in Romanian

tsundberg:
## Summary

The female, singular form for 'Given' was missing for Romanian.
Added it after suggestion from Bianca (native Romanian)


tsundberg:
I see, I will move this pull request then. Thank you for pointing that out.

PR #212, opened 2017-05-25
Cucumber: Fix 'What is not OK' not being formatted as header in Code of Conduct

mpkorstanje:
## Summary

Fix 'What is not OK' not being formatted as header. 

## Details

There is a non-space whitespace character after the header marker. This make githubs md parser ignore the header.

image


nodo:
Closed by #230. Thanks for your PR @mpkorstanje! 👍  This PR and #230 are very similar but #230, fixes another typo as well and I have merged that one.

PR #216, opened 2017-06-02
cucumber-expressions: make CAPTURE_GROUP_PATTERN aware of non-…

### LINKED ISSUE #211: RegularExpression constructor is not filtering non-capturing groups
_From @Jordyderuijter on May 23, 2017 11:4_

Hi there,

It seems like there is a bug in the RegularExpression class in the cucumber-expressions dependency.
The regex in this constructor looks for capturing groups when processing steps using the regex:

 `/\(([^(]+)\)/g`

This regex finds everything between braces (a.k.a. capturing groups), but it also accepts non-capturing groups. This causes the constructor to generate parameter types for the non capturing groups (instead of the capturing groups only). 

These parameter types are used in the buildArguments function, but are only applied to capturing groups, since the given step regex filters out the non-capturing groups correctly.

I would expect the RegularExpression constructor to filter out these capturing groups correctly.

The step and step definitions I used to reproduce the issue are:

Step:
`I can cancel the 1st slide upload`

Step definition:
`(\S+) ?(can|cannot)? (?:delete|cancel) the (\d+)(?:st|nd|rd|th) (attachment|slide) ?(?:upload)?`

In this case 7 parameter types are being generated by the constructor, while it should only create 4, since 3 of them are non-capturing groups. As a result, the parameter type that belongs to (\d+) is being applied to (attachment|slide), treating it as an integer. This results into a NaN value that is being passed to my step definition.

Version info:

- Cucumberjs: 2.1.0
- Node: v6.9.1
- Npm: 3.10.8

_Copied from original issue: cucumber/cucumber-js#837_

kAworu:
…capturing group.

## Summary

Before this patch, non-capturing group would be accounted for in the
match count, wrongly shifting the transformation.

## Motivation and Context

Fixes #211 

Using a combination of non-capturing group regexp in combination with regexp matching a registered type would apply transformation to the wrong group, yielding unexpected results.

## How Has This Been Tested?

Tests updated, before the patch we get:
```
  1) RegularExpression take into account non-capturing group:

      AssertionError: [ '22', NaN ] deepEqual [ 22, 'tail' ]
      + expected - actual

       [
      -  "22"
      -  NaN
      +  22
      +  "tail"
       ]

```

## Types of changes

Stricly speaking it is a Breaking change, although the probability that users worked around it are slim.

## Checklist:

### DIFF
--- a/cucumber-expressions/javascript/src/regular_expression.js
+++ b/cucumber-expressions/javascript/src/regular_expression.js
@@ -5,7 +5,7 @@ class RegularExpression {
     this._regexp = regexp
     this._parameterTypes = []
 
-    const CAPTURE_GROUP_PATTERN = /\(([^(]+)\)/g
+    const CAPTURE_GROUP_PATTERN = /\((?!\?:)([^(]+)\)/g
 
     let typeIndex = 0
     let match
--- a/cucumber-expressions/javascript/test/regular_expression_test.js
+++ b/cucumber-expressions/javascript/test/regular_expression_test.js
@@ -54,6 +54,13 @@ describe(RegularExpression.name, () => {
     assert.deepEqual(match(/(.*)/, '-1.22', [s => parseFloat(s)]), [-1.22])
   })
 
+  it('take into account non-capturing group', () => {
+    assert.deepEqual(match(/(?:head) (\d+) (tail)/, 'head 22 tail'), [
+      22,
+      'tail',
+    ])
+  })
+
   it('returns null when there is no match', () => {
     assert.equal(match(/hello/, 'world'), null)
   })
--- a/cucumber-expressions/javascript/test/regular_expression_test.js
+++ b/cucumber-expressions/javascript/test/regular_expression_test.js
@@ -54,7 +54,7 @@ describe(RegularExpression.name, () => {
     assert.deepEqual(match(/(.*)/, '-1.22', [s => parseFloat(s)]), [-1.22])
   })
 
-  it('take into account non-capturing group', () => {
+  it('takes into account non-capturing group', () => {
     assert.deepEqual(match(/(?:head) (\d+) (tail)/, 'head 22 tail'), [
       22,
       'tail',


yaronassa:
This is still problematic - the new regexp skips capturing groups that *contain* non-capturing groups (e.g. `(demo(?:nstration)?)`).

So, for example, the following mocha test will fail:
```javascript
it("matches parameters with custom parameter type using optional capture group", () => {
        parameterTypeRegistry = new ParameterTypeRegistry()
        parameterTypeRegistry.defineParameterType(new ParameterType(
            'color',
            Color,
            [/red|blue|yellow/, /(?:dark|light) (?:red|blue|yellow)/],
            s => new Color(s)
        ))
        const expression = new RegularExpression(/I have a ((?:dark|light) (?:red|blue|yellow)) ball/, [], parameterTypeRegistry)
        const transformedValue = expression.match("I have a dark red ball")[0].transformedValue
        assert.equal(transformedValue.name, "dark red")
    })
```

Basically we need to recursively process the groups, which is not something JS regular expressions were meant to do.
I don't see any elegant solution but to go through `regexp.source` by character and manage a group stack.

yaronassa:
I propose this fix, which still captures groups with inner non-capturing groups

https://github.com/yaronassa/cucumber-expressions-javascript/commit/444dbdc680887457a36b11d77e88fa75d2f356a2

kAworu:
@yaronassa thank you for your answer and detailed explanation.

I fully agree that my patch is not flawless and only adress a particular case (described in the test added). I am aware, as you described well and illustrated with another issue, that the more broad challenge of parsing a regexp can not be done by regexp.

In my humble opinion, the fix you propose is a step in the over-engineering way. Surely it handle correctly recursivity (as you pointed out already) and positive look-ahead, but it misses (at least) double backslash before a capturing group:
```diff
diff --git a/test/regular_expression_test.js b/test/regular_expression_test.js
index acee347..1cfe3d9 100644
--- a/test/regular_expression_test.js
+++ b/test/regular_expression_test.js
@@ -38,6 +38,10 @@ describe(RegularExpression.name, () => {
     assert.deepEqual(match(/(\d+)/, "22"), [22])
   })
 
+  it('handle double backslash', () => {
+      assert.deepEqual(match(/\\(\d+) (tail)/, '\\22 tail'), [22, 'tail'])
+  })
+
   it("transforms float without integer part", () => {
     assert.deepEqual(match(/(.*)/, ".22", ['float']), [0.22])
   })
```
and maybe other cases that I am missing.

To me, the cleanest solution is to simply forgo transformations on regexp and let the user handle it. Less code, no bugs. I *think* the upside of doing this dance is that it make migration from cucumber expr to (or from) regexp easy, so it's a trade off.

My two cents anyway, I will be happy of a resolution of #211 regardless of the direction chosen by the cucumber team.

yaronassa:
Thanks for the thoughtful response!
I might have jumped the gun and not though of a thorough enough solution, and as you pointed out, if it's needed at this phase.

In the meanwhile, I've found http://xregexp.com/ which implements recursive regexp processing in JS, and could probably be more elegantly used if a more broad solution will be the way to go.

Thanks

kAworu:
For to sake of completness, using a full blown regexp parser (like https://github.com/JexCheng/regulex) might also be considered.

aslakhellesoy:
I've closed #211 since it was a dupe of #131.

Let's not worry about nesting in this issue. As pointed out, that's too hard to handle with regexp. Regulex would have been a nice solution if it weren't for the fact that we need feature parity in all implementations (including Ruby and Java). If we really want to tackle the issue with nesting I think it would be best to implement a custom parser, as outlined in http://matt.might.net/articles/parsing-regex-with-recursive-descent/. I'm not convinced it's worth the effort.

aslakhellesoy:
Closed in favour of #220. See #211 for details. Thanks!

PR #253, opened 2017-08-22
Backward support for escaped parens in steps

jaysonesmith:
## Summary

When creating step def snippets from features that have parenthesis, cucumber can sometimes escape them with backslashes and this breaks the current 3.0 handling of step defs.

## Details

* Remake of [pr2](https://github.com/cucumber/cucumber-expressions-ruby/pull/2) from the expressions repo.

### Bugfix

* When generating step def snippets, Cucumber escapes parens with backslashes and cucumber-expressions-ruby was catching erroneously on that.
* Error: `Expression /^Across the line\(s\)$/ has 0 capture groups ([]), but there were 1 parameter types (["s\\"]) (Cucumber::CucumberExpressions::CucumberExpressionError)`
* [Full stack error here](https://gist.github.com/jaysonesmith/3b17b3dd830cd4b83157a865e82efadf)
* I added a test to the `regular_expression_spec.rb` file and all tests pass successfully. Can someone let me know if my test is configured correctly and in the right place, please?

### Example from my own project
* Step of: `Given I want to request "<caption_count>" caption(s)` in my feature
* Generated snippet from 2.4.0: `Given(/^I want to request "([^"]*)" caption\(s\)$/) do`

## Types of changes

## Checklist:

### DIFF
--- a/cucumber-expressions/ruby/spec/cucumber/cucumber_expressions/regular_expression_spec.rb
+++ b/cucumber-expressions/ruby/spec/cucumber/cucumber_expressions/regular_expression_spec.rb
@@ -46,6 +46,10 @@ module Cucumber
           "I can cancel the 1st slide upload")
         ).to eq(["I", "can", 1, "slide"])
       end
+      
+      it "ignores escaped 2.x parenthesis" do
+        expect( match(/Across the line\(s\)/, 'Across the line\(s\)') ).to be_nil
+      end
 
       it "exposes source and regexp" do
         regexp = /I have (\d+) cukes? in my (\+) now/
--- a/cucumber-expressions/ruby/spec/cucumber/cucumber_expressions/regular_expression_spec.rb
+++ b/cucumber-expressions/ruby/spec/cucumber/cucumber_expressions/regular_expression_spec.rb
@@ -48,7 +48,7 @@ module Cucumber
       end
       
       it "ignores escaped 2.x parenthesis" do
-        expect( match(/Across the line\(s\)/, 'Across the line\(s\)') ).to be_nil
+        expect( match(/Across the line\(s\)/, 'Across the line(s)') ).to eq([])
       end
 
       it "exposes source and regexp" do
--- a/cucumber-expressions/ruby/lib/cucumber/cucumber_expressions/regular_expression.rb
+++ b/cucumber-expressions/ruby/lib/cucumber/cucumber_expressions/regular_expression.rb
@@ -5,7 +5,7 @@ require 'cucumber/cucumber_expressions/tree_regexp'
 module Cucumber
   module CucumberExpressions
     class RegularExpression
-      CAPTURE_GROUP_PATTERN = /\((?!\?:)([^(]+)\)/
+      CAPTURE_GROUP_PATTERN = /(?<!\\)\((?!\?:)([^(]+)\)/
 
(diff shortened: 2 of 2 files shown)


jaysonesmith:
Before this PR can be completed, similar work will need to be done in the Java and Javascript portions of expressions. I had tried my hand with the Java code but was unable to implement things appropriately. 

I'm unsure how the javascript issue will need to be handled as JS doesn't currently support lookbehinds, which is what I've implemented in the Ruby version. Any ideas here? I'm not as good with JS either, so I'd like to request additional assistance if possible.

Thanks!

aslakhellesoy:
I don't think the test you added illustrates the bug as it passes even without your fix.

cucumber-expressions/ruby/spec/cucumber/cucumber_expressions/regular_expression_spec.rb
```diff
@@ -46,6 +46,10 @@ module CucumberExpressions
           "I can cancel the 1st slide upload")
         ).to eq(["I", "can", 1, "slide"])
       end
+      
+      it "ignores escaped 2.x parenthesis" do
+        expect( match(/Across the line\(s\)/, 'Across the line\(s\)') ).to be_nil
```
aslakhellesoy: This should be:

```ruby
expect( match(/Across the line\(s\)/, 'Across the line(s)') ).to eq([])
```

That will reproduce the bug when the `CAPTURE_GROUP_PATTERN` is as before. Can you push another commit with that fix please?

I'll have a think about what to do with negative lookahead - ideally we'll use the same regexps in all implementations...

jaysonesmith: I struggled it for a bit, but thank you for the guidance! I definitely will.

I'm not fully certain about the regex part, perhaps another negative lookahead instead?

jaysonesmith:
@aslakhellesoy Updated to utilize your fix. Thanks again!

aslakhellesoy:
The root cause of the bug is that we're parsing regular expressions in two inconsistent ways. We're parsing with `RegExp`, but also in `TreeRegexp`.  Parsing a regular expression is tricky to do with `RegExp`, which is why `TreeRegexp` has its own character-based parser.

I think a more robust fix is to have one way to parse regular expressions (the `TreeRegexp` one)

`RegularExpression` could ask its `TreeRegexp` to return a list of `parameter_type_regexp` rather than relying on `CAPTURE_GROUP_PATTERN`.
We'd have to initialise `GroupBuilder` with the text that is between parenthesis in capture groups, so the parsing logic in `TreeRegexp` needs to pass that along.

I think that involves adding a stack of indices next to the existing stack to keep track of where a capture group starts. If that sounds like gobbeligook don't worry - this is a little complex :-)

Feel free to have a go at it if you know what I mean. If not I'll happily do it.

jaysonesmith:
Hah! It definitely is a bit out there, but definitely stuff I should be learning. Perhaps I'll try implementing a more basic version in a separate project or something. What type of timeline do you have in mind for this fix? (That'll dictate whether I bow out of this.)

It was neat stepping through the `TreeRegexp` process and gaining an understanding of how it looked for various matches. 👍 to that!

aslakhellesoy:
Superseded by #254 - but I kept your test @jaysonesmith!

Issue #272, opened 2017-09-26
Gherkin: cannot build language update on Ubuntu 16.04


RomkeVdMeulen:
I'm trying to build Gherking master on Ubuntu 16.04 after making an update to `gherkin-languages.json`. I followed the [instructions](https://github.com/cucumber/cucumber/blob/master/gherkin/CONTRIBUTING.md#adding-or-updating-an-i18n-language), making sure the prerequisite tools are installed. The instructions seem to be out of date however, since one of the commands launched during `make` uses mono. I installed mono, but the command still fails. Here is what I'm seeing on my console:

```
$ make
cd python && make default
make[1]: Entering directory '/home/romke/workspace/cucumber/gherkin/python'
mono berp/berp.exe -g gherkin.berp -t gherkin-python.razor -o gherkin/parser.py
Can't find custom attr constructor image: /home/romke/workspace/cucumber/gherkin/python/berp/CommandLine.dll mtoken: 0x0a000014
Unhandled loader error: 8, (null) System.Core, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089 (null)
* Assertion: should not be reached at loader.c:298
```

RomkeVdMeulen:
I managed to complete the build by installing `mono-complete`, `pip`, `python-nose` and `golang`, and by installing .NET core 1.0.1 using [these instruction](https://github.com/dotnet/core/blob/master/release-notes/download-archives/1.1.2-download.md). However, there may be additional dependencies I'm not aware of already present on my system.

However, a lot of files were removed with `make clean` that weren't replaced once `make` completed. Here's my git status:

```
        deleted:    ../c/include/rule_type.h
        deleted:    ../c/src/dialect.c
        deleted:    ../c/src/parser.c
        modified:   ../gherkin-languages.json
        deleted:    ../java/src/main/java/gherkin/Parser.java
        deleted:    ../javascript/dist/gherkin.js
        deleted:    ../javascript/dist/gherkin.min.js
        deleted:    ../javascript/lib/gherkin/parser.js
        deleted:    ../objective-c/Gherkin/GHParser.h
        deleted:    ../objective-c/Gherkin/GHParser.m
        deleted:    ../perl/lib/Gherkin/Generated/Languages.pm
        deleted:    ../perl/lib/Gherkin/Generated/Parser.pm
        deleted:    ../ruby/lib/gherkin/parser.rb
```

RomkeVdMeulen:
[...]

PR #273, opened 2017-09-26
Gherkin (i18n): add additional Dutch keywords for Given.

RomkeVdMeulen:
Although it is technically possible in Dutch to write sentences that start with `Gegeven` and aren't immediately followed by `dat`, such sentences are very unnatural. In practice, `Gegeven dat` is almost always used. However, the same is not true when using a conjunction like `En` or `Maar`. To prevent having to write step definitions with the word `dat` as optional, depending on whether the keyword `Gegeven` or `En / Maar` was used in this step, it would be much easier to match `dat` as part of the keyword.

The old keywords are preserved, to this change should be backwards-compatible.

## Types of changes

I've built Gherkin with the changed language definitions without problems, however see #272.

### DIFF
--- a/gherkin/gherkin-languages.json
+++ b/gherkin/gherkin-languages.json
@@ -2114,7 +2114,9 @@
     ],
     "given": [
       "* ",
+      "Gegeven dat ",
       "Gegeven ",
+      "Stel dat "
       "Stel "
     ],
     "name": "Dutch",


muggenhor:
1. This will unconditionally remove "dat" from the string passed to the step definition, as a result it isn't backwards compatible.
2. This isn't that different from "Given that", although I agree that common Dutch usage will have "dat" more often than English gets 'that". It's rarely mandatory though, just common.

RomkeVdMeulen:
@muggenhor 

1. "Gegeven" is still a keyword, so if you've got an existing codebase with steps starting with "dat", they should still match. Have to admit I didn't thoroughly test that though: I'll look into that.
2. There is a difference though: you can write a full sentence in English without "that", but in Dutch that wouldn't work. E.g.:

* Given I'm currently on page X
* And I'm currently on page X
* ~Gegeven ik ben op pagina X~
* ~Gegeven ik op pagina X ben~
* Gegeven **dat** ik op pagina X ben
* En ik op pagina X ben

RomkeVdMeulen:
Come to think of it: "Stel" doesn't seem to have this problem.

* Stel ik ben op pagina X
* En ik ben op pagina X
* Stel dat ik op pagina X ben
* En ik op pagina X ben

Both seem to work. I'd suggest still adding "Stel dat" (provided this is all backwards-compatible of course) to keep things consistent.

mpkorstanje:
While I think this would be an awesome addition I think this will break backwards compatibility. 
```
Situation | Sentence         | Expression | 
Current   | Gegeven dat iets | dat iets   | 
Proposed  | Gegeven dat iets | iets       | 
```
In the current situation the gherkin sentence "Gegeven dat iets" will result in the expression "dat iets" but in the proposed situation it will result in "iets". This means that the step definition `Gegeven("dat iets" ()->{})` no longer matches.

RomkeVdMeulen:
That is unfortunate. Is there any way to provide a good migration path to the new keywords?

aslakhellesoy:
Accorging to semver we'd have to make a new major release of Gherkin, and therefore Cucumber. That sounds un-pragmatic - this sounds fairly innocuous.

We could just explain in the release notes that people have to add `dat` to all their scenarios using `Gegeven`.

In any case, I don't think they can both coexist - if we add `Gegeven dat` as a keyword, we should remove `Gegeven`

RomkeVdMeulen:
I'm not sure how to proceed here. I for one would promote moving to `Gegeven dat`, however this is not a decision I can make for all Dutch Cucumber users.

aslakhellesoy:
Hmm - the French translation has a similar variant to what you're proposing: https://github.com/cucumber/cucumber/blob/4bfaac0e4cb2c8f14106cbd3a361566fcecdb516/gherkin/gherkin-languages.json#L1108-L1123

I don't recall anyone complaining about that. Can you add some tests that verify/falsify it's working both with and without the conjunction?

mpkorstanje:
Removing Gegeven in favour of Gegeven dat is not viable. "Gegeven dat" is
often but not always the start of a sentence. E.g. "Geveven een drukke dag".

We'd either have to keep "Gegeven" or keep "Gegeven" and add "Gegeven dat".

Removing "dat " from step defs seems a reasonable migration. But I think we
should try to come up with a few more example sentences to make sure they
work with all keywords.

RomkeVdMeulen:
@mpkorstanje I'm not sure this is exhaustive, but I'll give it a try.

As far as I can tell, there's three different possibilities for the sentence that follows a keyword:

1. No verb
* Gegeven een item X met status Y
* ~Gegeven dat een item X met status Y~
* ~Stel een item X met status Y~
* ~Stel dat een item X met status Y~
* En een item A met status B
2. Sub-clause (bijzin)
* ~Gegeven X de status Y heeft~
* Gegeven dat X de status Y heeft
* ~Stel X de status Y heeft~
* Stel dat X de status Y heeft
* En A de status B heeft
3. Main clause
* Stel X heeft de status Y
* ~Stel dat X heeft de status Y~
* ~Gegeven X heeft de status Y~
* ~Gegeven dat X heeft de status Y~
* En A heeft de status B

I suspect option 1 is hardly ever used. But if we want to continue to support it, like @mpkorstanje said, we'll have to keep keyword "Gegven". Option 2 is the most obvious, natural, and most often used, I think. It's debatable whether option 3 is correct, but it doesn't really fit natural language use. If we want to support option 3, keyword "Stel" also has to stay.

keetron:
In general, if you implement a feature and it will break existing use, it makes sense to either have really good motivation for it or make it backwards compatible. As discussed, backwards compatibility is not an option.

Your reason to do this is an opinion dressed as a fact. Please deliver more facts about actual use before moving to technical implementation.

mpkorstanje:
Mmh. You're right. That seems to exhaust all grammatical options.

>  I suspect option 1 is hardly ever used. But if we want to continue to support it, like @mpkorstanje said, we'll have to keep keyword "Gegeven". 

I'm seeing allot of sentences that follow the "Gegeven een ding met een eigenschap"  pattern. E.g.: "Gegeven een relatie met een partner", "Gegeven een affiliate in Rotterdam".

RomkeVdMeulen:
@keetron I'm not sure how to aquire hard data on usage. I tried some [Google fu](https://www.google.nl/search?q=site%3Agithub.com+intitle%3Afeature+gegeven&oq=site%3Agithub.com+intitle%3Afeature+gegeven) but didn't get any promising results. I agree unless the backward compatibility problem can be solved this change is probably more trouble than it is worth.

@mpkorstanje Supporting "Gegeven een ding met een eigenschap" isn't a problem, I think. We simply have to keep the "Gegeven" keyword. The problem is that any existing code written like `Given("dat een ding een eigenschap heeft")` wil break with this change.

RomkeVdMeulen:
@aslakhellesoy Interesting: it looks like the French keywords were [changed in a similar fashion two years ago](https://github.com/cucumber-attic/gherkin/commit/42e733e0609fdd0635122b9004662d1433ae105b). Perhaps French feature files weren't in widespread use yet?

brasmusson:
That French keyword change was introduced in a major release (4.0.0), https://github.com/cucumber-attic/gherkin/pull/134. If it was the only reason to increment the major version, I do not know for sure.

mpkorstanje:
@RomkeVdMeulen yes, this will be breaking change. If explained properly the impact will be minor I reckon.

RomkeVdMeulen:
We could wait with this change until a major version change happens for some other reason and then lift along on that.

keetron:
The size of the change will not make a difference. There will still be no backwards compatibility and no data to give insight into the amount of breakage. The only thing that a major change will do is allow you to say: "well, of course your tests that were always fine now break, this is a major change!"

RomkeVdMeulen:
@keetron I understand that this isn't an ideal scenario and if you have a better idea for how to implement this, I'd love to know. Or if you are saying this change simply cannot happen, I can live with that as well.

keetron:
The lack of a better idea from me, does not discount that breaking backwards compatibility is a bad idea (which is my main point).
You and @mpkorstanje figured out that there is no real backwards compatibility possible.

My main opinion is that this change is not important enough, not a change with sufficient impact, to break existing tests. If you and others who make actual contributions (unlike me), disagree and go forward with it, then I will have to live with it which is easy as all my Gherkin is in English anyway. It is just that this PR caught my interest, being a Dutch native speaker myself :smile:

mlvandijk:
@RomkeVdMeulen @mpkorstanje 
Personally, I don't see the point of introducing a breaking change for this.
It can easily be solved differently:

2. Sub-clause (bijzin)
Make the following change in the example:
Gegeven item X de status Y heeft => Gegeven item X heeft status Y

3. Main clause
Gegeven item X heeft de status Y

Additionally, when writing the step definitions, it would be possible to create a step definition that covers both "Gegeven dat iets" en "Gegeven iets" and does the same thing for both. It would be at the users discretion when to equate the two and when not to.

Imho we are using a certain syntax which I agree is limited and might not be as fluent as the full natural language. I personally don't think it is worth introducing a breaking change, when it can easily be solved differently by individual users.

aslakhellesoy:
Thanks for all the careful analysis everyone. I've decide not to merge this because I don't want to break backwards compatibility.

PR #282, opened 2017-10-06
Experimental CLI for tag expressions

aslakhellesoy:
## Summary

Command-line tool that prints the location of scenarios matching a certain tag expression

## Details

The tool reads a Cucumber event stream (produced by the `gherkin` command line tool) from `STDIN` and prints the locations (in "rerun.txt" format) to `STDOUT`

## Motivation and Context

This tool can be used as a preprocessor before running `cucumber` to get a list of scenarios to execute before running `cucumber`.

While `cucumber` can filter scenarios with the `--tags` option, it is sometimes useful to perform this filtering before running `cucumber` so that several `cucumber` processes can be spawned in parallel.

This can speed up the execution for slow scenarios.

## How Has This Been Tested?

Manually

## Types of changes

## Checklist:

## Caveats

When multiple `cucumber` processes are run in parallel, each process will produce an independent report. These reports will have to be combined in order to produce a single report, and this isn't trivial to do until there are reporters that are based on the Cucumber event protocol.

### DIFF
--- a/tag-expressions/ruby/bin/cucumber-tag-expressions
+++ b/tag-expressions/ruby/bin/cucumber-tag-expressions
@@ -1,8 +1,8 @@
 #!/usr/bin/env ruby
 #
 # This script reads NDJSON-encoded Cucumber events from STDIN and prints
-# the location of each pickle (scenario) matching a tag expression, If
-# no tag expression is specified, all pickle location is printed.
+# the location of each pickle (scenario) matching a tag expression. If
+# no tag expression is specified, all pickle locations are printed.
 #
 # Example:
 #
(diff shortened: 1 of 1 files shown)


jaysonesmith:
I really like this idea! I had thought about accomplishing the same thing via `-f json -d -o output.json`, parsing output.json for each tag, grabbing the needed data, and then splitting things out that way. Having this built in would be really helpful with that type of behavior!

aslakhellesoy:
This PR was created in response to [this slack discussion](https://cucumberbdd.slack.com/archives/C62GZFLLT/p1507313771000026). I think that was you @fordprior?

PR #291, opened 2017-10-21
add datatable/java

mpkorstanje:
DataTable is a simple data structure that allows the use and transformation of Gherkin data tables in cucumber without directly using the table from Gherkin. 

It also features a type registry similar to that of cucumber expressions to assist the transformation of a data table into other types and data structure. I believe that it currently can handle the transformation of a data table to any reasonable combination of lists, lists of lists, lists of maps, maps, maps of lists and maps of maps.

 It does not (yet) feature the automagic conversion currently present in Cucumber JVM. This is a feature that may be added at a later date. 

For more usage see the `datatable/README.md` and `datatable/java/README.md`

This PR is needed to support the implementation of [Cucumber Expressions in Cucumber JVM](https://github.com/cucumber/cucumber-jvm/pull/1248). As such it also includes several fixes for `cucumber-expressions`.


cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/CucumberExpression.java
```diff
@@ -68,21 +68,16 @@ private String buildCaptureRegexp(List<String> regexps) {
     }
 
     @Override
-    public List<Argument<?>> match(String text, final DataTable argument) {
+    public List<Argument<?>> match(String text, List<List<String>> tableArgument) {
```
mpkorstanje: I have changed the to match(String text, List<List<String>> tableArgument) so callers don't have to provide their own data table with converter which needs a parameter type registry instance. This simplifies the logistics of using Expressions.

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/ExpressionFactory.java
```diff
@@ -18,20 +22,42 @@
...
         this.parameterTypeRegistry = parameterTypeRegistry;
+        this.tableConverter = new ParameterTypeRegistryTableConverter(parameterTypeRegistry);
     }
 
-    public Expression createExpression(String expressionString, Type tableType) {
-        return createExpression(expressionString, parameterTypeRegistry.lookupTableTypeByType(tableType));
+    public Expression createExpression(String expressionString, final Type tableType) {
+        if (tableType == null) throw new CucumberExpressionException("tableType can not be null");
+
+        return createExpression(expressionString, new Transformer<List<List<String>>, Object>() {
```
mpkorstanje: We can extract the `Transformer` and leave its creation to the caller. If we also provide the default table transform we  can cut the dependency between datatable and cucumber-expressions.

mpkorstanje:
> It's potentially 3 libs, but it makes sense to spike it all in the same project before extracting.

I've opted to use composition instead of extension. Javas type system isn't quite powerful enough to create a sufficiently generic parameter type. So there are now two separate type registries and a third one to tie them together. This makes things allot simpler.

We now have a concept of a StepExpression which is a CucumberExpression + an optional tableOrDocStringType.

cucumber-expressions/java/src/main/java/io/cucumber/java/TypeRegistryTableConverter.java
```diff
@@ -0,0 +1,214 @@
...
+        List<K> keys = new ArrayList<>();
+        for (String keyString : keyStrings) {
+            keys.add(keyConverter.transform(singletonList(keyString)));
+        }
+        List<List<String>> valueRows = dataTable.rows(1);
+        for (List<String> valueRow : valueRows) {
+            Map<K, V> map = new LinkedHashMap<>();
+            int i = 0;
+            for (String cell : valueRow) {
+                map.put(keys.get(i), valueConverter.transform(singletonList(cell)));
```
mpkorstanje: This is wrong. We should use the match method first.

mpkorstanje:
I've come to the conclusion that adding a default object mapper will run us into the same problems and complexity as we've had with XStream.  It binds both cucumber and and its user to a specific mapper and requires that our users follow the conventions and ceremonies of that mapper. This makes it harder to use business objects objects in cucumber.

On other other hand by  moving the object mapper into the parameter type configuration the end user can decide which mapper to use and use one that works in conjunction with his own business objects. So if the user is using jackson in his domain he can use jackson in cucumber. If the user is using something else, he can use that.

This adds a little verbosity but with java 8 it already folds up nicely:

```
        parameterTypeRegistry.defineDataTableType(DataTableType.tableOf("groceries", ShoppingStepdefs.Grocery.class, new TableRowTransformer<ShoppingStepdefs.Grocery>() {
            @Override
            public ShoppingStepdefs.Grocery transform(Map<String, String> row) {
                return new ObjectMapper().convertValue(row, ShoppingStepdefs.Grocery.class);
            }
        }));
```

mpkorstanje:
Currently our possible set of transformations for data table is not complete. The current implementation does not support the transform of datatable rows to map of lists or the transform to map of maps.

```gherkin
Scenario: Datatable to single object
  Given the following chessboard:
    | |1|2|3|
    |A|♘| |♝|
    |B| | | |
    |C| |♝| |
  Then I have a board with one knight and two bishops

Scenario: Datatable rows to list of object
  Given the following groceries:
    | name  | price |
    | milk  | 9     |
    | bread | 7     |
    | soap  | 5     |
  Then I have:
  """
  [
    {name: "milk", price: "9"}, 
    {name: "bread", price: "7"}, 
    {name: "soap", price: "5"}
  ]
  """

Scenario: Datatable rows to map
  Given the priorities for groceries:
    | low    |  milk | 
    | medium | bread |
    | high   | soap  | 
  Then I have:
  """
  {
    low:  "milk", 
    medium: "bread", 
    high: "soap"
  ]
  """

Scenario: Datatable rows to map of maps
  Given the following groceries:
    |        | name  | price |
    | low    |  milk | 9     |
    | medium | bread | 7     |
    | high   | soap  | 5     |
  Then I have:
  """
  {
    low:  {name: "milk", price: "9"}, 
    medium: {name: "bread", price: "7"}, 
    high: {name: "soap", price: "5"}
  ]
  """

Scenario: Datatable rows to map of lists
  Given the following groceries:
    | jones   |  milk | bread  | soap |
    | smith   | bread | milk   |      | 
    | jackson | soap  |        |      |
  Then I have:
  """
  {
    jones:  ["milk", "bread", "soap" ],
    smith:   ["bread", "milk" ],
    jackson: ["soap" ], 
  ]
  """
```

mpkorstanje:
To make development a bit easier I've moved the datatable code into the cucumber-expressions branch of cucumber-jvm. 

In hindsight the datatable to list objects scenario is a bit silly.

aslakhellesoy:
I want to have a cross-platform implementation of data tables, just like cucumber-expressions and tag-expressions. Part of that is a shared test suite and synchronised releases.

This becomes harder if it’s moved to
Cucumber-Jvm.

What became easier by moving it?

mpkorstanje:
Refactoring and integrating with the existing code. IDEA can't refactor over multiple projects.

As I hadn't quite untangled the DataTable from the StepExpression from the StepDefinition yet I ended up in a cycle of refactor-compile-fix-the-other-project. By putting the data tables and step expression into a single project the tooling works and it the refactorings happen instantly.

Once cucumber expressions is done but before it is merged I reckon we can extract data tables out again.

mpkorstanje:
@aslakhellesoy I've put the code is back in. The good parts to review are `README.md`, `DataTable.java` and `DataTableTypeRegistryTableConverter`.  The other code came from cucumber-jvm and has only been lightly touched. I assume you're familiar with it.

PR #292, opened 2017-10-22
gherkin: Pass the content type of a docstring down into its pickle string form… 

rjwittams:
… for use in formatters.

## Summary

Pass the content type of a docstring down into its pickle string form. 
This allows formatters to use it. 
This used to work in Cucumber 1.2.5. 
A PR for cucumber-jvm accompanies this. 

## Motivation and Context

https://github.com/cucumber/cucumber-jvm/issues/1263
As suggested here by mpkorstanje , I am doing a PR against the mono repo.

## How Has This Been Tested?

Added in a unit test - not sure if it should be moved.
Also not sure that Gson-ing pickles is expected - given camelCase vs snake_case. 
Ran against my updated cucumber-jvm. Added a unit test there. 
Verified that my json report processor which works with 1.2.5 picks up the docstrings.

Ran all tests in cucumber-jvm. Examples: Pax-Exam : Calculator Test is broken before and after the change due to a missing deployment I think.

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue).

Adds a field to PickleString. Not sure if that gets serialised within this project. 
Maintains the old constructor signature to avoid compile breaks. Could be deprecated.

## Checklist:

Not sure which if any docs should be updated - I think 1.2.5 preserved docs will now be correct.


rjwittams:
https://github.com/cucumber/cucumber-jvm/pull/1265 <- this is the PR for cucumber-jvm

rjwittams:
Updated the acceptance tests data. Used snake_case as I assumed that was the convention.

gherkin/java/src/main/java/gherkin/pickles/Compiler.java
```diff
@@ -152,7 +152,8 @@ private void compileScenarioOutline(List<Pickle> pickles, List<PickleStep> backg
             result.add(
                     new PickleString(
                             pickleLocation(ds.getLocation()),
-                            interpolate(ds.getContent(), variableCells, valueCells)
+                            interpolate(ds.getContent(), variableCells, valueCells),
+                            ds.getContentType()
```
mpkorstanje: Shouldn't this also be interpolated?

rjwittams:
I hadn’t quite understood its significance. I see it’s for scenario outlines with examples. Can do so. What I’m not sure of if is the docstring content type can be used within a step. Eg can you use a PickleString as an argument in a step and get the whole thing rather than just its contents? It seems that it could be useful if not. Also may be more useful with the Transformers rework - could we discriminate transformers by content type.

Bit of a cross repo conversation ...

mpkorstanje:
We're on slack btw. https://cucumberbdd-slack-invite.herokuapp.com/ the #commiters-jvm channel.

gherkin/java/testdata/good/docstrings.feature.pickles.ndjson
```diff
@@ -1 +1 @@
...
...:[{"column":11,"line":4}],"text":"a simple DocString"},{"arguments":[{"content":"<foo>\n  <bar />\n</foo>","content_type":"xml","location":{"column":7,"line":12}}],"locations":[{"column":11,"line":11}],"text":"a DocString with content type"},{"arguments":[{"content":"wrongly indented line","location":{"column":7,"line":18}}],"locations":[{"column":9,"line":17}],"text":"a DocString with wrong indentation"},{"arguments":[{"content":"first line\nsecond line","location":{"column":7,"line":22}}],"locations":[{"column":9,"line":21}],"text":"a DocString with alternative separator"},{"arguments":[{"content":"first line\n\"\"\"\nthird line","location":{"column":7,"line":27}}],"locations":[{"column":9,"line":26}],"text":"a DocString with normal separator inside"},{"arguments":[{"content":"first line\n```\nthird line","location":{"column":7,"line":33}}],"locations":[{"column":9,"line":32}],"text":"a DocString with alternative separator inside"},{"arguments":[{"content":"first line\n\"\"\"\nthird line","location":{"column":7,"line":39}}],"locations":[{"column":9,"line":38}],"text":"a DocString with escaped separator inside"}],"tags":[]},"type":"pickle","uri":"testdata/good/docstrings.feature"}
```
brasmusson: This file should not be edited directly, it is [gherkin/testdata/good/docstrings.feature.pickles.ndjson](https://github.com/cucumber/cucumber/blob/master/gherkin/testdata/good/docstrings.feature.pickles.ndjson) which is the source file that is pushed down to each language version of the gherkin library (using [rsync_files](https://github.com/cucumber/cucumber/blob/master/scripts/functions.sh#L54-L63)). Since we use the same testdata for acceptance tests for all language version for the gherkin library, this pull-request cannot be merged until all 8 language versions have been updated.

rjwittams:
Do you expect each contributor to fix all 8 languages, including those they’ve not used?

brasmusson:
@rjwittams Many people can contribute to a pull-request, so "each contributor" does not need to fix all languages. However the acceptance test data should not diverge between language versions.

brasmusson:
> And while the content type is not passed onto the step definition, ...

The Pickles are primarily for executing the test cases from each feature file, so if the content type is not needed for that, maybe the content type should not be added to the Pickles data.

A formatter can still access the content type, since the formatter has access to the complete feature file content.

rjwittams:
Ie each reporter should reparse the feature file? Does that seem right?

What is the purpose of the content type? Given that it is not in the Pickle, it can not be used in the test execution. However, that is an arbitrary choice - argument transformers could make use of it. In fact I was planning to do this.

If there is no point in continuing let me know. It is however a clear regression in Cucumber JVM vs 1.2.5 .

brasmusson:
> argument transformers could make use of it

This is a real argument for including the content type in the Pickle. That it is easier for the formatters is not. For instance the step keyword is not in the Pickle, even though practically every formatter needs to use it.

rjwittams:
Also, as mentioned above, it may be useful (and consistent) for it to be transformed by interpolation.

gherkin/javascript/lib/gherkin/parser.js
```diff
@@ -48,9 +48,6 @@ module.exports = function Parser(builder) {
     if(typeof tokenScanner == 'string') {
       tokenScanner = new TokenScanner(tokenScanner);
     }
-    if(typeof tokenMatcher == 'string') {
```
rjwittams: Not sure why this change was made in the generated parser - I assume it was missed out. I assume it is benign.

rjwittams:
The build seemed to fail due to yarn timing out. Can it be restarted?
On binary and source compatibility: 
  C and go I did not try, as a struct size needs to change.
 Java and C# should have a compatible PickleString constructor defaulting content type to null.
 The rest are dynamic I think.

mlvandijk:
@rjwittams Restarted the build. It fails with the following message:

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

aslakhellesoy:
Hi @rjwittams,

Aslak Hellesøy
Creator of Cucumber

PR #293, opened 2017-10-23
gherkin: Retain leading and trailing newlines in data table cells

### LINKED ISSUE #290: gherkin: Leading and trailing "\n" are trimmed in DataTable cells
After upgrading from cucumber-jvm 1.2.4 to 2.0.1, some of our tests failed.  
We have steps like this:
```
And the text should be
|something\n|
```
The important thing here is the "**\n**" at the end.  
With the older version, the "\n" was retained, but the new one trims it.  
I tracked down the issue and I think it's caused by an old change here: https://github.com/cucumber/gherkin-java/blob/v4.1.3/src/main/java/gherkin/GherkinLine.java#L67  
It's parsing the line from the feature file and this part of the code handle the new line character perfectly: https://github.com/cucumber/gherkin-java/blob/v4.1.3/src/main/java/gherkin/GherkinLine.java#L90  
But at the end of the line(at the "|" char) it trims the string: https://github.com/cucumber/gherkin-java/blob/v4.1.3/src/main/java/gherkin/GherkinLine.java#L86  
It doesn't make too much sense to me, because the original string as the variable name suggests(*trimmedLineText*) is already trimmed, so I think everything inside the "|" characters should remain the same, with all of the white space.

brasmusson:
## Summary

Do not trim leading and trailing newlines, that is `"\n"` converted to `newline`, in data table cells.

## Details

Trimming leading and trailing newlines in data table cells is a regression compared to Gherkin v2.

- [ ] C
- [ ] .NET
- [ ] Go
- [ ] Java
- [ ] Javascript
- [ ] Objective-C
- [ ] Ruby
- [ ] Perl
- [X] Python

## Motivation and Context

Fixes #290.

## How Has This Been Tested?

The acceptance test has been updated to verify this behavior.

## Types of changes

## Checklist:


aslakhellesoy:
This has gone stale in the light of #425. It would be cool to have this implemented in the Gherkin-Go, but that should be done in a new PR.

PR #313, opened 2017-12-06
gherkin java: Extended the languages list to include a column with the language name and a column with the native name

tsundberg:
## Summary

I updated the language list to include not just the language code but also the language name in English as well as the native language name.

## Details

The list of languages looked like this:

```
af
am
ar
ast
az
bg
…
```

There are 73 languages supported. It isn't easy to know which language is which.

I extended the list to look like this:

```
af            Afrikaans               Afrikaans
am            Armenian                հայերեն
ar            Arabic                  العربية
ast           Asturian                asturianu
az            Azerbaijani             Azərbaycanca
bg            Bulgarian               български
…
```

## Motivation and Context

This will make it easier to understand what language a language code is supporting.

## How Has This Been Tested?

I added a unit test that verifies that the list contains one row 

```
sv            Swedish                 Svenska
```

## Types of changes

## Checklist:

### DIFF
--- a/gherkin/java/src/main/java/gherkin/GherkinDialect.java
+++ b/gherkin/java/src/main/java/gherkin/GherkinDialect.java
@@ -6,20 +6,22 @@ import java.util.Map;
 import java.util.Objects;
 
 public class GherkinDialect {
-    private final Map<String, List<String>> keywords;
+    private final Map<String, Object> dialect;
     private String language;
 
-    public GherkinDialect(String language, Map<String, List<String>> keywords) {
+    public GherkinDialect(String language, Map<String, Object> dialect) {
         this.language = language;
-        this.keywords = keywords;
+        this.dialect = dialect;
     }
 
+    @SuppressWarnings("unchecked")
     public List<String> getFeatureKeywords() {
-        return keywords.get("feature");
+        return (List<String>) dialect.get("feature");
     }
 
+    @SuppressWarnings("unchecked")
     public List<String> getScenarioKeywords() {
-        return keywords.get("scenario");
+        return (List<String>) dialect.get("scenario");
     }
 
     public List<String> getStepKeywords() {
@@ -32,36 +34,44 @@ public class GherkinDialect {
         return result;
     }
 
+    @SuppressWarnings("unchecked")
     public List<String> getBackgroundKeywords() {
-        return keywords.get("background");
+        return (List<String>) dialect.get("background");
     }
 
+    @SuppressWarnings("unchecked")
     public List<String> getScenarioOutlineKeywords() {
-        return keywords.get("scenarioOutline");
+        return (List<String>) dialect.get("scenarioOutline");
     }
 
+    @SuppressWarnings("unchecked")
     public List<String> getExamplesKeywords() {
-        return keywords.get("examples");
+        return (List<String>) dialect.get("examples");
     }
 
+    @SuppressWarnings("unchecked")
     public List<String> getGivenKeywords() {
-        return keywords.get("given");
+        return (List<String>) dialect.get("given");
     }
 
+    @SuppressWarnings("unchecked")
     public List<String> getWhenKeywords() {
-        return keywords.get("when");
+        return (List<String>) dialect.get("when");
     }
 
+    @SuppressWarnings("unchecked")
     public List<String> getThenKeywords() {
-        return keywords.get("then");
+        return (List<String>) dialect.get("then");
     }
 
+    @SuppressWarnings("unchecked")
     public List<String> getAndKeywords() {
-        return keywords.get("and");
+        return (List<String>) dialect.get("and");
     }
 
+    @SuppressWarnings("unchecked")
     public List<String> getButKeywords() {
-        return keywords.get("but");
+        return (List<String>) dialect.get("but");
     }
 
     public String getLanguage() {
@@ -69,11 +79,11 @@ public class GherkinDialect {
     }
 
     public String getName() {
-        return (String) (Object) keywords.get("name");
+        return (String) dialect.get("name");
     }
 
     public String getNativeName() {
-        return (String) (Object) keywords.get("native");
+        return (String) dialect.get("native");
     }
 
     @Override
@@ -81,12 +91,12 @@ public class GherkinDialect {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
         GherkinDialect that = (GherkinDialect) o;
-        return Objects.equals(keywords, that.keywords) &&
+        return Objects.equals(dialect, that.dialect) &&
                 Objects.equals(language, that.language);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(keywords, language);
+        return Objects.hash(dialect, language);
     }
 }
--- a/gherkin/java/src/main/java/gherkin/GherkinDialect.java
+++ b/gherkin/java/src/main/java/gherkin/GherkinDialect.java
@@ -3,6 +3,7 @@ package gherkin;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 public class GherkinDialect {
     private final Map<String, List<String>> keywords;
@@ -65,5 +66,27 @@ public class GherkinDialect {
 
     public String getLanguage() {
         return language;
+    }
+
+    public String getName() {
+        return (String) (Object) keywords.get("name");
+    }
+
+    public String getNativeName() {
+        return (String) (Object) keywords.get("native");
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        GherkinDialect that = (GherkinDialect) o;
+        return Objects.equals(keywords, that.keywords) &&
+                Objects.equals(language, that.language);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(keywords, language);
     }
 }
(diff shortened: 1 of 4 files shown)


brasmusson:
I'm not sure that this logic belong in the Gherkin library. Cucumber-Ruby already implements this logic locally ([right here](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/cli/options.rb#L553-L557)). If this logic belong int the Gherkin library, then all Gherkin language version should implement this logic.

tsundberg:
Unfortunately, the Gherkin library only exposes a String with the language. If it exposed something richer that include language code, name, and native name then I would agree.

Would it be a better idea to expose a list of Language objects that holds all values necessary for creating the list? That would move most of the logic to Cucumber and possibly make better sense.

brasmusson:
The return value of `getLanguages()` is a list of things (which happens to be strings - the exact strings that should be but in a feature file to specify the language of the feature file), which can be passed as arguments to [`getDialect()`](https://github.com/cucumber/cucumber/blob/master/gherkin/java/src/main/java/gherkin/GherkinDialectProvider.java#L43) to get a `GherkinDialect` object. With this change the semantic of `getLanguages()` changes from returning a list of something that can be used to pass to `getDialect()` to something useful to display to a human user. That makes this change a non-backward compatible change. It would be possible to add a new method `getLanguageListForDisplay()`, but I still question if such a method belong in the Gherkin library. Shouldn't Cucumber-JVM instead use the `getLanguages()` output to iterate over the `GherkinDialect` object to construct the list to display to the user (as Cucumber-Ruby does).

aslakhellesoy:
I think it would be better to expose a `public List<GherkinDialect> getDialects()` method and leave it to the consumer to decide how to format them. We'd probably also have to add a getter to get the native translation of each dialect.

What do you think @tsundberg ?

aslakhellesoy:
If we do this we should also deprecate `List<String> getLanguages()`.

gherkin/java/src/test/java/gherkin/GherkinDialectProviderTest.java
```diff
@@ -2,13 +2,37 @@
...
     }
+
+    @Test
+    public void should_include_swedish_in_languages() {
+        GherkinDialectProvider gherkinDialectProvider = new GherkinDialectProvider();
+        List<String> actual = gherkinDialectProvider.getLanguages();
+
+        String expected = getExpected(gherkinDialectProvider);
+
+        assertTrue("Swedish with explanation should be available", actual.contains(expected));
```
mpkorstanje: I think that `assertThat("Swedish with explanation should be available", actual, CoreMatchers.containsString(expected))` is a bit more readable and provides better diagnostics when the string doesn't contain the expected value. 

Unfortunately it looks like we're using `mockito-all` in this project which bundles a version of `hamcrest-core` that doesn't have `containsString`. 

I'll add a commit to resolve this

mpkorstanje: Looking at the conversation this may not be relevant any more.

tsundberg:
@brasmusson I agree with your concern about semantics and backwards compatibility.

@aslakhellesoy I agree, exposing `public List<GherkinDialect> getDialects()` and let the consumer decide on it's usage is better. That was pretty much what I had in mind when talking about a richer object.

Deprecating  `List<String> getLanguages()` in the process seems reasonable.

I'll fix that.

tsundberg:
As suggested, I introduced a method `List<GherkinDialect> getDialects()` that exposes all dialects. 
I added getters for name and native name on `GherkinDialect` to allow a consumer to extract the values and display them as needed.

I also deprecated `List<String> getLanguages()` and recomend to use `List<GherkinDialect> getDialects()` in its Javadoc.

This allows me to implement the desired list with language, name and native name in Cucumber-JVM instead.

gherkin/java/src/main/java/gherkin/GherkinDialect.java
```diff
@@ -66,4 +67,26 @@ public GherkinDialect(String language, Map<String, List<String>> keywords) {
     public String getLanguage() {
         return language;
     }
+
+    public String getName() {
+        return (String) (Object) keywords.get("name");
```
aslakhellesoy: I haven't seen this kind of double casting before. Can you explain?

tsundberg: Not really. I haven't seen it myself before either.

I expected that the result from `keywords.get("name");` would be a list. It turns out that it isn't. It's a String. But as it is declared as a `List<String>`, the compiler will not allow me to cast it to a `String`. My first attempt was divided on two lines and casts. Then I inlined the result and ended up here. I was puzzled for a while and then I gave up. I blaim the internals of GSON.

The attached screenshot is from when I looked at af with a debugger.


screen shot 2017-12-07 at 10 10 09

tsundberg:
The test described above is obsolete. There are three unit tests that verifies
* the returned list contains the dialect Swedish
* it is possible to get the name Swedish from the dialect
* it is possible to get the native name Svenska from the dialect

tsundberg:
Is there anything holding this PR back? It has been stale for some six week by now.

gherkin/java/src/main/java/gherkin/GherkinDialect.java
mpkorstanje: There reason this happens is because of an unchecked cast in GherkinDialectProvider which forces the type to be a  `Map<String, Map<String, List<String>>>`. 

However the data structure is not actually a `Map<String, Map<String, List<String>>>` but  `Map<String, Map<String,  Object>>`  as the properties of a language in `gherking-languages.json` are either a list or a string.

This can be avoided by using the `fromJson` that accepts a type as input and using  `Map<String, Map<String,  Object>>` as type.

mpkorstanje:
@tsundberg I've just fixed the double casting so I can't think of any reason.

olleolleolle:
Pep comment from the sidelines: great usability addition!

aslakhellesoy:
The same API should be present in all other implementations. When that’s confirmed this can be merged.

aslakhellesoy:
What code will be calling the new method? Who's the consumer?

@brasmusson made a comment above that this should be implemented in all the other ports. I agree, and nobody has taken the time to do that.

I don't know exactly what use case this is supposed to support, but I wonder if it's easier to just parse the JSON directly.

For example: `cat gherkin-languages.json | jq -r 'to_entries[] | [.key, .value.name, .value.native] | @tsv'`

mpkorstanje:
This is intended to support: https://github.com/cucumber/cucumber-jvm/pull/1290

Parsing the json is possible but it seems to be an implementation detail of the gherkin library.

brasmusson:
In Cucumber-JVM v1.2.x the option `--i18n help` printed a list of language codes:
```
af
am
...
```
The [`GherkinDialectProvider.getLanguages()` was introduced](https://github.com/cucumber/cucumber/commit/4e28ec71cba6f2b0f3a066e1d9e7e03ff2c389d5) to enable the same behavior in Cucumber-JVM v2.x.x.

In Cucumber-Ruby the corresponding option `--i18n-languages` prints a list with the code, name and native name for each language:
```
  | af        | Afrikaans           | Afrikaans             |
  | am        | Armenian            | հայերեն               |
...
```
This is basically the behavior @tsundberg want Cucumber-JVM to also have in https://github.com/cucumber/cucumber-jvm/pull/1290.

Cucumber-Ruby [implements this behavior by by direct access to the `DIALECTS` in Gherkin-Ruby](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/cli/options.rb#L555-L558). With the stronger typing and encapsulation of Gherkin-Java, https://github.com/cucumber/cucumber-jvm/pull/1290 needs this PR to be able to implement something similar.

I opposed the first version of this PR (https://github.com/cucumber/cucumber/pull/313/commits/1821623c3bb814b786660de4d52ab9c96f2c5af6) where `GherkinDialectProvider.getLanguages()` was change to return a strings like "`af         Afrikaans            Afrikaans`" for each language, and pointed out that such a behavior should be available also in the other language implementations of Gherkin if available in Gherkin-Java. The current version of this PR, where the there is an accessor for the list of `GherkinDialect` object available in Gherkin-Java, on the other hand is more a consequence of the Java language. In Gherkin-Ruby there is not `GherkinDialectProvider` class, in the more flexible Ruby language the `Gherkin::Dialect` class and the module variable `DIALECT` suffice.

I think this PR can be merged.

brasmusson:
API for what?. I think that creating of tables like:
```
 | af        | Afrikaans           | Afrikaans             |
```
should be kept out of the gherkin library, so the gherkin library can be focused on parsing and compiling feature files.

As I see it the only thing missing in the gherkin library for java to support cucumber/cucumber-jvm#1290 is accessor for the name and native name in the `GherkinDialect` class. The `GherkinDialectProvider.getLanguages()` and `GherkinDialectProvider.getDialect()` enables Cucumber-JVM to iterate over all the `GherkinDialect` object, and with accessors for the name and native name, Cucumber-JVM can produce the extended language table.

aslakhellesoy:
Would something like [this](https://cucumber.netlify.com/gherkin/#gherkin-dialects) address the problem you're trying to solve @tsundberg? (The page is regenerated whenever there is a new Gherkin release).

This is the new [docs website](https://github.com/cucumber/docs.cucumber.io) some of us have been working on for the past months. It obviously needs some more styling, but you get the idea.

mpkorstanje:
I believe the intent was to improve the commandline output. 

The difference with the page is that the commanline refelects the current state of the app rather then the latest.

aslakhellesoy:
I think the intent was to improve the documentation of the dialects.

One way to do that is through the command line. Another way is through a web page.

I think a web page is sufficient, but I’m not going to object if someone wants to implement this change in 8 implementations.

tsundberg:
I just reverted the change in the Gherkin API. The current API should be compatible with the rest of the implementations.

I had to keep the addition of getting the name and the native name from `GherkinDialect` so the Java implementation can pick up name and native name. These properties wasn't exposed earlier.

tsundberg:
I'm closing this pull request and suggest that the changes are included from #366

Issue #326, opened 2018-01-19
cucumber-expressions: second string argument captures qoutes


mpkorstanje:
## Summary

Matching the expression `three {string} and {string} mice` against  the text `three 'blind' and 'crippled' mice` would result in two matches. `blind` and `crippled`. After #323 it produces `blind` and `'crippled'`.

## Steps to Reproduce (for bugs)

See failing CucumberExpressionTest in: 

https://github.com/cucumber/cucumber/tree/cucumber-expressions-regression-323

mpkorstanje:
Pushed a fix for java to https://github.com/cucumber/cucumber/tree/cucumber-expressions-regression-323

The problem appears to be caused by skipping the processing of `(`, `)`, `?` and `:` after setting the escape flag to false. Looks like the other implementations have the same problem.

PR #327, opened 2018-01-19
Cucumber expressions regression 323

aslakhellesoy:
WIP fix for #326 

- [x] Java
- [x] Ruby
- [x] JavaScript

### DIFF
--- a/cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/ParameterTypeRegistry.java
+++ b/cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/ParameterTypeRegistry.java
@@ -14,8 +14,7 @@ public class ParameterTypeRegistry {
     private static final List<String> FLOAT_REGEXPS = singletonList("-?\\d*[\\.,]\\d+");
     private static final List<String> HEX_REGEXPS = singletonList("0[xX][0-9a-fA-F]{2}");
     private static final List<String> WORD_REGEXPS = singletonList("\\w+");
-    private static final List<String> STRING_REGEXPS = singletonList("\"([^\"\\\\]*(\\\\.[^\"\\\\]*)*)\"|\\'([^\\'\\\\]*(\\\\.[^\\'\\\\]*)*)\\'");
-
+    private static final List<String> STRING_REGEXPS = singletonList("\"([^\"\\\\]*(\\\\.[^\"\\\\]*)*)\"|'([^'\\\\]*(\\\\.[^'\\\\]*)*)'");
     private final Map<String, ParameterType<?>> parameterTypeByName = new HashMap<>();
     private final Map<String, SortedSet<ParameterType<?>>> parameterTypesByRegexp = new HashMap<>();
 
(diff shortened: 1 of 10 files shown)


aslakhellesoy:
I added the same tests to Ruby and JavaScript (locally, without committing), and they both passed without any changes!

I then changed the implementation in both Ruby and JavaScript to be aligned with the changes to the Java implementation. The JavaScript implementation still passed, but the Ruby implementation now failed!

That worries me a little. I don't want the implementations to diverge, since that would make maintenance harder.

I haven't tried to understand why the implementations would behave differently. Any ideas?

aslakhellesoy:
I'm investigating this a little further. Let's look at the source of the regular expressions generated by the Cucumber Expression `three {string} and {string} mice`:

```
Java:
^three ("([^"\\]*(\\.[^"\\]*)*)"|\'([^\'\\]*(\\.[^\'\\]*)*)\') and ("([^"\\]*(\\.[^"\\]*)*)"|\'([^\'\\]*(\\.[^\'\\]*)*)\') mice$

JavaScript:
^three ("([^"\\]*(\\.[^"\\]*)*)"|'([^'\\]*(\\.[^'\\]*)*)') and ("([^"\\]*(\\.[^"\\]*)*)"|'([^'\\]*(\\.[^'\\]*)*)') mice$

Ruby:
^three ("([^"\\]*(\\.[^"\\]*)*)"|'([^'\\]*(\\.[^'\\]*)*)') and ("([^"\\]*(\\.[^"\\]*)*)"|'([^'\\]*(\\.[^'\\]*)*)') mice$
```

The JavaScript and Ruby regexps are identical, but the Java one differs in that `'` characters are escaped.

Maybe related - I [discovered today](https://github.com/cucumber/cucumber/issues/325#issuecomment-358985126) that Ruby and JavaScript escapes `/` differently in regexp source.

Tricky stuff!

mpkorstanje:
The `\'` is in the java defintion of `{string}`.

mpkorstanje:
My fix is definitely wrong.

aslakhellesoy:
Looks like changing `\\'` to `'` in the `{string}` regexp fixes it! And no changes needed to Ruby/JavaScript because they don't have it!

mpkorstanje:
That should do it. `'` is not a character that would have to be escaped.

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/ParameterTypeRegistry.java
```diff
@@ -14,8 +14,7 @@
     private static final List<String> FLOAT_REGEXPS = singletonList("-?\\d*[\\.,]\\d+");
     private static final List<String> HEX_REGEXPS = singletonList("0[xX][0-9a-fA-F]{2}");
     private static final List<String> WORD_REGEXPS = singletonList("\\w+");
-    private static final List<String> STRING_REGEXPS = singletonList("\"([^\"\\\\]*(\\\\.[^\"\\\\]*)*)\"|\\'([^\\'\\\\]*(\\\\.[^\\'\\\\]*)*)\\'");
-
+    private static final List<String> STRING_REGEXPS = singletonList("\"([^\"\\\\]*(\\\\.[^\"\\\\]*)*)\"|'([^'\\\\]*(\\\\.[^'\\\\]*)*)'");
```
mpkorstanje: I think I did this right.

mpkorstanje:
Eh there is a bit in the ParameterTypeRegistry too that needs to change. Have to look at it another time.

aslakhellesoy:
I fixed the missing bits (I think!)

PR #336, opened 2018-02-23
Fixes #335

### LINKED ISSUE #335: gherkin: python implementation is not thread safe
## Summary
gherkin-python may fail to parse valid feature document in multi-thread context

## Possible Solution
The cause of this problem is using mutable object as default arguments in parser.py (which is generated from gherkin-python.razor). It could be fixed easily by deferring default argument instantiation.

* Version used: gherkin-official 4.1.3

link89:
#335


link89:
It's kind of you to point this out

PR #353, opened 2018-04-07
Allow Examples under Scenario (deprecate Scenario Outline keyword)

aslakhellesoy:
## Summary

With this change it is possible to write scenario outlines with the regular `Scenario` keyword:

```gherkin
Scenario: minimalistic
  Given the <what>

  Data:
    | what       |
    | minimalism |
```

The `Examples` keyword has a new synonym: `Data`.

## Details

The `Scenario Outline` keyword (and its synonyms) are still supported, but using them causes a deprecation warning to be printed to STDERR.

The `Examples` keyword (and its synonyms) are still supported, but using them causes a deprecation warning to be printed to STDERR.

Currently this is only implemented in Java, just to get a feel for it and to gather feedback. All the other implementations would have to be updated too before we can merge this.

## Motivation and Context

There are a couple of reasons for this proposed change:

* It simplifies the grammar
* It simplifies i18n translations

At Cucumber Ltd we have taught Cucumber to hundreds of people. One of the exercises we make people do is to refactor three scenarios into a scenario outline. They usually keep one of the scenarios, add the `<placeholders>` to the existing steps, add an `Example` section with a header and 3 rows of data. They *very often forget* to change the keyword from `Scenario` to `Scenario Outline`. Many have commented that a separate keyword shouldn't really be necessary, and they're right.

We can process the scenario as a scenario outline if there are one or more `Example` sections underneath, and as a scenario if there are none. This makes the Gherkin syntax a little easier to learn.

Another reason for this change is the proposed rename of `Scenario` to `Example` over in #351. @mattwynne pointed out that he didn't like `Example Outline`, and if we remove it altogether that is a non-issue.

Removing `Scenario Outline` from the grammar (and merging #351) would mean that "scenario outline" needs a new name. It's no longer a syntactic keyword of Gherkin, but it's still a **thing**.

If we merge this PR I propose we start calling them **parameterised examples** or  **data examples** instead. An `Example` with `Examples` tables underneath is a parameterised example (or a data example, or whatever we end up calling it).

This brings up another thing we should change. Now we have `Example` and `Examples` and that's too similar. 

@charlierudolph suggested in https://github.com/cucumber/cucumber/pull/351#issuecomment-378980521  that we add a `Data` synonym for `Examples`. I like that, and it's implemented in this PR. I also suggest `Examples` and existing synonyms get deprecated just like `Scenario Outline`.

## Consequences

This change should be backwards compatible, both in terms of API and syntax. The only behavioural change I am aware of is `Scenario Outline` without examples:

https://github.com/cucumber/cucumber/blob/6973e0f4395d2a1b311080438e81356fb2724399/gherkin/testdata/good/incomplete_scenario_outline.feature#L6

Before this change, this would not result in a pickle being created, but with this change it will. (There is no way to detect that the user intended to create a "WIP" scenario outline). This might affect a small number of users who have such WIP scenario outlines, and they might be marked as undefined and cause the build to fail.

Another consequence is that syntax highlighters in text editors, IDEs, GitHub etc will have to be updated. We should help the maintainers of those tools to update them.

## How Has This Been Tested?

Most of the `testdata` scenarios are unchanged (they still use `Scenario Outline` and `Examples`). Some of them have been changed to verify that the new keywords also work.


aslakhellesoy:
Glad you like this! I'll remove the deprecation notice - you convinced me @mpkorstanje

mattwynne:
I agree with @mpkorstanje about keeping the `Examples` keyword for the table and I'd add one more thing.

I think in general that one of the good things about Gherkin is that it is loose - *just* formal enough to be understood by a computer, but flexible to the messy ways humans like to express things.

For that reason I'd prefer to not deprecate `Scenario Outline` and let people continue to use it if that's what they'd prefer / are used to. There must be thousands of existing uses of this keyword out there in the wild, and I'm not sure what benefit we're giving people by forcing them to move to a different keyword.

I don't agree that `Data` is a good keyword for what we've historically called an examples table. New users already get confused about the difference between Data Tables (the multi-line argument to a step) and examples tables, and I think that would only make this problem worse. `Examples` is definitely the best name for me to represent what's in those tables.

mattwynne:
Oh and to be clear, I do think that being flexible about whether people have said `Scenario` or `Scenario Outline` is a brilliant idea, I've also seen loads of people trip over this when they're getting started with Cucumber.

aslakhellesoy:
Allright, we won't deprecate anything.

Matt, consider this traditional `Scenario Outline`:

```gherkin
Scenario Outline: eating
  Given there are <start> cucumbers
  When I eat <eat> cucumbers
  Then I should have <left> cucumbers

  Examples:
    | start | eat | left |
    |  12   |  5  |  7   |
    |  20   |  5  |  15  |
```

Assuming I want to replace `Scenario Outline` with `Example` (as supported by this PR and #351) - are you suggesting leaving `Examples` as-is without a new keyword? You don't think `Example` + `Examples` is a little odd?

aslakhellesoy:
I'm not married to `Data`. I agree it could be confused with "data tables". However, I think it's equally confusing that an `Example` has `Examples`, so we have to come up with a better synonym.

`Arguments`? `Variants`? `Values`? `Combinations`?

Help me out here people :-) I want to hear what you think.

mattwynne:
@aslakhellesoy I do think `Example` / `Examples` is a little bit odd, but I don't really think it does any harm. If I wanted to make a clear distinction I could use one of the other synonyms for `Example` like `Scenario` or even good old fashioned `Scenario Outline`.

mattwynne:
I hate to throw this into the mix, because I like the spirit of being flexible and allowing `Scenario` as well as `Scenario Outline`, but I have thought for a long time that a more appropriate name for `Scenario Outline` would be `Rule` because in general what you see after the keyword is a rule, e.g.

```
Rule: Passwords must be 6 characters long
   When I try to create an account with password <password>
   Then I am told that my password is <valid or invalid>

  Examples: Too Short
    | password | valid or invalid |
    | x        | invalid          |
    | xxxxx    | invalid          |

  Examples: Long enough
    | password | valid or invalid |
    | xxxxxx   | valid            |
```

Sorry if that's taking this too far O/T. :)

aslakhellesoy:
@mattwynne thanks for raising the suggestion to make `Rule` a synonym for `Scenario Outline`. I think it's a good idea in principle, but I don't think it's practical for a few reasons outlined below.

Let's examine the consequences of some of the grammar suggestions:

* #351: `Scenario` == `Example`
* #353: `Scenario` == `Scenario Outline`
* Your suggestion: `Rule` == `Scenario Outline`
* #250: `Rule` is a new container for `Scenario`/`Example`

The consequence of the first three is that `Scenario`, `Example`, `Scenario Outline` and `Rule` are all synonyms. They can be placed directly underneath `Feature`, and they can all contain `Given/When/Then` steps and an optional set of `Examples`.

Adding #250 to the mix (where `Rule` can contain `Example`) would make the grammar recursive. 

This would require changes to Berp (which has limited support for recursive grammars) and some significant work in the AST builder I imagine. But most importantly, a recursive grammar (with lots of vocabulary synonyms and ambiguity) would make Gherkin too hard to learn. People would end up writing this kind of nonsense:

```gherkin
Feature:
  Example:
    Rule:
      Rule:
```

A word should only mean one thing. More specifically, a `Rule` should either be a container of `Example` (ref Example Mapping and #250) *or* it should be a synonym for `Scenario Outline`.

But not both.

I think mimicking the Example Mapping structure is more important. With that grammar `Rule` is just a container for grouping related `Examples`. Your example could be written like this:

```gherkin
Feature:
  Rule: Passwords must be 6 characters long

    Example: Validation of password length
       When I try to create an account with password <password>
       Then I am told that my password is <valid or invalid>

      Examples: Too Short
        | password | valid or invalid |
        | x        | invalid          |
        | xxxxx    | invalid          |

      Examples: Long enough
        | password | valid or invalid |
        | xxxxxx   | valid            |
```

(I still don't like the `Example/Examples` combo, but I'll address that in a separate reply).

aslakhellesoy:
> If I wanted to make a clear distinction I could use one of the other synonyms for Example like Scenario or even good old fashioned Scenario Outline

Sure, but I don't want to use `Scenario` or `Scenario Outline` anymore. I'm sold on the `Example` terminology. 

I think my preference so far is for `Combinations`. 

If you don't like it you could always go for `Scenario Outline` + `Examples`, or `Example` + `Examples`. I would imagine a dyslexic person would struggle to differentiate between `Example` and `Examples`. In some languages (like Chinese) there isn't even a distinction - both plural and singular are `例子` (AFAIK).

gasparnagy:
I also don't like `Rule` to be a synonym of `Scenario Outline`, basically because of the arguments @aslakhellesoy has mentioned. I think scenario outline describes a special kind of rule (or better a subset of the examples of a special kind of rule), that can be better illustrated with data variations. But even in that case I sometimes need multiple scenario outlines today to illustrate the entire rule (e.g. positive / negative cases). So keeping the rule as a higher-level element is better in my opinion.

I also not that much fond of using `Example` instead `Scenario`, but I will make my comments at #351.

aslakhellesoy:
I think this is good to go now.

I intend to port this to JavaScript and Ruby. Please fill in your name if you're volunteering to port this to another language.

* [ ] c
* [ ] .net
* [ ] go
* [x] java (@aslakhellesoy)
* [ ] javascript (@aslakhellesoy)
* [ ] objective-c
* [ ] ruby (@aslakhellesoy)
* [ ] perl
* [ ] python

I'll merge this to master a week or so after javascript/ruby are complete. Any implementations lacking behind will be stuck and stop to evolve. I'll have to disable their builds. Releases will only be made on "enabled" implementations.

After this is merged, I intend to start implementing #250, followed by #351. Then I reckon it's time for a major release. I might want to include #378 in that release too.

So lots going on @cucumber/gherkin team. I know this is a lot to keep up with.

Keeping up with changes is a challenge for this project. There are 9 implementations of Gherkin. I'm getting more and more convinced that the [pickle runner](https://github.com/cucumber/cucumber-pickle-runner) (we need a new name I think @charlierudolph) will reduce the cost of ownership considerably. When if it becomes a reality, I don't see a strong reason to maintain more than one or two implementations of Gherkin. I'm thinking Go and JavaScript (because you know - web).

Let me know your thoughts. /cc @mattwynne @jbpros @tooky @mpkorstanje @sebrose @mlvandijk

mlvandijk:
@aslakhellesoy What are the considerations to pick the one or two languages for the implementations of Gherkin? (Personal note: I've been wanting to learn Go (but haven't had a reason to, professionally), but would be way more comfortable with Java atm ;) )

aslakhellesoy:
The reasons for Go are:

* Binaries start with little lag, and performance is better than most other implementations (except for C)
* It's easy to compile static binaries to most platforms/OSes
* It's easier to write correct Go code than correct C code
* Binaries don't require a runtime

Non-noticeable startup lag is essential for a tool like Cucumber. This rules out JVM.

A shared binary built on top of a runtime that must be installed separately is a no-go. Requiring C# or JavaScript developers to install a JVM would turn people off.

The pickle runner is written in Go (for the above reasons), so the Cucumber "standard library" must also be in Go.

aslakhellesoy:
@xtrasimplicity where did I mention a rewrite?

xtrasimplicity:
@aslakhellesoy Sorry, I completely misread your initial post!

SabotageAndi:
@aslakhellesoy I could make the changes to the .NET version, but I will not be able to do this in such a short timeframe. I have to much to do at the moment with .NET Core/Standard support for SpecFlow.

my 2 cent about having only 2 implementations: calling native libraries and packaging them is always a hassle in the .NET world and so we tend to have everything in managed (.NET code). So calling in the future in Go code makes a lot of work for us. So I would prefer the .NET implementation will survive.

aslakhellesoy:
@SabotageAndi there won't be any calling of native libraries. Calling native libraries from a different platform is a hassle with most platforms, not just .NET. For this reason the shared binary will use [protobuf](https://developers.google.com/protocol-buffers/) over STDIN/STDOUT for communication.

Packing executables inside a nuget package (or other kind of package) should be relatively easy to do. The approach taken by [protoc-jar](https://github.com/os72/protoc-jar) is to package the `protoc` executable for 4 common platforms inside the package. The executable for the current platform is unpacked and written to disk if needed.

What survives or will depend on what gets maintained. Personally I have to reduce the amount of code to maintain, so going forward I only plan on maintaining Java/Ruby/JavsScript/Go code. I'll happily merge PRs for other implementations though - and make releases.

mlvandijk:
@aslakhellesoy Makes sense, thanks for explaining (Finally, a reason to learn go \o/ )

KyleFairns:
Sorry for being really late to the party, but having read through this it seems like what we're wanting is just an extra layer between Feature and Scenario.

```gherkin
Ability: To document the rules of the Features/Abilities/Business Needs
   
     Rule: A descriptor to add intent to the Scenarios/Examples

           Example: I have a rule that I want to demonstrate
                When I <have?> a rule that I want to demonstrate
                Then I <should?> include the "Rule" descriptor

             Scenarios:
                 | have?        | should?     |
                 | have         | should      |
                 | have not got | should not  |
   
```

In other words, where a Feature is a collection of Rules and Scenarios that demonstrate that Feature, the Rules are a collection of Examples and Scenarios that demonstrate that rule.

Having them in the same file for instance:

```gherkin
Ability: Having cheese on crackers

    Rule: The type of cheese should decide how I put it on my cracker
         
         Example: Soft cheese is spreadable
             Given I have soft cheese
             When I am preparing my cracker
             Then I should be able to spread the cheese onto my cracker
         
         Example: Hard cheese is cuttable
             Given I have hard cheese
             When I am preparing my cracker
             Then I should be able to cut cheese slices onto my cracker

    Scenario: Eating a cracker
        Given I have a <type_of_cracker> cracker
        When I have prepared my cracker with <type_of_cheese>
        Then I should be able to eat my cracker

          Examples:
              | type_of_cracker | type_of_cheese |
              | Ritz            | Philadelphia   |
              | Cream           | Cheddar        |
```

aslakhellesoy:
@KyleFairns the latter won't work. `Example` and `Scenario` are synonyms. The parser can't know whether the last `Scenario` should be under the `Abiility/Feature`, or under the `Rule`. The parser doesn't use indentation as a hint like Python does.

Once you've added a `Rule` in a file, all subsequent `Scenario`/`Example` will belong to that rule, until you add another `Rule`.

KyleFairns:
Would it be possible to separate their functionality? Whereby to have an Example block, you must have a parent Rule, and all Scenarios are instead parsed as children of the feature?

This would mean that when you use the Scenario/Examples keywords, you are documenting how the Feature works, and when using the Rules Example/Scenarios keywords, you are demonstrating how the Rule for that Feature works?

Would it be useful to have this distinction in order to fully document the features functionality, without having all of the use cases tied to a rule?

KyleFairns:
As a side note as well, how will Backgrounds be worked in with all of this?

Will they be applied to every Example in every Rule, and also be applied to the Scenarios?

PR #355, opened 2018-04-09
[gherkin-dotnet] make compilation on macOS & linux the same as on Windows

SabotageAndi:
## Summary

This PR makes the compilation for gherkin-dotnet the same on all platforms.
Before this there was no build for net45 on macOS and Linux.

## Motivation and Context

The last release was probably build on Mac OS and so it doesn't contain a net45 assembly. 
This PR solves this issue

## Types of changes

## Checklist:


aslakhellesoy:
Hi @SabotageAndi - did you see the Travis errors? Any idea how to fix that?

SabotageAndi:
As far as I understand it, it looks like mono is missing in the docker container.
I wasn't aware that the build server uses docker to run the tests and so I didn't changed there anything. I will have a look at it.

aslakhellesoy:
Mono is an optional dependency. It's only used to run the Berp parser generator, but the Makefile allows it to fail (hence the hyphenated `-mono`) in case it's not installed. In that case the parser is not regenerated. That's OK as long as there are no changes to the `gherkin.berp` grammar - the generated `Parser.cs` is in Git.

So missing mono is not the problem with the [build](https://travis-ci.org/cucumber/cucumber/builds/364019806?utm_source=github_status&utm_medium=notification) (line 6364).

The problems is [further down](https://travis-ci.org/cucumber/cucumber/builds/363987358#L6413) - something about assembly signing.

I suggest you revert b3eb9d6 as it shouldn't be needed.

SabotageAndi:
@aslakhellesoy It wasn't only missing mono, that caused problems.
I have to use mono to compile on non Windows platforms for .NET Framework 4.5.
I hope the additional dependency is no problem.

The rest of the changes are needed to use msbuild from mono to build the stuff and get the unit tests running again.

After you publish a new docker image, the travis build should be green.

aslakhellesoy:
The [monorepo build](https://travis-ci.org/cucumber/cucumber) (which uses Docker) passed.

However, the [subrepo build](https://travis-ci.org/cucumber/gherkin-dotnet) of [gherkin-dotnet](https://github.com/cucumber/gherkin-dotnet) now fails. It says `msbuild` is not found.

Any ideas how to fix this? If you want to experiment you can commit directly to that repo and I'll sync back to the monorepo.

SabotageAndi:
@aslakhellesoy The builds are using different versions of .NET Core (1.1.2 and 1.1.5). I will fix that in the next days.

PR #357, opened 2018-04-09
Cucumber expressions: Remove single transformer

aslakhellesoy:
I didn't like the choices between `Transformer` and `Function`, so I removed `Function` (and `SingleTransformer`).

Merging this will cause cucumber/cucumber#1248 to break, but since it's never been released I'm not sure we need to bump the minor release.

I will have to make some minor consistency changes to Ruby and JavaScript, but I don't think it will affect behaviour externally.


mpkorstanje:
Ideally, when using a lambda expression, you won't notice those.  By defining the lambda as  `(String[] matches) -> ....` or `String match -> ...`. you can decide if you want to use all parameter or only the first one. I expect that most parameter types will use the whole matching making this a nice bit of sugar.

The reason we have a function class is because I had to back port this to 1.7 which doesn't support lambdas yet (but this won't prevent clients at 1.8 from using them).

In javascript you can get the same behaviour when using `transformFn.call(null, matches)`.  If the `transformFn` is defined as  `function(arg1, arg2, ... argN)` the matches will be bound to the arguments.

aslakhellesoy:
Ok @mpkorstanje I added some more generics, bringing back support for single argument transforms while keeping a single interface.

I couldn't figure out a way to avoid the static factory methods `ParameterType.single` and `ParameterType.multi`, but perhaps that's not so bad?

WDYT?

mpkorstanje:
You've found the only two ways to do this. Java needs the type information which requires that you spell out exactly what each method looks like. And as you've also noticed generics are erased which means you also need to use static factory methods to differentiate when using different generics.

So either the information about the number of arguments is contained in the lambda signature `(String[] args) -> ...` vs `(String arg) -> ...` or it is contained in the name of the static factory method `ParameterType.single` vs `ParameterType.multi`.

Given that choice, the type information has to be conveyed one way or the other, I would prefer to use the lambda signature as it conveys more clearly what the type of `args` will be and does so very near the location that `args` will be used. The static factory method conveys it from further away and does so implicitly. 

It is also important to keep in mind is that the `DataTableType` in datatable follows the same pattern. If we change to static constructors here we should do so too for `DataTableType`. And looking at the difference between `DataTableType` and `ParameterType` I can see that that the former spells out the different transformers `TableTransformer`, `TableEntryTransformer`, `TableRowTransformer` and `TableCellTransformer`. In the latter we have `Transformer` and `Function`. So I think we just need to come up with a better name for those two.

aslakhellesoy:
Ok. How about `Transformer` (using `String`) and `MultiTransformer` (using `String[]`)?

aslakhellesoy:
I added `MultiTransformer` and got rid of the static ctors.

The only thing that annoys me now is that method references only work if I cast:

```java
(Transformer<Currency>) Currency::getInstance
```

Should we try to find a way to avoid that?

mpkorstanje:
> Should we try to find a way to avoid that?

You can't really avoid that. Both `ParameterType` and `Currency::getInstance` are overloaded. This combination prevents resolution. It will work fine with non overloaded method references.

mpkorstanje:
I've removed `Function` all together.  We only expose `Transformer<T>` and `MultiTransformer<T>` so `ParameterType.transformer` will always be either one or the other. We could try to capture this somehow but this provides us no benefit. 

So I've put the adapter pattern back in again. But it is no longer exposed to the world.

aslakhellesoy:
@mpkorstanje I tried it out with Java8, and this doesn't seem to work with method references. Did you try it?

mpkorstanje:
Yes! They work for me provided you're using one that is not overloaded. This means `Currency::getInstance` just won't won't work in this setup. I don't think this is such a bad thing though.

I do think switching parameter order is worse. It makes the API inconsistent. 

If you think this is a must we could make the `Transform` an instance with no public methods and static  constructors for both use cases: 

```
<T> Transformer.transform(Function<String,T>)
<T> Transformer.transformCaptureGroups(Function<String[], T>)
```

aslakhellesoy:
I think it would be a shame not to be able to use method references. User code would be very verbose without them. Not sure how you'd implement it though - want to give it a go?

mpkorstanje:
@aslakhellesoy  check this:

https://github.com/cucumber/cucumber/blob/d7f11dc1335f6db67ce2e776493b0d7cc1faaf9c/cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/Examples.java#L50

I faked out a few parts to just get the point across. 

Broadly considered we have to make a trade off between 3 options.

1. We only support a single transform  `(String[] args) -> ....`. This means that people will often have to do `args[0]` in their lambda's and can only use static constructors that accept an string array. Otherwise it's very concise. 

2. We support  `(String[] args) -> ....` and  `(String arg) -> ....`. This means that people can use the more common static constructors that accept a string. They won't be able to  use overloaded method references but is otherwise quite concise. The verbosity of a lambda is not much higher then a method reference.

3.  We support `transformCaptureGroups((String[] args) -> ....)` and `transform(String arg) -> ....)`. This means people can use overloaded method references but adds a constant cost to all cases and doesn't improve the verbosity much when dealing with over loaded method references.

I reckon option 1 is right out. Normally I'd data mine the choice between one and 2 and 3, but from experience I'd say 2 is the best one.

mpkorstanje:
I've renamed the `MultiTransformer` to `CaptureGroupTransformer`. It doesn't transform capture groups, but it does transform the parts of the string matching those specific capture group. So... close enough.

You didn't revert the commit in which you swapped the parameter order. I've fixed that too.

I'm happy for now. Lets release this!

PR #366, opened 2018-04-20
Exposed name and native name to enable Cucumber-JVM to print the lang…

tsundberg:
## Summary

Cucumber-JVM can't print the name and native name for a supported language. I exposed two methods in Gherkin so the name and native name can be read from Cucumber-JVM.

## Details

This is the same as the pull request #313 but with only one commit. #313 was a mistake in the sense that it started out with a Gherkin change that made the Java implementation diverge from the other implementations. exposing two methods like this is less intrusive.

## Motivation and Context

This will make it easier to understand what language a language code is supporting.

## How Has This Been Tested?

I added unit tests that veriffies that it is possible to get the name 'Swedish' and the native name 'Svenska'

## Types of changes

## Checklist:

### DIFF
--- a/gherkin/java/src/main/java/gherkin/GherkinDialect.java
+++ b/gherkin/java/src/main/java/gherkin/GherkinDialect.java
@@ -3,22 +3,25 @@ package gherkin;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 public class GherkinDialect {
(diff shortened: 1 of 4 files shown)


aslakhellesoy:
This PR has gone stale in the light of #425. Gherkin 6 will have a `--dialects` option for printing the JSON file, which can be piped to various tools for generating documentation or code.

Thanks for your contribution though!

PR #370, opened 2018-04-21
[docs] Drop residual gitbook settings files

olleolleolle:
## Summary

After #367 this residual Gitbook configuration file and ToC Markdown file can also go.

## Details

A json and a markdown file related to the now-defunct Gitbook were dropped from the repository.

## Motivation and Context

We've now moved efforts to the docs.cucumber.io repo.

## How Has This Been Tested?

In no way at all, I'm asking for a review instead.

## Types of changes

Remove a redundancy.


olleolleolle:
I can now see that SUMMARY.md is also a Gitbook-related file, which should also go.

olleolleolle:
As I squint into the history of these files,  I also nominate `OLD_README.md` to be deleted:

- its test refers to a no-code repo
- it technically explains monorepos and how to create them
- its link to the documentation site is outdated and only points to the documentation start page

PR #385, opened 2018-05-10
datatable: implemented js datatable

lenntt:
## Summary
Bringing the ability to define custom datatables to js. 

## Details
Before we're getting into details, I'd like to point this out: 
There's a lot of duplication with parameter type registry (in cucumber-expressions).
Ideally, I'd want to predefine some types, likes `[int]`... but that requires the paramter type registry.
Shouldn't we just build this at the same place the parameter type registry is? (move parameters out of expressions?)
Shouldn't it just they both just be the same registry?
@aslakhellesoy @mpkorstanje 

When we do get into details:
There's some things to be aware of / TODOs
- Its my first (real) commit: dont hold back on your comments, I likely messed up a lot
- Although I wrote tests before code, I didnt consider many edge cases yet, at least should do some parameter checks.
- I pretty much ignored any kind of CI for now
- I haven't really looked at how this ties into `cucumber-js` yet
- I did not look into a general (for all languages) DataTable test set, nor writing documentation

## Motivation and Context
conversation started here: https://cucumberbdd.slack.com/archives/C6QJ6N695/p1525332789000331

## How Has This Been Tested?
Merely unit tests - not even sure if I built the right thing (see above)

## Types of changes

## Checklist:

### DIFF
--- a/datatable/javascript/src/data_table.js
+++ b/datatable/javascript/src/data_table.js
@@ -1,0 +1,69 @@
+class DataTable {
+  // TODO: jsdoc
+  // TODO: fail on unequal rows?
+
+  constructor(rawTable) {
+    this._rawTable = rawTable
+  }
+
+  get cells() {
+    return this._rawTable
+  }
+
+  get height() {
+    return this._rawTable.length
+  }
+
+  get width() {
+    return this._rawTable[0].length
+  }
+
+  get isEmpty() {
+    return this.height <= 1 && this.width === 0
+  }
+
+  static create(rawTable) {
+    return new DataTable(rawTable)
+  }
+
+  transpose() {
+    let transposedRaw = this.cells[0].map((col, i) =>
+      this.cells.map(row => row[i])
+    )
+    return DataTable.create(transposedRaw)
+  }
+
+  row(index) {
+    return this.cells[index]
+  }
+
+  rows(fromRow, toRow) {
+    let rows = this.cells.filter((row, rowIndex) => {
+      return rowIndex >= fromRow && (!toRow || rowIndex < toRow)
+    })
+    return DataTable.create(rows)
+  }
+
+  column(index) {
+    return this.cells.map(row => {
+      return row[index]
+    })
+  }
+
+  columns(fromColumn, toColumn) {
+    let columns = this.cells.map(row => {
+      return row.filter((cell, columnIndex) => {
+        return (
+          columnIndex >= fromColumn && (!toColumn || columnIndex < toColumn)
+        )
+      })
+    })
+    return DataTable.create(columns)
+  }
+
+  subTable(fromRow, fromColumn, toRow, toColumn) {
+    return this.rows(fromRow, toRow).columns(fromColumn, toColumn)
+  }
+}
+
+module.exports = DataTable
--- a/datatable/javascript/src/data_table.js
+++ b/datatable/javascript/src/data_table.js
@@ -7,7 +7,7 @@ class DataTable {
   }
 
   get cells() {
-    return this._rawTable
+    return this._rawTable.slice(0)
   }
 
   get height() {
@@ -33,7 +33,15 @@ class DataTable {
     return this.cells[index]
   }
 
+  /**
+   * The table as a 2-D array, without the first row, unless a range is specified
+   * @param {number} fromRow The first row (including).
+   * @param {number} toRow The last row (excluding).
+   */
   rows(fromRow, toRow) {
+    if (typeof fromRow === 'undefined') {
+      fromRow = 1
+    }
     let rows = this.cells.filter((row, rowIndex) => {
       return rowIndex >= fromRow && (!toRow || rowIndex < toRow)
     })
@@ -46,7 +54,15 @@ class DataTable {
     })
   }
 
+  /**
+   * The table as a 2-D array, without the first column, unless a range is specified
+   * @param {number} fromColumn The first column (including).
+   * @param {number} toColumn The last column (excluding).
+   */
   columns(fromColumn, toColumn) {
+    if (typeof fromColumn === 'undefined') {
+      fromColumn = 1
+    }
     let columns = this.cells.map(row => {
       return row.filter((cell, columnIndex) => {
         return (
@@ -59,6 +75,39 @@ class DataTable {
 
   subTable(fromRow, fromColumn, toRow, toColumn) {
     return this.rows(fromRow, toRow).columns(fromColumn, toColumn)
+  }
+
+  /**
+   * @returns  returns an array of objects where each row is converted to an object (column header is the key)
+   */
+  hashes() {
+    const copy = this.cells
+    const keys = copy[0]
+    const values = copy.slice(1)
+
+    return values.map(row => {
+      const obj = {}
+      keys.forEach((key, i) => {
+        obj[key] = row[i]
+      })
+      return obj
+    }, {})
+  }
+
+  /**
+   * @returns the table as a 2-D array
+   */
+  raw() {
+    return this.cells
+  }
+
+  /**
+   * @returns returns an object where each row corresponds to an entry (first column is the key, second column is the value)
+   */
+  rowHash() {
+    return this.cells.reduce((acc, target) => {
+      return Object.assign(acc, { [target[0]]: target[1] })
+    }, {})
   }
 }
 
--- a/datatable/javascript/src/data_table.js
+++ b/datatable/javascript/src/data_table.js
@@ -22,15 +22,11 @@ class DataTable {
     return this.height <= 1 && this.width === 0
   }
 
-  static create(rawTable) {
-    return new DataTable(rawTable)
-  }
-
   transpose() {
     let transposedRaw = this.cells[0].map((col, i) =>
       this.cells.map(row => row[i])
     )
-    return DataTable.create(transposedRaw)
+    return new DataTable(transposedRaw)
   }
 
   row(index) {
@@ -41,7 +37,7 @@ class DataTable {
     let rows = this.cells.filter((row, rowIndex) => {
       return rowIndex >= fromRow && (!toRow || rowIndex < toRow)
     })
-    return DataTable.create(rows)
+    return new DataTable(rows)
   }
 
   column(index) {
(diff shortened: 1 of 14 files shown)


mpkorstanje:
> There's a lot of duplication with parameter type registry (in cucumber-expressions).
> Ideally, I'd want to predefine some types, likes [int]... but that requires the paramter type registry.
> Shouldn't we just build this at the same place the parameter type registry is? (move parameters out of expressions?)
> Shouldn't it just they both just be the same registry?

They look the same but they aren't the same.

- Paramter types in cucumber expressions work on the results of regexes. 
- The data table types work on table cells, table rows, table entries and whole tables. 

This means that you can't reuse cucumber expression types as data table types.  An example of this difference would be `string`. In cucumber expressions escaped quotes are unescaped. In data table they are not. 

What does look similar all the other functions that transform a string into a basic type. But this is merely a coincidence and trivial enough to ignore the duplication effects off.

datatable/javascript/src/data_table.js
```diff
@@ -0,0 +1,69 @@
...
+
+  get width() {
+    return this._rawTable[0].length
+  }
+
+  get isEmpty() {
+    return this.height <= 1 && this.width === 0
+  }
+
+  static create(rawTable) {
```
mpkorstanje: I added this in Java so I could hide the constructor. This makes it easier to add in other `create` methods later on that do some preprocessing before constructing the table. 

You can't do this in JS so I think create is not needed.

datatable/javascript/src/data_table.js
```diff
@@ -0,0 +1,69 @@
...
+    return new DataTable(rawTable)
+  }
+
+  transpose() {
+    let transposedRaw = this.cells[0].map((col, i) =>
+      this.cells.map(row => row[i])
+    )
+    return DataTable.create(transposedRaw)
```
mpkorstanje: Neat!

lenntt:
@mpkorstanje thanks for the quick feedback.

I think I see your argument for keeping some separation. Just as a brain dump, I think one could desire to have a type defined both for expressions and datatables:
```
When winning numbers are 1, 2 and 3
// expression: `winning numbers are {[int]}`
```
```
When winning numbers are:
    | 1 | 
    | 2 | 
    | 3 |
// expression: `winning numbers are` + data table type `[int]`
```

It would get interesting is when you define a custom parameter type `mytype` and would like to get `[mytype]` table for free - which would be totally awesome. we could quite easily build that smartness into datatable I guess, but it (or another entity) has to know about the parameter type registry. 
together with a normalization (like the escaped quotes vs non escaped quotes you mention).


> They look the same but they aren't the same.

I feel they are conceptually more similar then is implemented, and it somewhat still feels unnatural tables are so diverged from types in expressions.

mpkorstanje:
You might be able to achieve this one abstraction level higher. There you could multiplex certain registrations to both the data table and parameter type registry. But you'd have to know if regex and transform function are safe to use.

For example:

```
defineParameterType(new ParameterType<>(
    'color',           // name
    /red|blue|yellow/, // regexp
    s => new Color(s)  // transformer function
))

defineDataTableType(new DataTableType<>(
    'color',           
    new TableCellTransformer(s => s is red blue or yellow ? new Color(s) : throw excpetion)
))
```

Could be replaced with a single call if  `defineParameterType` were to multiplex all registrations to `defineDataTableType` and the real `defineParameterType` like so:

```
defineParameterType(new ParameterType<>(
    'color',           // name
    /red|blue|yellow/, // regexp
    s => new Color(s)  // transformer function
))
```

But now you could put `orange` in the data table and end up with a valid, but rather unexpected color. I suppose you could wrap the transform function in a function that checks the regex first and aborts if it doesn't match.

mpkorstanje:
The above doesn't solve the problems of  `string`.

This gives us a few options:

1. Leave both registrations separate. Downside, no free conversion. Upside, no problems
2. Demultiplex registrations.  Free conversion, but we need a solution and we need to be able to explain it.

Question: How often are types between steps and data tables shared?  If often it'd be worth spending time one. Otherwise the actual duplicate registration will be a rare occurrence so not really worth adding complexity for .

aslakhellesoy:
That was fast! I'll need some time to review this. I'll post here later.

lenntt:
Transformers implemented somewhat differently than java: as static methods of `DataTableTransformer` (instead of their own classes). Found that a js-friendlier approach. 
And for consistency, used the name `DataTableTransformer` instead of `TableTransformer`

mpkorstanje:
>  And for consistency, used the name DataTableTransformer instead of TableTransformer

Perhaps I should consider that too.

mpkorstanje:
> Transformers implemented somewhat differently than java: as static methods of DataTableTransformer (instead of their own classes). Found that a js-friendlier approach.

Looks good to me!

aslakhellesoy:
At first glance this looks like a rather different implementation from the Java one. If you look at the other libraries (gherkin, cucumber-expressions, tag-expressions) you'll find that implementations are pretty similar. Same class names, same method names, same implementations, same tests etc.

Anyone who only knows one of the supported programming languages should be able to fix a bug or add a feature in all the other implementations. -Simply by finding their way around a consistently organised codebase.

Did you consider designing/organising the code (and tests) more similar to the Java implementation @lenntt?

lenntt:
@aslakhellesoy @mpkorstanje 

I mainly looked how `cucumber-expressions/javascript` was organized an took conventions from there, but tried to honor the java implementation as much as possible (though not finished).

I agree that the behavior and tests should be more similar. I'll work on that ~next week.

But, I don't want to introduce any unnecessary complexity. As javascript isn't statically typed like java, it doesn't need to be as verbose. Agreed?

Just curious - did you consider transpiling java to javascript?

aslakhellesoy:
> As javascript isn't statically typed like java, it doesn't need to be as verbose. Agreed?

Sure, but the overall structure should be the same.

> Did you consider transpiling java to javascript?

And Ruby and Go and other future languages? No :-)

lenntt:
Cool, I think I'm not _that_ far off - just not complete.

I hadn't really looked at `DataTableTypeRegistryTableConverter` (and tests) ealier.
It isn't documented (nor used anywhere?).

- Where does it fit in? Whats the difference between a converter and a transformer?
- Also, I tried not to break with whats already in [cucumber-js](https://github.com/cucumber/cucumber-js/blob/master/src/models/data_table.js), which seem to have a similar role (?) but is different?

mpkorstanje:
The converter is used behind the scenes by `DataTable.asX` methods. This can be used by the user to convert a data table when cucumber can't, or shouldn't.

aslakhellesoy:
> Also, I tried not to break with whats already in cucumber-js, which seem to have a similar role (?) but is different?

Don’t aim to keep this backwards compatible. Aim for an implementation that is as consistent as possible with datatable/java.

We can add backwards-compatible layer in cucumber-js later, or we can decide just to break backwards compatibility, as Cucumber-JVM 3.0.0 does.

lenntt:
@mpkorstanje Thanks, got it. 

Think the biggest feature gap is that the js implementation isn't typed. So, for example, I'm not able to specify the keys in my map are integers, without transforming to integers in my transformer myself.

Would this be acceptable to add that later? It should be possible a type system later as a non-breaking change. (and I do think it will be different from the java implementation).

I'd rather focus on polishing a minimal viable thing (tests, docs, ci -> integration in cucumber-js), as I can't estimate how much time I can spent in upcoming weeks.

@aslakhellesoy fair enough :)

lenntt:
for noise reduction, will reopen/create new one when ready. hopefully, later this/next week

lenntt:
Reopening, but still wip - starting by translating the tests (sidenote: not all are java tests are friendly-structured for translation to describe-it). Good to do, as java and javascript starting to look more like each other.

But I still have concerns and I'm not a 100% convinced we should aim for java parity, especially for a first step: javascripters aren't used to typed tables/objects nor **need** to be as strict in java. like ruby.
Therefore, the **use cases are limited.**

As a cucumber user, I'd be happy with a `defineDataTableType` where I can define a 'type' (as string-identifier?) together with a 'transform' function that gives me a `DataTable` as parameter.

All the extra's like trying to auto-convert the table (e.g. solve `{MyObject=> [MyOtherObject]}`), has very little value to me - though this is where a lot of the added complexity is (LOCs * 10 at least). I don't like writing 100s lines of code no one is going to use (or likes to maintain).

So @mpkorstanje , @aslakhellesoy , does it make sense to weaken the datatable spec for dynamically typed languages (javascript, ruby)?

mpkorstanje:
Solving my {MyObject=> [MyOtherObject]} would be over kill I imagine. Do you think {"String"=> [MyOtherObject]}  is also useless? That is what the language supports.

lenntt:
not _useless_, little value.
Perhaps because I haven't seen the datatable language full in action yet.

The few builtin converters, to support 'basic' type of tables, comparable to [cucumber-js](https://github.com/cucumber/cucumber-js/blob/master/docs/support_files/data_table_interface.md)

I'd be totally happy if I could do, for example, something like this:
```js
defineDataTableType('myOtherObjectTable', function(dataTable) {
   let myTable = dataTable.hashes()     // {String => String[]}
   Object.keys(myTable).forEach(function(key) {
       obj[key] = obj[key].map((cell) => {return new MyOtherObject(cell)}) // convert String[] to MyOtherObject[]
   });
   return myTable
});
```
The registry itself doesn't need to know about `MyOtherObject` itself.

mpkorstanje:
For Java I know that the example you've given would be a royal pain to implement. Especially if I had to do it for a few types. And while implementing the converter in Java I found that it was easier to implement the full set of theoretically possible transformations then to limit my self to the practical set. It drove out all the edge cases.

But I'm out of my depth when it comes to utility in js. If the primary uses are lists of hashes, or hashes of hashes then we wouldn't need the transformers at all in js. In which case it'd be better not to implement it at all.

Are there typescript bindings for cucumber?

lenntt:
Afaik typescript compiles to js, so you're just using the js bindings. 
I'm inexperienced with TS, but I can imagine it can be more painful like java.

What we _could_ do... implement my straightforward suggestion straight into `cucumber-js` instead?
That way, the gap between cucumber-js and cucumber grows. but:
* `cucumber` repo is kept consistent, 
* it also gives us some room for getting real use cases and something more concrete to discuss?

mpkorstanje:
I was thinking of doing it the other way around. I think I could extract out all the jvm specific stuff. Having the same basic data table implementation in a single location is a good thing.

.

mpkorstanje:
And this would also make it easier for other implementations to get started on data tables.

PR #408, opened 2018-06-12
datatable: Automatic list conversion

aslakhellesoy:
## Summary

Provide a simple (one-liner) to convert a data table to a list of objects:

```java
registry.defineDataTableType(DataTableType.entry(Author.class))
```

Also provide a simple (one-liner) to convert a cell into a "string wrapper" value object:

```java
registry.defineDataTableType(DataTableType.cell(AirportCode.class))
```

## Details

* `entry` uses Jackson Databind's `ObjectMapper` to convert a table entry (`Map<String,String>`) into an object.
* `cell` uses reflection to call a String constructor.

## Motivation and Context

Cucumber-JVM 2 used to convert tables to `List<Something>` automatically, with zero configuration. Version 3 removed XStream and added a more flexible conversion API, but some users have reported this as a regression (in ease of use). They now need to write more code to do something that used to be automatic.

This PR makes it easy again, while keeping the flexible API.

## Types of changes

## Checklist:

### DIFF
--- a/datatable/README.md
+++ b/datatable/README.md
@@ -155,7 +155,20 @@ domain. Doing this has the following benefits:
 2. Document and evolve your ubiquitous domain language
 3. Enforce certain patterns
 
-A custom table type can be registered as follows:
+There are two helpers for defining custom table types:
+
+```java
+// Defines a DataTableType` that converts an entry (map of header name to row value) 
+// to an object, using reflection.
+registry.defineDataTableType(DataTableType#entry(Class))
+
+// Defines a DataTableType that converts a single cell
+// to an object, by calling its `String` constructor (if it exists).
+registry.defineDataTableType(DataTableType#cell(Class))
+```
+
+In cases where these two reflection based helpers are insufficient,
+a custom table type can be registered as follows:
 
 ```java
 registry.defineDataTableType(
@@ -204,6 +217,13 @@ represent geolocations. Airports are represented by `Airport(code:String)`.
 ```
 
 By registering two table types:
+
+```java
+registry.defineDataTableType(DataTableType.cell(Airport.class));
+registry.defineDataTableType(DataTableType.entry(Geolocation.class));
+```
+
+Alternatively, you can implement your own types if you need more control:
 
 ```java
 registry.defineDataTableType(
--- a/datatable/README.md
+++ b/datatable/README.md
@@ -158,7 +158,7 @@ domain. Doing this has the following benefits:
 There are two helpers for defining custom table types:
 
 ```java
-// Defines a DataTableType` that converts an entry (map of header name to row value) 
+// Defines a DataTableType that converts an entry (map of header name to row value) 
 // to an object, using reflection.
 registry.defineDataTableType(DataTableType#entry(Class))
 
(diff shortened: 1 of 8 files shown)


aslakhellesoy:
This is a proposed fix for https://github.com/cucumber/cucumber-jvm/issues/1388

datatable/java/datatable/src/main/java/io/cucumber/datatable/DataTableType.java
```diff
@@ -78,6 +81,51 @@ private DataTableType(Class<?> type, Type target, RawTableTransformer<?> transfo
...
+     * @throws CucumberDataTableException if <code>type</code> does not has a public String constructor.
+     */
+    public static <T> DataTableType cell(final Class<T> type) {
+        try {
+            final Constructor<T> constructor = type.getConstructor(String.class);
+            return new DataTableType(type, new TableCellTransformer<T>() {
+                @Override
+                public T transform(String cell) {
+                    try {
+                        return constructor.newInstance(cell);
```
mpkorstanje: To the best of my knowledge `objectMapper.convertValue("someThing", SingleArgConstructor.class);` works. So you use Jackson here as well. I believe it is also a bit more flexible as it will check for methods name "fromString" as well.


 It simplifies the documentation (refer to Jackons)

aslakhellesoy:
> I am not 100% confident our shaded jackson version will handle non-shaded annotations from the same version properly.

I think that's fine. People don't *have* to use those annotations. Classes with setters and public fields work fine without annotations. If people need to use annotations, they can write their own transformer, using an unshaded Jackson or something else.

datatable/README.md
```diff
@@ -155,7 +155,20 @@ domain. Doing this has the following benefits:
...
+There are two helpers for defining custom table types:
+
+```java
+// Defines a DataTableType that converts an entry (map of header name to row value) 
+// to an object, using reflection.
+registry.defineDataTableType(DataTableType#entry(Class))
+
+// Defines a DataTableType that converts a single cell
+// to an object, by calling its `String` constructor (if it exists).
+registry.defineDataTableType(DataTableType#cell(Class))
```
mpkorstanje: Can we add an implementation note here: Something along the lines of we're using Jackson, but don't rely on anything but getters/setter/constructor/fromString/valueOf methods?

mpkorstanje:
> People don't have to use those annotations.

They will. And someone will blog about this amazing thing they discovered too. :D

Let's at least warn them.

datatable/java/datatable/src/main/java/io/cucumber/datatable/DataTableType.java
```diff
@@ -78,6 +78,42 @@ private DataTableType(Class<?> type, Type target, RawTableTransformer<?> transfo
...
+        });
+    }
+
+    /**
+     * Creates a data table type that transforms individual cells to another type, calling the type's
+     * String constructor.
+     *
+     * @param type the type of the cell
+     * @param <T>  see <code>type</code>
+     * @throws CucumberDataTableException if <code>type</code> does not has a public String constructor.
```
mpkorstanje: Not thrown anymore. If only we could test documentation. ;)

PR #422, opened 2018-06-27
Expose the optional Background element as a property

### LINKED ISSUE #421: Background not exposed in the C#/dotnet Feature type
The Background steps are not parsed as part of the Feature class.

kzu:
This makes it far more discoverable. 

Fixes #421

### DIFF
--- a/gherkin/dotnet/Gherkin/Ast/Feature.cs
+++ b/gherkin/dotnet/Gherkin/Ast/Feature.cs
@@ -1,9 +1,11 @@
 ﻿using System.Collections.Generic;
+using System.Linq;
 
 namespace Gherkin.Ast
 {
     public class Feature : IHasLocation, IHasDescription, IHasTags
     {
+        public Background Background => Children.OfType<Background>().SingleOrDefault();
         public IEnumerable<Tag> Tags { get; private set; }
         public Location Location { get; private set; }
         public string Language { get; private set; }


aslakhellesoy:
This will be obsolete when Gherkin.NET is updated for #416, so i don’t think we should add this now.

kzu:
@aslakhellesoy it just makes the API more discoverable. The `Children` thing isn't, and doing `Of<ScenarioDefinition>` and `Of<Background>` for something that are the two core use cases for the `Feature` type itself, doesn't make a lot of sense. I'd expect to "dot into" properties to discover what's there, and having a 1..1 mapping between the few concepts Gherkin provides to top-level properties in the object model, would make the API easier to use. 

When processing the gherkin file via the object model, you need to aggregate steps from the background to each of the scenarios, so you'd typically iterate scenarios, and prepend the steps from background, so it would make sense to make both first-class properties on Feature.

aslakhellesoy:
Can you tell us a bit more about what sort of software you're building that needs to use this? I might be able to suggest an alternative way to solve the problem you're trying to solve.

On a different note, with #416 we're switching to [protocol buffers](https://developers.google.com/protocol-buffers/) to represent the AST as well as pickles. Take a look at [messages.proto](https://github.com/cucumber/cucumber/blob/master/messages/messages.proto).

The generated classes will let you write code like this in Java (and, I expect, similarly in C#):

https://github.com/cucumber/cucumber/blob/7f826f0a719231066b20b5917c6c783c5ca6de3c/gherkin/java/src/main/java/gherkin/pickles/PickleCompiler.java#L51-L64

aslakhellesoy:
I think I found it: https://github.com/kzu/Gherkinator

So, when we move to protobuf, you'll have a slightly different AST API, which I hope will give you what you need.

Have you considered using Gherkin's Pickle compiler to process the AST? It takes care of traversal of the AST, inlinging background steps and expanding Scenario Outlines. And the final `Pickle` model is easier to deal with. It's what Cucumber uses.

kzu:
If there's no .NET "binding" for it, it's a no-go for my little thingy :(

PR #423, opened 2018-06-28
Add ParameterType.fromEnum(Class enumClass)

aslakhellesoy:
## Summary

Simplify creation of `ParameterType` from an enum class

## Details

The enum constants' names are used to build a regexp for the match (`red|green|blue`), and the name is the simplified name of the enum class (`Color`).

## Motivation and Context

See https://github.com/cucumber/cucumber-jvm/issues/1393#issuecomment-400985459

## Types of changes

## Checklist:

### DIFF
--- a/cucumber-expressions/java/src/test/java/io/cucumber/cucumberexpressions/EnumParameterTypeTest.java
+++ b/cucumber-expressions/java/src/test/java/io/cucumber/cucumberexpressions/EnumParameterTypeTest.java
@@ -2,22 +2,28 @@ package io.cucumber.cucumberexpressions;
 
 import org.junit.Test;
 
+import java.util.List;
+import java.util.Locale;
+
 import static java.util.Collections.singletonList;
 import static org.junit.Assert.assertEquals;
 
 public class EnumParameterTypeTest {
-    @Test
-    public void constructs_enum() {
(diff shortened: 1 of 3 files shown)


cucumber-expressions/java/src/test/java/io/cucumber/cucumberexpressions/EnumParameterTypeTest.java
```diff
@@ -2,22 +2,28 @@
...
+        meh,
+        sad
     }
 
-    public enum Color {
-        RED, BLUE, YELLOW
+    @Test
+    public void converts_to_enum() {
+        ParameterTypeRegistry registry = new ParameterTypeRegistry(Locale.ENGLISH);
+        registry.defineParameterType(ParameterType.fromEnum(Mood.class));
```
mpkorstanje: Is this a parameter type *from*Enum or *for*Enum?

When using a static import I think `registry.defineParameterType(forEnum(Mood.class));`  reads a bit better.

mpkorstanje: OTOH: fromEnum is kinda the well known standard.

PR #424, opened 2018-07-02
Remove the Java Gherkin parser. Rely on Go binary/protobuf

aslakhellesoy:
## Summary

The Java Gherkin parser is gone!

## Details

Parsing is delegated to a subprocess (gherkin-go executable)
The package has moved to `io.cucumber.gherkin`

## Motivation and Context

Less code to maintain.

## TODO

- [x] Package prebuilt go binaries for various platforms
- [x] Detect platform and extract the right binary
- [x] Detect more platforms - see [protoc-jar](https://github.com/os72/protoc-jar/blob/master/src/main/java/com/github/os72/protocjar/PlatformDetector.java)
- [x] Download binary from S3 if not in jar
- [x] Set up build process


aslakhellesoy:
Currently, in order to build this, you have to:

```
cd gherkin/go
make cross-compile
cd ../java
mkdir gherkin-go && cp ../go/dist/*
```

We can automate that for the monorepo build, but for the subrepo build, `../go` is not available. I think the best is to download the binaries from GitHub in that case. The `publish-release` target in `gherkin/go` can do the upload.

mlvandijk:
fyi, `make cross-compile` gave me an error; command 'gox' not recognized (or something similar).
Adding the following to my `.bash_profile` fixed it:
# Add go bin to path
export PATH=$(go env GOPATH)/bin:${PATH}

mlvandijk:
Also `cp ../go/dist/* ./gherkin-go` to copy the files to the created dir

aslakhellesoy:
You shouldn't have to do that @mlvandijk - `MagicFile` will download from S3.

PR #426, opened 2018-07-04
Gherkin cross compiled binary compression option in Makefile for gherkin-go

l3pp4rd:
## Summary

Uses [upx](https://upx.github.io/) to compress cross compiled gherkin-go-cli binaries.
It skips compression for unsupported processor architectures and compresses all supported.
In result around **70%** of size is reduced for every supported executable.

## Risks

I've used **upx** for around a year compressing linux and windows based binaries, have experienced an issue once, when produced binary was broken. Though, the reason it seems was in golang toolchain and it was fixed soon. It would be best if we could test and validate produced binaries. UPX provides the flag `-t` to test the integrity, but not sure whether this fully validates the binary or it is not false positive:

Says darwin binaries are corrupt. Though, probably need to test it more if size reduction matters a lot.


aslakhellesoy:
@xtrasimplicity how do you suggest we do that?

aslakhellesoy:
Superseded by #428

PR #429, opened 2018-07-06
[datatable] Add ability to register default transformers for table cell and entry

lsuski:
## Summary

This PR resolves https://github.com/cucumber/cucumber-jvm/issues/1388

## Details

`cucumber.api.TypeRegistry` has new methods for registering default data table transformers.  Those transformers receives `Class<T>` of expected type (declared in step method parameter `List<T>` or `List<List<T>>`). 
Custom `TypeRegistryConfigurer` can be implemented as follows (Kotlin):
```
override fun configureTypeRegistry(p0: TypeRegistry) {
        p0.setDefaultDataTableEntryTransformer(SimpleDefaultDataTableEntryTransformer(object : SimpleDefaultDataTableEntryTransformer.InstanceCreator {
            override fun <T : Any?> createInstance(p0: Class<T>): T {
                return //create empty instance
            }
        }))
    }
```

## Motivation and Context

Cucumber 3.0.0 removed automatic conversion from DataTable to List<MyCustomRowClass>. For large projects with lots of row classes migration from 2.4.0 is hard, implementing scenarios require more boilerplate than before (need for registering every new row class).

## How Has This Been Tested?

I've added unit tests and tested it on my Android project

## Types of changes

## Checklist:

### DIFF
--- a/datatable/java/datatable/src/main/java/io/cucumber/datatable/DataTableTypeRegistry.java
+++ b/datatable/java/datatable/src/main/java/io/cucumber/datatable/DataTableTypeRegistry.java
@@ -16,6 +16,8 @@ import static java.lang.String.format;
 public final class DataTableTypeRegistry {
 
     private final Map<JavaType, DataTableType> tableTypeByType = new HashMap<>();
+    private DefaultDataTableEntryTransformer defaultDataTableEntryTransformer;
+    private TableCellByTypeTransformer defaultDataTableCellTransformer;
 
     public DataTableTypeRegistry(Locale locale) {
         NumberFormat numberFormat = NumberFormat.getNumberInstance(locale);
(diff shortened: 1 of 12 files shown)


datatable/java/datatable/src/test/java/io/cucumber/datatable/DataTableTypeRegistryTest.java
```diff
@@ -37,8 +46,132 @@ public Place transform(DataTable table) {
...
+
+
+    @SuppressWarnings({"unchecked", "ConstantConditions"})
+    @Test
+    public void returns_default_data_table_type_for_cell_if_none_match_and_default_registered() {
+
+        registerCellEntryAndDefaults();
+
+        String here="here";
+        DataTableType lookupTableTypeByTypeForCell = registry.lookupTableTypeByType(constructTypeForCell());
```
mpkorstanje: From reading this test it is not immediately clear that `constructTypeForCell` creates a list of lists of Place type. Please use the same convention as used in `DataTableTypeRegistryTableConverterTest`.

datatable/java/datatable/src/test/java/io/cucumber/datatable/DataTableTypeRegistryTest.java
```diff
@@ -37,8 +46,132 @@ public Place transform(DataTable table) {
...
+        List<List<Place>> transformedCells = (List<List<Place>>) lookupTableTypeByTypeForCell.transform(singletonList(singletonList(here)));
+
+        assertEquals(1,transformedCells.size());
+        assertEquals(1,transformedCells.get(0).size());
+        assertEquals(here,transformedCells.get(0).get(0).name);
```
mpkorstanje: This can be replaced with a single assert against a list of lists of places. It is okay to make place implement equals and hashcode.

datatable/java/datatable/src/test/java/io/cucumber/datatable/DataTableTypeRegistryTest.java
```diff
@@ -37,8 +46,132 @@ public Place transform(DataTable table) {
...
+    }
+
+    private void registerCellEntryAndDefaults()  {
+        registry.defineDataTableType(DataTableType.cell(DataTableTypeRegistryTest.class));
+        registry.defineDataTableType(DataTableType.entry(DataTableTypeRegistryTest.class));
+        registerDefaultCellTransformer();
+        registerDefaultEntryTransformer();
+    }
+
+    private void registerDefaultCellTransformer() {
```
mpkorstanje: Please use the same convention as used in `DataTableTypeRegistryTableConverterTest`. You can make a constant out of the anonymous `TableCellByTypeTransformer` and put the actual registration call in the test. This way nothing his hidden by calling it "Default".

datatable/java/datatable/src/test/java/io/cucumber/datatable/DataTableTypeRegistryTest.java
```diff
@@ -37,8 +46,132 @@ public Place transform(DataTable table) {
...
+
+    private JavaType constructTypeForCell() {
+        return aListOf(aListOf(Place.class));
+    }
+
+    private JavaType constructTypeForEntry() {
+        return aListOf(Place.class);
+    }
+
+    private void registerCellEntryAndDefaults()  {
```
mpkorstanje: This method is hiding quite a bit of non-trivial test setup and it isn't immediately clear how the tests that use it are affected by it. Can you change the structure of the tests/code so that it is?

datatable/java/datatable/src/main/java/io/cucumber/datatable/SimpleDefaultDataTableEntryTransformer.java
```diff
@@ -0,0 +1,71 @@
...
+
+	private CamelCaseStringConverter fieldNameConverter = new CamelCaseStringConverter();
+	private InstanceCreator instanceCreator;
+
+	public SimpleDefaultDataTableEntryTransformer(InstanceCreator instanceCreator) {
+		this.instanceCreator = instanceCreator;
+	}
+
+	@Override
+	public <T> T transform(Map<String, String> value, Class<T> type, TableCellByTypeTransformer cellTransformer) throws Throwable {
```
mpkorstanje: Why not use the ObjectMapper from Jackson instead? It has plugins that allow you to change the name-to-field conversion.

datatable/java/datatable/src/main/java/io/cucumber/datatable/DataTableTypeRegistry.java
```diff
@@ -35,47 +37,61 @@ public BigDecimal transform(String cell) {
...
             @Override
             public Integer transform(String cell) {
                 return Integer.decode(cell);
             }
-        }));
+        };
+        defineDataTableType(new DataTableType(Integer.class, integerTableCellTransformer));
 
-        defineDataTableType(new DataTableType(Long.class, new TableCellTransformer<Long>() {
+        defineDataTableType(new DataTableType(int.class, integerTableCellTransformer));
```
mpkorstanje: Please make sure the whitespace usage is consistent between all `defineDataTableType` calls for the same type.

datatable/java/datatable/src/main/java/io/cucumber/datatable/DataTableTypeRegistry.java
```diff
@@ -35,47 +37,61 @@ public BigDecimal transform(String cell) {
...
-        defineDataTableType(new DataTableType(Byte.class, new TableCellTransformer<Byte>() {
+        TableCellTransformer<Byte> byteTableCellTransformer = new TableCellTransformer<Byte>() {
             @Override
             public Byte transform(String cell) {
                 return Byte.decode(cell);
             }
-        }));
+        };
+        defineDataTableType(new DataTableType(Byte.class, byteTableCellTransformer));
+        defineDataTableType(new DataTableType(byte.class, byteTableCellTransformer));
```
mpkorstanje: Neat!

lsuski:
Regarding `TableCellByTypeTransformer`:
when I started implementing `DefaultDataTableTransformer` I realized that I don't want to provide any default implementation for cell and I wanted to avoid something like `throw new UnsupportedOperationException`. The second thing was that I needed cell transformer for transforming `String` to expected field type. I agree that name of both should be similar.

datatable/java/datatable/src/main/java/io/cucumber/datatable/SimpleDefaultDataTableEntryTransformer.java
lsuski: I'm not Jackson expert but I couldn't make it work with class without any annotations. Eventually I'd like to use it on Android and use `Unsafe` or reflection to instantiate object using constructor with default null values. `SimpleDefaultDataTableEntryTransformer` does not need to be part of datatable project if this is a problem, I just wanted to provide simple implementation for people, but can remove it from here

mpkorstanje: No worries. Ill try and come up with a nice way to structure this.

mpkorstanje:
Sounds reasonable. But I would like to limit the configuration required and I am expecting that most uses can be covered by a single mapper.

I'll try and come up with something

mpkorstanje:
Cheers!

PR #435, opened 2018-07-12
Refactor down Gherkin::Messages::* to Gherkin::*

jaysonesmith:
## Summary

Collapsing out the messages module for naming/usage

## Motivation and Context

Let's make naming and using things easier!

## How Has This Been Tested?

make


gherkin/ruby/lib/gherkin/gherkin.rb
```diff
@@ -0,0 +1,23 @@
+require 'open3'
+require 'gherkin/protobuf_cucumber_messages'
+require 'gherkin/exe_file'
+
+module Gherkin
+  class Gherkin
+    def initialize(paths, print_source, print_ast, print_pickles)
+      @paths, @print_source, @print_ast, @print_pickles = paths, print_source, print_ast, print_pickles
+      path_pattern = "#{File.dirname(__FILE__)}/../../gherkin-go/gherkin-{{.OS}}-{{.Arch}}{{.Ext}}"
```
jaysonesmith: I think this is causing troubles. If I leave with `/../../../` it doesn't pass locally, but now it's failing in Travis.

gherkin/ruby/Makefile
```diff
@@ -14,7 +14,7 @@ ERRORS       = $(patsubst testdata/%.feature,acceptance/testdata/%.feature.error
 .tested: .compared
 
 .deps:
-	./scripts/download-gherkin-go $(LIBRARY_VERSION)
+	# ./scripts/download-gherkin-go $(LIBRARY_VERSION)
```
aslakhellesoy: Uncomment please

gherkin/ruby/lib/gherkin/gherkin.rb
aslakhellesoy: My guess is it’s failing because you disabled the download of the executables.

jaysonesmith:
@aslakhellesoy Looks like you were right about it being due to the disabled downloads section. I had deleted your comment saying so as I thought GH would require the code that you commented on to be changed, but I guess not. Merging!

PR #444, opened 2018-07-31
gherkin: Cucumber-Ruby integration

brasmusson:
## Summary

Changes for the Cucumber-Ruby integration.

## Details

- [X] Provide an interface for passing the feature file content as a string. The testing of Cucumber-Ruby should not require that the test feature files are written to the file system. In version 5.x.x there is an interface for passing the feature file content as a string.
- [X] Provide the possibility to set the default dialect for the gherkin parser (needed for the parsing of nested steps in Cucumber-Ruby).
- [X] ~Implement the `--dialects` options correctly in the go binary~ Use the `--dialects` option (that only works in the binaries built for dist), so that `gherkin/ruby/lib/gherkin/gherkin-languages.json` can be removed.  

## Motivation and Context

See Details.

## How Has This Been Tested?

The automated test are updated to cover this functionality

## Types of changes

## Checklist:


gherkin/ruby/lib/gherkin/gherkin.rb
```diff
@@ -1,23 +1,57 @@
...
       path_pattern = "#{File.dirname(__FILE__)}/../../gherkin-go/gherkin-go-{{.OS}}-{{.Arch}}{{.Ext}}"
       @gherkin_executable = ExeFile.new(File.expand_path(path_pattern)).target_file
     end
 
     def messages
+      args = base_args
+      args = args.concat(@paths)
+      stdin, stdout, stderr, wait_thr = Open3.popen3(*args)
+      stdin.close
+      messages = ProtobufCucumberMessages.new(stdout, stderr).messages
```
olleolleolle: Does this assignment (and the same one in the next method) do anything?

gherkin/ruby/lib/gherkin/gherkin.rb
```diff
@@ -1,23 +1,57 @@
...
+    end
+
+    def parse(uri, data)
+      args = base_args
+      stdin, stdout, stderr, wait_thr = Open3.popen3(*args)
+      stdin.binmode
+      stdin.write(encode_source_message(uri, data))
+      stdin.close
+      messages = ProtobufCucumberMessages.new(stdout, stderr).messages
```
olleolleolle: (This assignment, does it do anything?)

gherkin/ruby/lib/gherkin/gherkin.rb
```diff
@@ -1,23 +1,57 @@
...
       args = [@gherkin_executable]
       args.push('--no-source') unless @print_source
       args.push('--no-ast') unless @print_ast
       args.push('--no-pickles') unless @print_pickles
-      args = args.concat(@paths)
-      stdin, stdout, stderr, wait_thr = Open3.popen3(*args)
-      ProtobufCucumberMessages.new(stdout).messages
+      args.push("--default-dialect=#{@default_dialect}") unless @default_dialect.nil?
+      args
+    end
```
olleolleolle: 👏 These extracted method are mega-helpful.

gherkin/ruby/lib/gherkin/gherkin.rb
```diff
@@ -1,23 +1,57 @@
...
+      media_obj.encoding = 'UTF-8'
+      media_obj.content_type = 'text/x.cucumber.gherkin+plain'
+      source_obj = Cucumber::Messages::Source.new
+      source_obj.uri = uri
+      source_obj.data = data
+      source_obj.media = media_obj
+      wrapper_obj = Cucumber::Messages::Wrapper.new
+      wrapper_obj.source = source_obj
+      buf = Cucumber::Messages::Wrapper.encode(wrapper_obj)
+      buf[1..-1] # A leading "\n" needs to be removed
```
aslakhellesoy: Why is there an extra `\n` at the beginning?

brasmusson:
Now only the proper implementation of the `--dialects` option in the go binary is needed.

gherkin/ruby/lib/gherkin/gherkin.rb
aslakhellesoy: I found the problem. Protobuf messages written to gherkin-go's STDIN should not be wrapped in a Wrapper object. The Wrapper object is only necessary when the reader expects different kinds of messages.

aslakhellesoy:
I updated the API a little to conform with the API in Java/JavaScript. Use this instead:

```ruby
messages = Gherkin.from_paths(paths)
```

PR #447, opened 2018-08-03
Update gherkin-languages.json

stevenhuijgens:
Extended dutch 'then' to improve on grammar

## Summary
Extended dutch 'then' to improve on grammar

## Motivation and Context
Steps as 'Given / When / Then / And the window is opened and active' translate into 'Gegeven / Als / Dan / En het venster actief en geopend is'. 'Dan' should be extended to increase readability. 'Verwacht dat / Valideer dat het venster actief en geopend is' is correct.

## Types of changes

## Checklist:

### DIFF
--- a/gherkin/gherkin-languages.json
+++ b/gherkin/gherkin-languages.json
@@ -2362,7 +2362,9 @@
     ],
     "then": [
       "* ",
-      "Dan "
+      "Dan ",
+      "Verwacht dat ",
+      "Valideer dat "
     ],
     "when": [
       "* ",


mlvandijk:
Thoughts @mpkorstanje?

mpkorstanje:
Both `Verwacht` and `Valideer` describe actions, not state. Translated to English it would equivalent to adding `Expect` and `Assert that` which is test-language, not scenario language.

I think the problem 

The example could be rewritten to

```
Dan is het venster actief en geopend
```

aslakhellesoy:
When and Then are temporal adverbs. Translations should be too (at least for indo-european languages, with similar grammar).

This proposal doesn’t sound right to me.

mlvandijk:
Agreed with @mpkorstanje and @aslakhellesoy

aslakhellesoy:
Thanks for your contribution @stevenhuigens - we’re not going to accept this, as it diverges too much from the established language.

PR #449, opened 2018-08-05
gherkin: ruby,go: Failing test for strange size limitiation for feature files

brasmusson:
## Summary

Between [886c1c4...285b244](https://github.com/cucumber/cucumber/compare/886c1c4...285b244) there seems to have appeared a strange size limitation for feature files when passing them as source (instead of as path) to the gherkin parser.

## Details

The failing test says that a feature file with 49 minimal scenarios works as expected (51 messages produced by the parser) but a feature file with 50 minimal scenario does not work (no messages produced by the parser). It does not occur when passing the feature file as path (and it is not specifically about 50 scenarios), since when a use https://github.com/cucumber/cucumber-ruby/blob/master/features/docs/formatters/junit_formatter.feature from the `gherkin_spec.rb` and send it by path the parser produces 12 messages, but when sending its content as source the parser produces 0 messages.

## Types of changes

## Checklist:

### DIFF
--- a/gherkin/go/gherkin_messages.go
+++ b/gherkin/go/gherkin_messages.go
@@ -6,6 +6,7 @@ import (
 	gio "github.com/gogo/protobuf/io"
 	"io"
 	"io/ioutil"
+	"math"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


brasmusson:
I ran `git bisect` and the message was that d4dae16 is the first bad commit.

brasmusson:
`test.sh` of a0d4e9a can be used to check if `bin/gherkin` can handle source messages for "large" feature files on stdin.

aslakhellesoy:
This could be a bug in gogo/protobuf, or it could simply be that it’s not used correctly.

My motivation for switching to gogo was its built-in support for varint-delimited streams. The official protobuf leaves that entirely to the app developer.

Perhaps it’s safer to switch back to the official one.

Issue #455, opened 2018-08-14
Incorrect variable names in generated cucumber expressions


aslakhellesoy:
When I define a parameter type with name `country-code` and have a snippet generated, the variable name is the same as the parameter type name (`country-code`). This is not valid Java, nor is it conventional casing. It should be `countryCode`.

We should remove common invalid characters, and add an enum option for `camelcase` or `snakecase`. (In Java it will always be camelCase, but the Go library should use the option - when used inside cucumber-engine it will have to support both, depending on the front-end language).

aslakhellesoy:
Shut up stalebot - gimme some time!

mlvandijk:
Seems like this should be a slow burner ;)

mpkorstanje:
For people looking to pick this up the relevant code in java can be found in `GeneratedExpression.getParameterName` and for go `getParameterName` in `generated_expression.go`.

mpkorstanje:
> . (In Java it will always be camelCase, but the Go library should use the option - when used inside cucumber-engine it will have to support both, depending on the front-end language).

I think we can leave this out for now. We're not planning to use Cucumber Engine so only camel case should be sufficient.

mpkorstanje:
Because this is part of the snippet generation and because it shares several common parts with snippet generation this has to be solved in in a language specific implementation. Probably not worth extracting it though.

Moving this to JVM.

PR #462, opened 2018-08-22
WIP: cucumber-expressions: java: Fix for #455

aslakhellesoy:
This is a fix for #455

Introduces `LetterCase` and `ProgrammingLanguage` enums that can be used to configure how parameter names in snippets are generated. We should port this over to Go (and add the following enum values):

* `LetterCase`
  * `CAMEL_CASE`
  * `LOWER_CAMEL_CASE`
  * `SNAKE_CASE`
* `ProgrammingLanguage`
  * `JAVA`
  * `RUBY`
  * `GO`
  * `JAVASCRIPT`

This way, [cucumber-engine](https://github.com/cucumber/cucumber-engine) will be able to generate expressions for multiple languages.

We do not need to port this code over to Ruby or JavaScript, as the plan is to adopt cucumber-engine. See the [roadmap](https://docs.google.com/document/d/12Y3Ut-SVSGdw9L_aW6NGQY9JaCcn3z0alY4d7kzOndU/) for details.

### DIFF
--- a/cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/GeneratedExpression.java
+++ b/cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/GeneratedExpression.java
@@ -1,39 +1,17 @@
 package io.cucumber.cucumberexpressions;
 
-import java.text.Collator;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


aslakhellesoy:
Still relevant - should be ported to Go.

mpkorstanje:
We mothballed the cucumber-engine. This fix is better done in combination with snippet generation in a specific language implementation.

PR #472, opened 2018-09-08
java: Java Platform Module System using Automatic-Module-Name

nhojpatrick:
## Summary

Define Automatic-Module-Names for all jar files generated so that they can be used within the Java Platform Module System, instead of being put into the UNNAMED bucket or using filename based module names.

## Details

Change only the build so each jar created has a 'Automatic-Module-Name' added into the META-INFO/MANIFEST.MF file.

Similar changes are being applied to cucumber-jvm, see PR https://github.com/cucumber/cucumber-jvm/pull/1451

## Motivation and Context

Trying to write some cucumber frameworks (state between steps, object to json conversion, json manipulation, calling rest endpoints sending/receiving json), as i rewrite this type of thing at each client i work for so want to write it once and feed back to cucumber as extra so when i start my next contract i can reuse instead of write from scratch.

I'm trying to write a pure java 9/10/11 build and getting issues with the framework I've written as they are UNNAMED modules, or modules based upon the jar file name which isn't a good idea.

## How Has This Been Tested?

No code changed, only jar file manifest altered.

## Types of changes

## Checklist:


mpkorstanje:
It is worth noting that the current version of Gherkin is currently planned to go into cucumber jvm v5 because it includes a significant amount of breaking changes. If we want to resolve this module as part of a v4 release we'll have to back fix Gherkin.

nhojpatrick:
Regarding the move, are you talking about io.cucumber.datatable.DataTableHasTheSameRowsAs, it can't move to datatable because it depends upon io.cucumber.datatable.matchers.*.

I've added TODO.

@mpkorstanje if you where talking about another class/file please confirm which.

regarding Gherkin which tag should i propose patch?

mpkorstanje:
> Regarding the move, are you talking about io.cucumber.datatable.DataTableHasTheSameRowsAs, it can't move to datatable because it depends upon io.cucumber.datatable.matchers.*.

Yeah I really don't know how to untangle this. I am considering dropping the matchers module. But I'll have to look at the downloads/usage.

> @mpkorstanje if you where talking about another class/file please confirm which.

I believe cucumber-jvm is currently using 5.1.0 and 6.0.0 will be the version used in. So patching 5.1.0 would be helpful.

nhojpatrick:
> > Regarding the move, are you talking about io.cucumber.datatable.DataTableHasTheSameRowsAs, it can't move to datatable because it depends upon io.cucumber.datatable.matchers.*.
> 
> Yeah I really don't know how to untangle this. I am considering dropping the matchers module. But I'll have to look at the downloads/usage.
> 
> > @mpkorstanje if you where talking about another class/file please confirm which.
> 
> I believe cucumber-jvm is currently using 5.1.0 and 6.0.0 will be the version used in. So patching 5.1.0 would be helpful.

have done individual PR's for each module separately based upon the version that cucumber-jvm is currently using, so cucumber-expressions, datatable, gherkin, tag-expressions

datatable/java/matchers/src/main/java/io/cucumber/datatable/DataTableHasTheSameRowsAs.java
```diff
@@ -5,6 +5,7 @@
 /**
  * @deprecated use {@link io.cucumber.datatable.matchers.DataTableHasTheSameRowsAs}
```
mpkorstanje: I can't find `io.cucumber.datatable.matchers.DataTableHasTheSameRowsAs`.

mpkorstanje:
> have done individual PR's for each module separately based upon the version that cucumber-jvm is currently using, so cucumber-expressions, datatable, gherkin, tag-expressions

Cheers. Fixing Gherkin might be a bit to much but upgrading tag expressions and datatable (and fixing it's package structure) should be possible.

The one for Gherkin will be a bit odd because it has to be done manually.

datatable/java/matchers/src/main/java/io/cucumber/datatable/DataTableHasTheSameRowsAs.java
mpkorstanje: Owh. I wrote it ages ago. Nevermind!

mpkorstanje:
I've merged this just to get it out of the way. It means that going forward we'll have our modules for most of the libraries. I'll have to see about back fixing the others.

PR #482, opened 2018-09-23
Automatically release maven artefacts

mpkorstanje:
Currently artefacts bundles are deployed to nexus but not yet released.
This requires an additional step that can only be performed by a few
people.

Automatically releasing maven artifacts removes this bottleneck.


mpkorstanje:
Not sure why the build is failing.

aslakhellesoy:
This is obsolete after we started inheriting from `cucumber-parent`

PR #486, opened 2018-09-30
cucumber-messages: Allow compatible version of google-protobuf with JRuby

mvz:
## Summary

There is no Java release of google-protobuf version 3.6.1. The latest working Java version seems to be 3.2.0.2. Allow JRuby users to select this older version by loosening the dependency on google-protobuf.

## Details

Loosens the gem dependency as declared in the gemspec.

## Motivation and Context

Builds with Cucumber 4.0.0.rc.1 currently fail on JRuby because no compatible version of google-protobuf can be installed.

## How Has This Been Tested?

I ran the specs and installed `cucumber-messages` locally on JRuby with the older `google-protobuf` version (3.2.0.2). I then was able to ~run Cucumber scenarios with~ install cucumber 4.0.0.rc.1 under JRuby.

## Types of changes

- Bug fix (non-breaking change which fixes an issue).

## Checklist:

### DIFF
--- a/cucumber-messages/ruby/Gemfile
+++ b/cucumber-messages/ruby/Gemfile
@@ -2,4 +2,5 @@
 source "https://rubygems.org"
 gemspec
 
+# Pick a known-good version when running builds for cucumber-messages on JRuby.
 gem 'google-protobuf', '~> 3.2.0.2', platform: :jruby
--- a/cucumber-messages/ruby/Gemfile
+++ b/cucumber-messages/ruby/Gemfile
@@ -1,3 +1,5 @@
 # frozen_string_literal: true
 source "https://rubygems.org"
 gemspec
+
+gem 'google-protobuf', '~> 3.2.0.2', platform: :jruby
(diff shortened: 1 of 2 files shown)


mvz:
Note: Actually *running* cucumber 4 under JRuby fails.

mvz:
I should have explained this better:

* The change in the gemspec allows any user of `cucumber-messages` the choice of picking a version of `google-protobuf` that works on their platform. If they use JRuby, they would probably pick 3.2.0.2.
* The change in the Gemfile is only for picking a working `google-protobuf` version while developing `cucumber-messages` on JRuby.

This solution is not ideal since the latest release of `google-protobof` with JRuby support is in fact broken.

aslakhellesoy:
Oh I see. Thanks for clarifying! Can we keep some of this knowledge in comments?

Feel free to merge!

gongo:
When is this fix released? (2.1.2 or 2.2.0 ?)

PR #488, opened 2018-09-30
cucumber-messages: Add testResult to the TestCaseFinished message

brasmusson:
## Summary

The TestCaseFinished message should have a testResult field.

## Details

Also the TestCaseFinished message should have a testResult field, not only the TestStepFinished.

## Motivation and Context

Since the summary is printed:
```
xxx scenarios (x1 Passed, ...)
yyy steps (y1 Passed, ...)
```
the concept of test cases having result clearly exists. It should be the responsibility of the cucumber-engine to calculate that result, not each and every listener to the message stream. The cucumber-engine need to know the result of the test case (so far) anyway to be able to decide how to treat the next step (run, skip ...).

## How Has This Been Tested?

Still WIP

## Types of changes

Is adding a field to a message a "new feature" or a "breaking change"?

## Checklist:

### DIFF
--- a/cucumber-messages/go/messages.proto
+++ b/cucumber-messages/go/messages.proto
@@ -214,7 +214,8 @@ message TestCaseStarted {
 
 message TestCaseFinished {
   string pickleId = 1;
-  google.protobuf.Timestamp timestamp = 2;
+  TestResult testResult = 2;
+  google.protobuf.Timestamp timestamp = 3;
 }
 
 message TestStepStarted {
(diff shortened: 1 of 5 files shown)


aslakhellesoy:
I think this makes sense.

@brasmusson there is a script to sync copies of the master `messages.proto` file into module directories.

    source scripts/functions.sh
    rsync_files

Could you try running that and commit the changes?
@jaysonesmith tried doing this yesterday, but the command gave no output and didn't modify any files, so there is a possibility you'll have the same problem. It works fine on my machine™.

Issue #493, opened 2018-10-05
Multiple enum ParameterType generates exception


abhishek8908:
I am trying to register more than one enum class in the parameter type:

```
public class TypeRegistryConfiguration implements TypeRegistryConfigurer {

    @Override
    public Locale locale() {
        return ENGLISH;
    }

    @Override
    public void configureTypeRegistry(TypeRegistry typeRegistry) {
        typeRegistry.defineParameterType(ParameterType.fromEnum(Color.class));
        typeRegistry.defineParameterType(ParameterType.fromEnum(Shape.class));
    }
}
```

> Exception:
> I couldn't decide which one to use. You have two options:
> 
> 1) Use a Cucumber Expression instead of a Regular Expression. Try one of these:
>    verify {Color}
>    verify  {Shape}
> 
> 2) Make one of the parameters types preferential and continue to use a Regular Expression.

Could you please suggest how to make this enum selection dynamic?

This conversion was not required in ver 2.4

mpkorstanje:
Could you include the full stack trace, contents of the enums and your step definitions?

abhishek8908:
Sure

Stack trace:

> io.cucumber.cucumberexpressions.AmbiguousParameterTypeException: Your Regular Expression /^assert color (CIRCLE|SQUARE)/
> matches multiple parameter types with regexp /CIRCLE|SQUARE/:
>    {SHAPE}
>    {TYPE}
> 
> I couldn't decide which one to use. You have two options:
> 
> 1) Use a Cucumber Expression instead of a Regular Expression. Try one of these:
>    assert shape {SHAPE}
>    assert shape {TYPE}
> 
> 2) Make one of the parameter types preferential and continue to use a Regular Expression.
> 
> 
> 	at io.cucumber.cucumberexpressions.ParameterTypeRegistry.lookupByRegexp(ParameterTypeRegistry.java:129)
> 	at io.cucumber.cucumberexpressions.RegularExpression.match(RegularExpression.java:32)
> 	at io.cucumber.stepexpression.StepExpression.match(StepExpression.java:21)
> 	at io.cucumber.stepexpression.ExpressionArgumentMatcher.argumentsFrom(ExpressionArgumentMatcher.java:20)
> 	at cucumber.runtime.java.JavaStepDefinition.matchedArguments(JavaStepDefinition.java:53)
> 	at cucumber.runner.Glue.stepDefinitionMatches(Glue.java:104)
[...]

PR #496, opened 2018-10-08
Anonymous Parameter Types

mpkorstanje:
## Summary

Introduces the anonymous parameter to cucumber expressions. 
## Details

Installs the anonymous parameter type into the registry:
```java
ParameterType(
  name: ""
  regexps: [\.*\]
  transformer: s -> s
  useForSnippets: false
  preferForRegexMatch: true
)
```
For Javascript and Ruby this will merely returns the captured text as a string.

However for strongly typed languages like Java and Go this allows the step definition to hint the type to which the captured text should be converted. When used in conjunction with an object mapper this makes it easier to quickly define step definitions without the need to define parameter types for each type. For example in Java the following will work without the need to configure additional parameter types.

```java
Given("^there is some date (.*)$", (Date a) -> { })

Given("there is some date {}", (Date a) -> { })

Given("there is some transaction {} {}", (BigDecimal a, Currency b) -> { })

Given("there is some type {}", (TypeEnum a) -> { })
```

To make the dynamic transform possible type hints are passed to the cucumber expression when matching a step. Using these hints the `s -> s` transform is replaced with `s -> defaultTransform.transform(s, typeHint)`. A simple implementation of the `defaultTransform` is provided that supports a selection of primitives but should generally speaking be replaced with a proper object mapper.

Fixes: #493

## Motivation and Context

To make it easier to upgrade from Cucumber-JVM v2 to v4. See https://github.com/cucumber/cucumber-jvm/issues/1478 for details.

## Types of changes

## Checklist:

Before merging this, the code should be ported to all languages:

* [x] Java
* [x] Go
* [x] JavaScript
* [x] Ruby


cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/ParameterTypeRegistry.java
```diff
@@ -26,69 +26,147 @@
...
+    private final ObjectMapper objectMapper;
+
+    public interface ObjectMapper {
+        <T> T convert(String fromValue, Class<T> toValueType);
+
+        Object convert(String fromValue, Type toValueType);
+    }
+
+
+    private static final class SimpleObjectMapper implements ObjectMapper {
```
mpkorstanje: We can make this configurable.

aslakhellesoy:
Haven’t had a close look yet, but since it seems a bit involved I’m reminding you it would have to be ported to the other 3 languages.

aslakhellesoy:
I pushed a few more commits @mpkorstanje after our pairing session today. Also find the bug (we incremented the typeIndex twice!) 

Let me know what you think!

mpkorstanje:
I extracted the `ObjectMapper` interface. I am not happy about the name yet because it only maps `Strings` to `Objects`. Maybe you know a better name.

Now I am trying to define an anonymous parameter type for the `CucumberExpressions`

An anonymous parameter type would be used as:

```
Given("there is some text {}", (String a) -> {

})

Given("there is some number {}", (Double a) -> {

})
```
Depending on the type hint it would transform the matching text (.*) to either a `String` or a `Double`. 

However `CucumberExpression` already creates the whole expression in the constructor. So we'd have to pass the type hints when creating the expression. I can't think of a reason not to do this for java. Is there anything that prevents the other languages?

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/RegularExpression.java
```diff
@@ -20,34 +23,55 @@
...
-                    new Transformer<String>() {
-                        @Override
-                        public String transform(String arg) {
-                            return arg;
-                        }
-                    }
-            );
+
+            if (parameterType != null && !parameterType.isAnonymous()) {
+                // Use hint for validation only
```
mpkorstanje: On second thought I think we should leave the validation out. The language specific step definitions have much more type information available. There is just no way we can get this right for every language here.

aslakhellesoy: I think we should do validation when we *do* have the information (i.e when the tyoe is a class). We’ll be able to make slightly better error messages because we have more context. The drawback is that we won’t always be able to (with generic types) but I think that’s ok.

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/CucumberExpression.java
```diff
@@ -19,16 +20,19 @@
...
         this.source = expression;
+        this.parameterTypeRegistry = parameterTypeRegistry;
 
         expression = processEscapes(expression);
         expression = processOptional(expression);
         expression = processAlternation(expression);
         expression = processParameters(expression, parameterTypeRegistry);
         expression = "^" + expression + "$";
         treeRegexp = new TreeRegexp(expression);
+
```
mpkorstanje: Unrelated to  this PR but we are using `StringBuffer` below. If we don't need threadsafety then `StringBuilder`is a better choice.

cucumber-expressions/java/pom.xml
```diff
@@ -183,8 +191,8 @@
                     <version>3.6.2</version>
                     <configuration>
                         <encoding>UTF-8</encoding>
-                        <source>1.7</source>
-                        <target>1.7</target>
+                        <source>1.8</source>
+                        <target>1.8</target>
```
mpkorstanje: Note to self: Do'nt forget.

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/RegularExpression.java
```diff
@@ -20,34 +23,52 @@
...
-    public List<Argument<?>> match(String text) {
-        List<ParameterType<?>> parameterTypes = new ArrayList<>();
+    public List<Argument<?>> match(String text, Type... typeHints) {
+        final ObjectMapper objectMapper = parameterTypeRegistry.getObjectMapper();
+        final List<ParameterType<?>> parameterTypes = new ArrayList<>();
+        int typeHintIndex = 0;
         for (GroupBuilder groupBuilder : treeRegexp.getGroupBuilder().getChildren()) {
-            String parameterTypeRegexp = groupBuilder.getSource();
+            final String parameterTypeRegexp = groupBuilder.getSource();
+            final Type typeHint = typeHintIndex < typeHints.length ? typeHints[typeHintIndex++] : Object.class;
```
mpkorstanje: This removes so many null checks. :smile:

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/RegularExpression.java
```diff
@@ -20,34 +23,52 @@
...
+                        // TODO: Tell users what to do if they attempt to convert \d+ into Float.
+                        "Capture group with %s transforms to %s, which is incompatible with %s. " +
+                                "Try changing the capture group to something that doesn't match an existing parameter type.",
+                        parameterTypeRegexp,
+                        paramClass.getTypeName(),
+                        hintClass.getTypeName())
+                );
+            }
+
             parameterTypes.add(parameterType);
```
mpkorstanje: @aslakhellesoy I rewrote the way we handled anonymous parameter types. Now the old comments have disappeared.

You mentioned that you wanted to do validation here as we have all the information available here. However that information (`typeHint`, and `parameterType.type`) is also available to the caller of `Expression.match`.

And when I think about the error message I need much more information. For example in the given situation:

```java
@Given("A transaction of (\d+) ([A-Z]{3})")
public void given_a_transaction_of(Amount a, Currency c)
```
The regex of capture group itself tells me very little. 

I'd like this error message to look something like this:

```
Encountered a problem while executing step: "A transaction of 1000 PTE".

Tried to invoke `given_a_transaction_of(Amount a, Currency c)` with these arguments `(Integer, String)`. 

The arguments types do not match up.

Argument values:
 *  Type: Integer
    Value: 1000
    Regex: \d+
    Name: null
 * Type: String
    Value: PTE
    Regex: [A-Z]{3}
    Name: null

You can resolve this problem by:
 * Change the method parameters to match the arguments.
 * Register the method parameters in the parameter type registry  

Link to docs here:
```

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/CucumberExpression.java
mpkorstanje: And we can't do that until Java 9.

mpkorstanje:
Neither ruby nor javascript have a way to telbl cucumber what the parameter type of the method should be so anonymous parameter types can't be implemented there

mpkorstanje:
Implemented the anonymous parameters in ruby and js without the type hinting.

cucumber-expressions/go/default_transformer_test.go
```diff
@@ -0,0 +1,62 @@
...
+	"testing"
+)
+
+func TestConvert(t *testing.T) {
+	t.Run("SimpleDefaultTransformer converts to kind string", func(t *testing.T) {
+		typeOfString := reflect.TypeOf("string")
+		assertTransforms(t, "Barbara Liskov", "Barbara Liskov", typeOfString)
+	})
+
+	t.Run("SimpleDefaultTransformer converts to kind string", func(t *testing.T) {
```
jaysonesmith: What's different/unique about this test versus the one on line 10 that creates the `typeOfString`?

mpkorstanje: Ah I see. The one on line 10 uses `Type` this one uses `Kind`. But I didn't rename the test.

mpkorstanje: We're using both `Type` and `Kind` so `cucumber-go` can use reflection to get `Type` directly from methods via reflect while we can define the transforms for the basic types as `Kind`. It is not possible in go to reference a `Type` directly outside of type assertions.

jaysonesmith:
I haven't looked at the other tests, but table driven tests could help out with repetition? https://github.com/golang/go/wiki/TableDrivenTests

mpkorstanje:
Cheers. I didn't know that. 

It would create a slightly nicer report because each test is separate but it'd still be a single line per test case. So in that respect it wouldn't help much. I'm also not impressed by the added complexity. Tests should be simple to read.

mpkorstanje:
Almost done. Want to integrate it into cucumber-jvm first.

PR #498, opened 2018-10-11
allow to provide custom PatternCompiler

lsuski:
## Summary

fix https://github.com/cucumber/cucumber/issues/494

## Types of changes

## Checklist:

### DIFF
--- a/cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/DefaultPatternCompiler.java
+++ b/cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/DefaultPatternCompiler.java
@@ -1,0 +1,14 @@
+package io.cucumber.cucumberexpressions;
+
+import java.util.regex.Pattern;
+
+/**
+ * Default {@link PatternCompiler} compiling {@link Pattern} with flag {@link Pattern#UNICODE_CHARACTER_CLASS}
+ */
+class DefaultPatternCompiler implements PatternCompiler {
+
+	@Override
+	public Pattern compile(String regexp) {
+		return Pattern.compile(regexp, Pattern.UNICODE_CHARACTER_CLASS);
+	}
+}
--- a/cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/PatternCompilerProvider.java
+++ b/cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/PatternCompilerProvider.java
@@ -1,0 +1,25 @@
+package io.cucumber.cucumberexpressions;
+
+import java.util.ServiceLoader;
+
+final class PatternCompilerProvider {
+	//visible from tests
+	static PatternCompiler service;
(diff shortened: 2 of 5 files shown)


cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/DefaultPatternCompiler.java
```diff
@@ -0,0 +1,14 @@
...
+import java.util.regex.Pattern;
+
+/**
+ * Default {@link PatternCompiler} compiling {@link Pattern} with flag {@link Pattern#UNICODE_CHARACTER_CLASS}
+ */
+class DefaultPatternCompiler implements PatternCompiler {
+
+	@Override
+	public Pattern compile(String regexp) {
+		return Pattern.compile(regexp, Pattern.UNICODE_CHARACTER_CLASS);
```
mpkorstanje: This method should have two arguments. The Android implementation can then filter anything it doesn't support while the the default passes everything on.

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/PatternCompilerProvider.java
```diff
@@ -0,0 +1,25 @@
...
+	//visible from tests
+	static PatternCompiler service;
+
+	private PatternCompilerProvider() {
+	}
+
+	static synchronized PatternCompiler getCompiler() {
+		if (service == null) {
+			ServiceLoader<PatternCompiler> loader = ServiceLoader.load(PatternCompiler.class);
+			for (PatternCompiler patternCompiler : loader) {
```
mpkorstanje: This should error if more then one provider is found.

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/DefaultPatternCompiler.java
mpkorstanje: To be exact it should have the same signature as: `Pattern compile(String regex, int flags)`

Given the difference in the JDK and Android implementation I imagine that these problems will pop up more often in the future.

mpkorstanje:
Looks good! :D

mpkorstanje:
Feel free to merge and make a release. @aslakhellesoy is the release process working?

aslakhellesoy:
LGTM! I'll fix the indentation (4 space indent) and make a release. The release process is currently described in [CONTRIBUTING_NEW](https://github.com/cucumber/cucumber/blob/master/CONTRIBUTING_NEW.md).

The process is still a bit new, so I prefer to do it myself to iron out any kinks.

mpkorstanje:
Awesome!

aslakhellesoy:
The release process still needs some work. I'll start by making a pom archetype because I'm tired of cpoy-pasting XML. Give me a few days.

PR #513, opened 2018-10-25
[Gherkin C] Empty lines in the description break the parser with memory access error

VjacheslavVytjagov:
## Summary
Lines in the description that comply condition GherkinLine_is_empty can break down the parser with memory access error

## Details
GherkinLine_is_empty returns true if the line is empty or contains only whitespaces (or/and tab characters). In the second case the line has real content and nonzero length. Let's see on the get_description_text implementation. At first we pass through description lines to calculate text_length and empty lines (GherkinLine_is_empty) is skipped here. At now let's see on the create_multi_line_text_from_tokens.
The buffer with text_length size for result text is allocated. Then we again pass through description lines and sequentially copy its to the prepared buffer. Empty lines isn't skipped on this step and summary size of the copied text is greater then buffer size. So we wrote outside of the prepared buffer.

I made a consistent calculation of the buffer size (text_length) and filling the buffer 

## Motivation and Context
Modern editors keep indentation of the new line by the previous line indent. So the line with only whitespaces/tabs in the description is common case, especially while editing file

## Types of changes

## Checklist:

### DIFF
--- a/gherkin/c/src/ast_builder.c
+++ b/gherkin/c/src/ast_builder.c
@@ -429,7 +429,7 @@ static const wchar_t* get_description_text(AstNode* ast_node) {
     int current_line = 0;
     while (queue_item) {
         ++current_line;
-        if (!GherkinLine_is_empty(((Token*)queue_item->item)->line)) {
+        if (/*!GherkinLine_is_empty(((Token*)queue_item->item)->line)*/true) { //create_multi_line_text_from_tokens doesn't skip GherkinLine_is_empty lines
             line_count = current_line;
             text_length += wcslen(((Token*)queue_item->item)->matched_text);
         }


aslakhellesoy:
So we can remove the `if` statement here?

VjacheslavVytjagov:
Yes. But I am afraid that someday the code can be modified and get_description_text and create_multi_line_text_from_tokens will be inconsistent again. The C language doesn't forgive errors with memory handling. So I would prefer to keep this comment, may be in other form, as reminder.

cyocum:
I would like to see a test that replicates this.  If the code changes, it should trigger the bug and fail the test, which would ensure that there is no regression in the future.

As for the comment, I think that we should remove the if statement and write a comment either here or at the top of the function.  I do not care too much about either placement but as long as it is nearby.

VjacheslavVytjagov:
Good proposition. I agree that my fix is quick and dirty rather than clean and clear. And I like tests too. So I will try to do this. But this will take a little bit more times.

xtrasimplicity:
Thank you, @VjacheslavVytjagov - once you've added some tests, I think we should merge this. :)

PR #539, opened 2018-11-22
datatable-java: use provided locale to parse BigDecimal in DataTableRegistry

lsuski:
## Summary

Parse `BigDecimal` in DataTableRegistry using locale provided by `TypeRegistryConfigurer`

## Details
Previously BigDecimal in data table cell was parsed by invoking `new BigDecimal(cell)`. This does not work well with non english locale like polish. However double and float are parsed with provided locale so BigDecimal should be treated the same way.

## How Has This Been Tested?

I've added unit tests

## Types of changes

## Checklist:

### DIFF
--- a/datatable/java/datatable/src/main/java/io/cucumber/datatable/NumberParser.java
+++ b/datatable/java/datatable/src/main/java/io/cucumber/datatable/NumberParser.java
@@ -1,5 +1,6 @@
 package io.cucumber.datatable;
 
+import java.math.BigDecimal;
 import java.text.NumberFormat;
 import java.text.ParseException;
 
@@ -16,6 +17,10 @@ class NumberParser {
 
     float parseFloat(String s) {
         return parse(s).floatValue();
+    }
+
+    BigDecimal parseDecimal(String s) {
+        return BigDecimal.valueOf(parse(s).doubleValue());
     }
 
     private Number parse(String s) {
--- a/datatable/java/datatable/src/test/java/io/cucumber/datatable/DataTableTypeRegistryTest.java
+++ b/datatable/java/datatable/src/test/java/io/cucumber/datatable/DataTableTypeRegistryTest.java
@@ -5,15 +5,14 @@ import org.junit.Test;
 import org.junit.rules.ExpectedException;
 
 import java.lang.reflect.Type;
+import java.math.BigDecimal;
+import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 
 import static io.cucumber.datatable.TypeFactory.aListOf;
 import static io.cucumber.datatable.TypeFactory.constructType;
-import static java.util.Arrays.asList;
-import static java.util.Collections.singletonList;
-import static java.util.Objects.requireNonNull;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertSame;
@@ -22,6 +21,7 @@ public class DataTableTypeRegistryTest {
 
     private static final Type LIST_OF_LIST_OF_PLACE = aListOf(aListOf(Place.class));
     private static final Type LIST_OF_PLACE = aListOf(Place.class);
+    private static final Type LIST_OF_LIST_OF_DECIMAL = aListOf(aListOf(BigDecimal.class));
     private static final TableCellByTypeTransformer PLACE_TABLE_CELL_TRANSFORMER = new TableCellByTypeTransformer() {
         @Override
         @SuppressWarnings("unchecked")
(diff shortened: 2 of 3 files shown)


datatable/java/datatable/src/test/java/io/cucumber/datatable/DataTableTypeRegistryTest.java
```diff
@@ -120,4 +120,24 @@ public void returns_entry_data_table_type() {
...
+    }
+
+    @Test
+    public void parse_decimal_with_polish_locale() {
+        DataTableTypeRegistry plRegistry = new DataTableTypeRegistry(Locale.forLanguageTag("pl"));
+        parse_decimal_with_provided_locale(plRegistry, "12,45", 12.45);
+
+    }
+
+    private void parse_decimal_with_provided_locale(DataTableTypeRegistry typeRegistry, String decimalString, double expectedDecimalValue) {
```
aslakhellesoy: This is an assertion method. Could you rename it to describe what it asserts, preferrably prefixing with `assert` or `expect`?

datatable/java/datatable/src/main/java/io/cucumber/datatable/NumberParser.java
```diff
@@ -18,6 +19,10 @@ float parseFloat(String s) {
         return parse(s).floatValue();
     }
 
+    BigDecimal parseDecimal(String s) {
+        return BigDecimal.valueOf(parse(s).doubleValue());
```
mpkorstanje: This will cause a loss of persission.

datatable/java/datatable/src/test/java/io/cucumber/datatable/DataTableTypeRegistryTest.java
```diff
@@ -120,4 +120,24 @@ public void returns_entry_data_table_type() {
...
+        expectDecimalValueForDecimalStringInLocale(Locale.ENGLISH, 12.45, "12.45");
+    }
+
+    @Test
+    public void parse_decimal_with_polish_locale() {
+        expectDecimalValueForDecimalStringInLocale(Locale.forLanguageTag("pl"), 12.45, "12,45");
+
+    }
+
+    private void expectDecimalValueForDecimalStringInLocale(Locale locale, double expectedDecimalValue, String decimalString) {
```
mpkorstanje: The use of custom assert methods usually indicates that the test is not properly structured. In this instance test is testing a property of the `NumberParser` but does so through the `DataTableTypeRegistry` and needs a fairly complicated setup to do so.

Creating a `NumberParserTest` and moving this test would solve that problem.

lsuski: Yes but it would not then check if DataTableRegistry uses NumberParser properly. We would need have another test in DataTableTypeRegistryTest which checks if proper method from NumberParser is invoked.

datatable/java/datatable/src/test/java/io/cucumber/datatable/DataTableTypeRegistryTest.java
```diff
@@ -120,4 +120,24 @@ public void returns_entry_data_table_type() {
...
+        DataTableTypeRegistry typeRegistry = new DataTableTypeRegistry(locale);
+        //noinspection unchecked
+        List<List<BigDecimal>> transform = (List<List<BigDecimal>>) typeRegistry.lookupTableTypeByType(LIST_OF_LIST_OF_DECIMAL).transform(Collections.singletonList(Collections.singletonList(decimalString)));
+
+        assertEquals(1,transform.size());
+        assertEquals(0,transform.get(0).get(0).compareTo(BigDecimal.valueOf(expectedDecimalValue)));
```
mpkorstanje: This again causes a loss of precision. Rather then creating the expected `BigDecimal` from a `double` a test should always create `BigDecimals` directly from string.

Note that this error is extra sneaky because it is hiding in all the setup code.

datatable/java/datatable/src/test/java/io/cucumber/datatable/DataTableTypeRegistryTest.java
mpkorstanje: Fair enough. I still think these test can be much more concise though. To the point of not needing the the custom assert.

mpkorstanje: @lsuski I've pushed a fix that will make the tests more concise.

mpkorstanje:
@lsuski feel free to make a release. You can find the instructions in  [CONTRIBUTING_NEW.md](https://github.com/cucumber/cucumber/blob/master/CONTRIBUTING_NEW.md).

mpkorstanje:
It's failing on java 1w when building in the repo. Curious.

lsuski:
I was a little confused by those builds, cucumber build passed so I pushed post release commit when it was green. Then I received notification about failing build of datatable-java and saw that release v1.1.8 also failed. I had to leave so couldn't dig deeper

lsuski:
Seems that Java 11 behaves differently with decimals

mpkorstanje:
I've fixed it.

mpkorstanje:
Something is wrong with Travis. Lots of jobs queued but they're being processed slowly. I reckon we'll have a release in a few hours though.

PR #552, opened 2018-12-31
gherkin-c: Extend to accept current Gherkin spec

Corristo:
## Summary
This implements the new `Rule` keyword and makes `Scenario Outline` and `Scenario` synonymous.

## Details
- Ran `berp` to create up-to-date `rule_type.h` and `parser.c` files
- Adjusted the `ast_builder` to use new enumerator names (e.g. `Rule_Scenario_Definition` was renamed to `Rule_ScenarioDefinition`)
- Removed the old `Scenario` struct, renamed the `ScenarioOutline` struct to `Scenario`
- Moved code to clean up `ScenarioDefinition`s from `feature.c` into its own `ScenarioDefintions_delete` function declared in the `scenario_definitions.h` header.
- Implement handling of the `Rule` keyword/struct in the `ast_builder`, `ast_printer`, `compiler` and `token_formatter_builder`.
- updated `dialect.c` so that it incorporates the new keyword (and general updates to existing keywords)

## Motivation and Context
With #353 the `Scenario Outline` keyword has been deprecated and #353 introduced the `Rule` keyword. Both of these features haven't been implemented in the gherkin-c library, however there are tests for these already in the repository, causing the build to fail (#533).

## How Has This Been Tested?
The previously failing `rule.feature` and `scenario_outline.feature` now pass, the other tests still pass.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue).
- [x] New feature (non-breaking change which adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).

## Checklist:
- [ ] I've added tests for my code.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #554, opened 2018-12-31
gherkin-c: Link against `libm` if required

Corristo:
## Summary
Duing the configuration of the build, check whether linking against `libm` is required to use `log10`, and if so, link against it. 

## Details
Using the `CheckSymbolExists` CMake module we now first check whether `log10` can be used without any additional library, if not we try again with `libm`, and if it still fails we give up and abort the build.

## Motivation and Context
The `glibc` C library used by most linux distributions requires to link against `libm` to use `log10`. gherkin-c uses `log10` in `error_list.c` but so far didn't link against `libm`, causing the build to fail on my Arch linux. 

## How Has This Been Tested?
By observing that after this change I can build gherkin-c on my linux box with both GCC and clang.

## Types of changes

## Checklist:

### DIFF
--- a/gherkin/c/CMakeLists.txt
+++ b/gherkin/c/CMakeLists.txt
@@ -68,6 +68,23 @@ add_library(gherkin ${GHERKIN_SRS})
 target_include_directories(gherkin PUBLIC
     "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include;${CMAKE_CURRENT_SOURCE_DIR}/src>")
 
+include(CheckSymbolExists)
+include(CMakePushCheckState)
+
+cmake_push_check_state()
+set(CMAKE_REQUIRED_QUIET ON)
+check_symbol_exists(log10 math.h RES)
+if(NOT RES)
+  list(APPEND CMAKE_REQUIRED_LIBRARIES m)
+  check_symbol_exists(log10 math.h RES2)
+  if(RES2)
+    target_link_libraries(gherkin PRIVATE m)
+  else()
+      message(FATAL_ERROR "Don't know what library to link against to use log10.")
+  endif()
+endif()
+cmake_pop_check_state()
+
 add_executable(gherkinexe ${GHERKIN_CLI})
 target_link_libraries(gherkinexe gherkin)
 target_include_directories(gherkinexe PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include ${CMAKE_CURRENT_SOURCE_DIR}/src)


cyocum:
My first caveat is that I know _nothing_ about CMake and I have never used it so my comments may be completely spurious.  I was able to run CMake using the instructions in the README (I generally just use make as I am much more comfortable with it).   The first command that I ran was `cmake ..` and I noticed

```
...
-- Looking for log10
-- Looking for log10 - not found
-- Looking for log10
-- Looking for log10 - found
...
```
Is it looking for libm twice?

After that, I ran the build and while it seems to have complied and ran the tests using `cmake --build . --target install`, I noticed:

```
gmake[2]: *** [CMakeFiles/gherkinexe.dir/build.make:108: gherkinexe] Error 1
gmake[2]: *** Deleting file 'gherkinexe'
gmake[1]: *** [CMakeFiles/Makefile2:73: CMakeFiles/gherkinexe.dir/all] Error 2
```

Is there a way to make it not do that so I can investigate the resulting errors?

I also noticed that the libgherkin generated by CMake has a different sha hashsum than the one created by Make.  I am assuming here that they use different compilation flags.

More worrisome is that the tests are not passing on Make either and the output from Make is that there are problems with the DocStringSeparator.

Anyway, thank you very much for the pull request and I hope my comments were not off point.

Corristo:
> Is it looking for libm twice?
No, the first time it tries to use `log10` without using an additional library, and since that fails (it says "not found")  it tries again this time linking against `libm`. I'll see if I can improve the output there, not sure whether that is supported by `CheckSymbolExists`.

> Is there a way to make it not do that so I can investigate the resulting errors?
I'll create another pull request for that. I already have a fix since I also had to change that behaviour in order to test my other changes.

> More worrisome is that the tests are not passing on Make either and the output from Make is that there are problems with the DocStringSeparator.
I'm working on making the tests pass again, see e.g. #552, but I hadn't noticed that `make` runs more thorough tests than CMake, thanks for bringing that to my attention.

Corristo:
So I just checked whether it is possible to change the message of `CheckSymbolExists` and unfortunately this isn't possible. However, from CMake 3.1 onwards (currently gherkin-c requires only CMake 3.0) it is possible to disable the output entirely so that there is only the fatal error message I provided when `log10` can not be found even when linking against `libm`.

Since disabling the output is possible in a backward compatible manner I've decided to disable the output. Users that still use CMake 3.0 will see
>-- Looking for log10
-- Looking for log10 - not found
-- Looking for log10
-- Looking for log10 - found

but everyone with a newer CMake version will only see an error message when `log10` doesn't work even with `libm`.

cyocum:
I think we can merge this now but I would like to get feedback from @aslakhellesoy and @brasmusson before we do anything. I have not looked at the CI build failures but they seem to be systemic and I have not had time to look into them.

luke-hill:
Again this looks the next smallest, so I'd do another compress/rebase and push. (Essentially there were lots of build issues in going to v4.rc1 (Which is current master), so rebasing is likely to fix most of your outstanding branches.

Then perhaps ping bjorn (I won't tag him as he might be busy)

Corristo:
@luke-hill I've rebased this as well as #555 onto the current master. Unfortunately the fixes for v4.rc1 you mentioned don't seem to have changed anything for `gherkin-c`, i.e. #552 and #556 are still required to make all the tests pass again.

@brasmusson Could you please review this?

aslakhellesoy:
> I hadn't noticed that make runs more thorough tests than CMake

It is problematic to have two build processes. Why do we need CMake?

Corristo:
@aslakhellesoy CMake has become the de-facto standard build system for open-source C++ projects (I'm not sure about the C ecosystem), and it is very easy to integrate 3rd-party libraries that are built with CMake into other CMake projects. Thus having a CMake project makes it easier to consume gherkin-c.

Moreover, CMake is cross-platform, while the Makefile of this project is a Unix-Makefile, i.e. it cannot be used directly on windows but requires e.g. msys2 or mingw. CMake on the other hand can generate e.g. Visual Studio projects on Windows and is even supported natively in Visual Studio 2017 and 2019.

mpkorstanje:
Fixed by https://github.com/cucumber/cucumber/pull/940

PR #555, opened 2018-12-31
gherkin-c: Use CTest to run the tests in CMake builds

Corristo:
## Summary
The tests are now run through CTest instead of as a custom build step of `gherkinexe`.
In addition, tests run on CMake builds now also compare the expected results to the actual results just like the Makefile does. 

## Details
- Enabled CTest in the `CMakeLists.txt`
- Added a `TestRunner.cmake` script that runs the tests
- Updated the README

## Motivation and Context
Currently the gherkin-c tests fail due to missing implementations for the `Rule` keyword and for `Scenario` being synonymous with the now deprecated `Scenario Outline`. However, since the tests currently run as a post-build step on the `gherkinexe` target, any test failure causes `gherkinexe` to be deleted again. This is very unfortunate as in this state it is not possible to use `gherkin-c` as a subproject in a larger project that doesn't require these new features yet. Moreover, debugging test failures also is not possible because the executable to debug is not being built.

This also has the added benefit that CTest runs all testcases instead of stopping after the first failed one.

## How Has This Been Tested?
Tested locally with CMake 3.13.1 by running
`cmake -Bbuild -H. && cmake --build build`
and noticing that it correctly builds `gherkinexe` even though `cmake --build build --target test` reports test failures.

## Types of changes

## Checklist:

### DIFF
--- a/gherkin/c/CMakeLists.txt
+++ b/gherkin/c/CMakeLists.txt
@@ -89,7 +89,22 @@ foreach(ENTITY ${GOOD_FEATURE_FILES})
     get_filename_component(TEST_NAME ${ENTITY} NAME_WE)
     add_test(
         NAME ${TEST_NAME}
-        COMMAND $<TARGET_FILE:gherkinexe> ${ENTITY}
+        COMMAND ${CMAKE_COMMAND}
+                -DGHERKINEXE=$<TARGET_FILE:gherkinexe>
+                -DGHERKIN_GENERATE_TOKENS=$<TARGET_FILE:gherkin_generate_tokens>
+                -DFEATUREFILE=${ENTITY}
+                -P ${CMAKE_CURRENT_SOURCE_DIR}/cmake/TestRunner.cmake
+    )
+endforeach()
+foreach(ENTITY ${BAD_FEATURE_FILES})
+    get_filename_component(TEST_NAME ${ENTITY} NAME_WE)
+    add_test(
+        NAME ${TEST_NAME}
+        COMMAND ${CMAKE_COMMAND}
+                -DGHERKINEXE=$<TARGET_FILE:gherkinexe>
+                -DGHERKIN_GENERATE_TOKENS=$<TARGET_FILE:gherkin_generate_tokens>
+                -DFEATUREFILE=${ENTITY}
+                -P ${CMAKE_CURRENT_SOURCE_DIR}/cmake/TestRunner.cmake
     )
 endforeach()
 
--- a/gherkin/c/CMakeLists.txt
+++ b/gherkin/c/CMakeLists.txt
@@ -64,6 +64,8 @@ LIST(APPEND GHERKIN_CLI
         src/gherkin_cli.c
         )
 
+include(CTest)
+enable_testing()
 
 add_library(gherkin ${GHERKIN_SRS})
 target_include_directories(gherkin PUBLIC
@@ -82,16 +84,15 @@ file(GLOB GOOD_FEATURE_FILES
 file(GLOB BAD_FEATURE_FILES
         ${CMAKE_CURRENT_SOURCE_DIR}/testdata/bad/*.feature
         )
-add_custom_target(invo DEPENDS invo)
-add_dependencies(invo gherkinexe)
-FOREACH(ENTITY ${GOOD_FEATURE_FILES})
-    add_custom_command(
-            TARGET gherkinexe
-            POST_BUILD
-            COMMAND gherkinexe  ${ENTITY}
-            DEPENDS invo
+
+foreach(ENTITY ${GOOD_FEATURE_FILES})
+    get_filename_component(TEST_NAME ${ENTITY} NAME_WE)
+    add_test(
+        NAME ${TEST_NAME}
+        COMMAND $<TARGET_FILE:gherkinexe> ${ENTITY}
     )
-ENDFOREACH()
+endforeach()
+
 ############ Installation section ############
 set(include_install_dir "include")
 set(lib_install_dir "lib/")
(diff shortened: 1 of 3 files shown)


Corristo:
Is there something I can do to help get this merged?

mpkorstanje:
@brasmusson can I ask you to review this one too?

aslakhellesoy:
@Corristo gherkin-c has been updated by @brasmusson to support the new `Rule` keyword. Using CTest seems ok to me, but I do not think we should allow the build to pass when there are test failures.

Can you update this PR please? And also merge master into your PR branch?

Thanks!

Corristo:
@aslakhellesoy I'm not sure it is supported by CMake to fail the build step when ctest fails, I'd have to look into that. But before I do that I'd like to understand why you'd like the build to fail and if we could achive the same goal with something more in line with the CMake workflow.

The typical CMake workflow is to use `cmake --build <build_directory>` to build everything, then have a separate `cmake --build <build_directory> --target test` step to run the tests and then finally `cmake --build <build_directory> --target install` to install everything.

If you're concerned that users might not be aware of this and assume a succeeding build step means they're good to go then I could update the README accordingly. If you only want the CI to fail when the tests fail (e.g. when testing pull-requests) I could also change the travis file to also run the tests in the separate step.

I'd really try to avoid having the build itself fail when the tests don't pass as that causes CMake to also delete the artifacts of the build step, which means in order to debug the issue you first have to disable running the tests in the build-system. 

Given that test failures in gherkin-c can happen basically at any time simply because upstream pushed new feature-files (e.g. because of feature additions of a new gherkin version) or new expected output, failing tests typically do not mean that gherkin-c is broken, but only that it doesn't support the newest version yet. Users then might still want to use that version of gherkin-c then if they don't require the latest features, and having them to fiddle with the build-system just to get the binaries seems like a big hurdle to use e.g. cucumber.ml which requires its users to build gherkin-c first.

aslakhellesoy:
> But before I do that I'd like to understand why you'd like the build to fail and if we could achive the same goal with something more in line with the CMake workflow.

What is the purpose of a test suite that doesn't fail when there are failures?

I use tests to tell me when something is wrong, so I can fix it. If the build doesn't fail when there are test failures then how are we supposed to even know about bugs? This applies to both CI builds and local developer builds.

> Given that test failures in gherkin-c can happen basically at any time simply because upstream pushed new feature-files (e.g. because of feature additions of a new gherkin version) or new expected output, failing tests typically do not mean that gherkin-c is broken, but only that it doesn't support the newest version yet.

I disagree. If someone adds a new `gherkin/testdata/good/*.feature` file and that causes a test failure in any of the gherkin implementations (such as gherkin-c), then that does indeed mean that the gherkin parser is broken and should be fixed. It doesn't matter if the feature file illustrates a new requirement for the parser or highlights an existing bug. The parser is broken and must be fixed.

Corristo:
> What is the purpose of a test suite that doesn't fail when there are failures?
The test suite would still fail if you choose to run it, but I'd not make it mandatory to run it just to get a gherkin-c binary.

Corristo:
@aslakhellesoy I just realized our disagreement comes from the fact that we were most likely talking about very different kind of "build"s. Am I correct in assuming that what you meant with "build" is a CI/CD build (e.g. the cucumber travis build) that consists of multiple steps, starting from checking out a specific version of the code, generating the artifacts, testing and optionally packaging or deploying the artifacts? 

Because if you were indeed talking about the cucumber travis build then I completely agree with all you comments. Of course the travis build should fail when the tests fail and in that scenario missing features of gherkin-c should also be a build failure - and this PR does not change that.

But that is not what I meant with "build". This pull request and all my comments were referring to the single step of the CI/CD pipeline that creates the artifacts. And there in my opinion it really does not make sense to make running the tests part of the build step. For example, if I need the gherkin-c binaries for a Raspberry PI I don't want to compile it on the slow Raspberry PI itself, but I'd cross-compile it from a more powerful x86 machine. But then I can't even run the tests on the machine that builds gherkin-c, unless I use an emulator which itself might be the reason the tests fail even though the generated binaries are perfectly fine.

aslakhellesoy:
@Corristo ahaaaaaaa! Thanks for spotting that misunderstanding. What a wonderful example of the importance of conversations. This would have been so much easier face to face.

Anyway - I agree with you. It should be possible to *build binaries* without having to run the tests first. Absolutely - please make it that way.

I'm still a bit puzzled about CMake though. We have Make, and it works fine. What does CMake add? It definitely complicates things a bit to have to maintain two.

Corristo:
@aslakhellesoy The PR already makes it so that one can create the binaries without running the tests - that was the entire point.

Regarding Make vs CMake:

The Makefile works fine on Linux, but as it is a Unix Makefile it cannot be used on Windows directly (nmake/jom Makefiles have their own syntax) but requires Windows users to use install mingw or msys2 in order to build gherkin-c.

CMake on the other hand is a cross-platform build system generator that can create Visual Studio solutions that can be used directly, and Visual Studio 2017 and 2019 even have builtin CMake support so you don't even need that extra step.

Another advantage of CMake is that it has become the de-facto standard build system for open source C++ projects, even Qt (which currently uses its own build system qmake and has developed another build system called qbs before) is switching to CMake. One of the reasons for this is that CMake makes it very easy to integrate other CMake projects as dependencies of your own project. Due to this proliferation, many IDEs have builtin CMake support, which allows to easily run their code analysis tools.

Moreover, most of the available C and C++ tooling nowadays is libclang or libtooling based, both of these require a compile_commands.json file in order to understand the project they're analysing. CMake has support to generate that file as part of the configuration step, with Make you need an additional tool called `bear` which creates this file as you're building the code. In particular, a failing build thus yields an incomplete compile_commands.json file, and since bear can't incrementally update that file you need to first fix compilation errors and then perform a clean build again before you can run the tools.

All in all, CMake is much more comfortable to work with than the plain Makefile.

PR #556, opened 2019-01-03
gherkin-c: Fix test failures due to changed JSON serialization format

Corristo:
## Summary
Fix tests failing due to changes to the JSON output format of pickles, ast and the source. 

## Details
- Updated the ast printer
- Updated the pickle printer
- Updated the source printer
- Updated the error message format
- QuickFix: Use the same name for `Scenario` and `Scenario Outline` when printing tokens
- Fixed a bug where the ordering of location data in pickles of scenarios with examples was wrong

## Motivation and Context
This is one of the changes required to get `gherkin-c` to build without failures again. In my opinion it makes sense to fix the tests in two sets of changes, this one and #552. By merging this PR first it will be easier to verify that #552 (which touches a lot of code) does indeed not cause regressions and makes all tests pass again.

## How Has This Been Tested?
Using the changes in #555 it is possible to run all tests instead of stopping after the first failure as running `make` does. When doing so one can see that there are now only three tests of the "good" test data set (`incomplete_scenario_outline`, `rule`, `scenario_outline`) and two of the tests in the "bad" test data set (`multiple_parser_errors`, `unexpected_eof`) which fail.

For both "bad" test cases as well as the `rule` and `scenario_outline` test cases failure is expected until #552 is merged. `incomplete_scenario_outline` however shows a real bug that incidentally also got fixed in #552. That bug has been around since at least v5.1.0, yet it wasn't caught because the reference pickles were also incorrect.

## Types of changes

## Checklist:

### DIFF
--- a/gherkin/c/include/doc_string.h
+++ b/gherkin/c/include/doc_string.h
@@ -16,9 +16,10 @@ typedef struct DocString {
     Location location;
     wchar_t* content_type;
     const wchar_t* content;
+    const wchar_t* delimiter;
 } DocString;
 
-const DocString* DocString_new(Location location, const wchar_t* content_type, const wchar_t* content);
+const DocString* DocString_new(Location location, const wchar_t* content_type, const wchar_t* content, const wchar_t* delimiter);
 
 void DocString_delete(const DocString* doc_string);
 
--- a/gherkin/c/src/ast_builder.c
+++ b/gherkin/c/src/ast_builder.c
@@ -166,7 +166,7 @@ static void* transform_node(AstNode* ast_node, AstBuilder* ast_builder) {
     }
     case Rule_DocString: {
         token = AstNode_get_token(ast_node, Token_DocStringSeparator);
-        const DocString* doc_string = DocString_new(token->location, token->matched_text, get_doc_string_text(ast_node));
+        const DocString* doc_string = DocString_new(token->location, token->matched_text, get_doc_string_text(ast_node), token->matched_keyword);
         Token_delete(token);
         AstNode_delete(ast_node);
         return (void*)doc_string;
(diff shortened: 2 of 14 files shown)


mpkorstanje:
@brasmusson can I ask you to review this?

aslakhellesoy:
Hi @Corristo,

Thanks for this PR and sorry for not responding earlier. We don't currently have a maintainer of the C implementation - it used to be @brasmusson, but I'm not sure we can count on him as much as we used to.

We're also looking to reduce the number of Gherkin implementations [we have to maintain](https://docs.google.com/document/d/12Y3Ut-SVSGdw9L_aW6NGQY9JaCcn3z0alY4d7kzOndU/edit#) and we're aiming for a consolidation approach based on protocol buffers.

What are you using `gherkin-c` for?

Corristo:
@aslakhellesoy 
I'm trying to improve cucumber-cpp by providing a means to have the test-executable itself read the feature files (or maybe pre-compiling the feature files into test executables, not settled on the exact details yet) instead of using the ruby cucumber server to run the tests.

There are several reasons for that. First of all, it was very painful for me as a C++ developer with no ruby experience to get cucumber-cpp working. In particular, getting it to work in a Windows docker container took me almost 4 days, mostly due to issues with ruby itself. And secondly, the last version of cucumber-ruby I was able to get cucumber-cpp working with is 2.0.0, so of course you can't use recent gherkin features with cucumber-cpp right now either.

In order to allow easier adoption of cucumber-cpp for new teams and developers I'd like it to become self-contained. In particular, it shouldn't have mandatory dependencies on other executables like gherkin-go or a cucumber server implementation which could be incompatible. With gherkin-c I can just link the library in statically and don't even need to deserialize json or protobuf generated by e.g. gherkin-go, but can use the data structures provided by gherkin-c directly.

PR #559, opened 2019-01-16
Correct Azerbijani scenario translation

phil-davis:
## Summary

Fix the translation of ``scenario`` in Azerbaijani to use the singular for ``example``

## Details

The translations of ``scenario`` has extra entries added so that the word "Example" could be used as well as "Scenario". There are also existing translations of ``examples`` which make an examples table within an scenario/example.

The Azerbaijani translation of ``scenario`` included "Nümunələr" which is the plural, and the same as the translation of ``examples``. Fix it to use the singular.

## Motivation and Context

If the translations of ``examples`` and ``scenarios`` translate to the same thing then there is trouble. e.g. when parsing a feature and seeing (in Azerbaijani) the keyword "Nümunələr", the parser is not going to know if it has reached an ``examples``` or ``scenario`` section.

This was noticed because of a warning in unit tests in ``Bhat/Gherkin`` see issue https://github.com/Behat/Gherkin/issues/145

## How Has This Been Tested?

See ``Behat/Gherkin`` PR https://github.com/Behat/Gherkin/pull/146 which demonstrates that the equivalent change there fixes the unit test warning.

## Types of changes

## Checklist:

### DIFF
--- a/gherkin/gherkin-languages.json
+++ b/gherkin/gherkin-languages.json
@@ -262,7 +262,7 @@
       "Rule"
     ],
     "scenario": [
-      "Nümunələr",
+      "Nümunə",
       "Ssenari"
     ],
     "scenarioOutline": [


ciaranmcnulty:
If this was causing an issue, how come the tests didn't pick it up? Shouldn't the parsers have been breaking?

phil-davis:
I guess nobody wrote a feature in Azerbaijani using the duplicate keyword?
I haven't actually tried writing scenarios and example tables in Azerbaijani! I just see that in all other languages the translation of ``scenario`` is to the singular form of "Example". And I theorize that the (I presume accidental) duplication in the translation here is going to cause ambiguity in some parsing situations.

stof:
@ciaranmcnulty I guess cucumber does not have the equivalent test than what we have in Behat/Gherkin: generating features using all possible keywords and then trying to parse them, for each language.

PR #561, opened 2019-01-24
Fix ruby issue 1337: Brackets/Integer/CukeExp compatibility

luke-hill:
## Summary

When using Integers captured using the {int} cucumber expression, issues crop up when trying to sanitize the brackets which are being read as an optional parameter.

The issue is occurring due to interoperability between brackets being partially escapable (To represent these are brackets and not a capture group), and cucumber expressions (Which have a different meaning for brackets)

## Details

I've allowed brackets to be used in a very specific context (As notified by the constant definition), when being used inside a Cucumber Expression Generator method (Which would previously crash erroneously)

## Motivation and Context

Fixes Ruby issue #1337 https://github.com/cucumber/cucumber-ruby/issues/1337

## How Has This Been Tested?

This code is likely to need porting and is only meant as a first draft. It's probably going to need some stern eyes to work out if this is desired.

### DIFF
--- a/cucumber-expressions/ruby/lib/cucumber/cucumber_expressions/cucumber_expression.rb
+++ b/cucumber-expressions/ruby/lib/cucumber/cucumber_expressions/cucumber_expression.rb
@@ -10,6 +10,7 @@ module Cucumber
       PARAMETER_REGEXP = /(\\\\)?{([^}]*)}/
       OPTIONAL_REGEXP = /(\\\\)?\(([^)]+)\)/
       ALTERNATIVE_NON_WHITESPACE_TEXT_REGEXP = /([^\s^\/]+)((\/[^\s^\/]+)+)/
+      BRACKETS_WANTED_REGEXP = /\\\\\({.+}\)/
       DOUBLE_ESCAPE = '\\\\'
       PARAMETER_TYPES_CANNOT_BE_ALTERNATIVE = 'Parameter types cannot be alternative: '
       PARAMETER_TYPES_CANNOT_BE_OPTIONAL = 'Parameter types cannot be optional: '
@@ -19,7 +20,7 @@ module Cucumber
       def initialize(expression, parameter_type_registry)
         @source = expression
         @parameter_types = []
-        
+
         expression = process_escapes(expression)
         expression = process_optional(expression)
         expression = process_alternation(expression)
@@ -51,7 +52,7 @@ module Cucumber
         # Create non-capturing, optional capture groups from parenthesis
         expression.gsub(OPTIONAL_REGEXP) do
           g2 = $2
-          check_no_parameter_type(g2, PARAMETER_TYPES_CANNOT_BE_OPTIONAL)
+          check_no_parameter_type(g2, PARAMETER_TYPES_CANNOT_BE_OPTIONAL) unless brackets_wanted?(expression)
           # look for double-escaped parentheses
           $1 == DOUBLE_ESCAPE ? "\\(#{g2}\\)" : "(?:#{g2})?"
         end
@@ -100,6 +101,10 @@ module Cucumber
         if PARAMETER_REGEXP =~ s
           raise CucumberExpressionError.new("#{message}#{source}")
         end
+      end
+
+      def brackets_wanted?(expression)
+        expression =~ BRACKETS_WANTED_REGEXP
       end
     end
   end
--- a/cucumber-expressions/ruby/lib/cucumber/cucumber_expressions/cucumber_expression.rb
+++ b/cucumber-expressions/ruby/lib/cucumber/cucumber_expressions/cucumber_expression.rb
@@ -10,7 +10,7 @@ module Cucumber
       PARAMETER_REGEXP = /(\\\\)?{([^}]*)}/
       OPTIONAL_REGEXP = /(\\\\)?\(([^)]+)\)/
       ALTERNATIVE_NON_WHITESPACE_TEXT_REGEXP = /([^\s^\/]+)((\/[^\s^\/]+)+)/
-      BRACKETS_WANTED_REGEXP = /\\\\\({.+}\)/
+      BRACKETS_WANTED_REGEXP = /\\\\\(.*{.+}.*\)/
       DOUBLE_ESCAPE = '\\\\'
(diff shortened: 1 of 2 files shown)


olleolleolle:
@aslakhellesoy This is about Cucumber Expressions, and it's a usability improvement, do you think it's "what users expect"?

luke-hill:
I'll add some comments to this to try illustrate the specificity of the problem it solves @olleolleolle

olleolleolle:
Those are *chef's kiss* perfect "why comments".

luke-hill:
ping @mattwynne / @mpkorstanje - Could you take a look perhaps and see if this needs porting to other language (My hunch says yes) I've no idea who else to tag

mpkorstanje:
@jaysonesmith would you be able to do do Go and Javascript? I can do Java and also Go if need be but I'm somewhat preoccupied.

mpkorstanje:
@luke-hill Cheers. Tests would be useful. Would make implementing the other languages a bit faster.

jaysonesmith:
@mpkorstanje I'm a bit busy, too, but I can certainly give them a shot. :)

mpkorstanje:
@charlierudolph are you still available to work on Go?

jaysonesmith:
@mpkorstanje I can try to get to them tonight

danascheider:
Is this functionality that we should have tests for? It seems to me like we probably should.

jaysonesmith:
@danascheider Yeah, I feel we absolutely should especially since it wasn't immediately clear to me about the functionality and as @mpkorstanje said, they could help make other languages easier. :D

jaysonesmith:
We should document this at the end of the [expressions section](https://docs.cucumber.io/cucumber/cucumber-expressions/). Anyone want to take a crack at it? I'm happy to review if so.

luke-hill:
This is RtM then @mpkorstanje / @jaysonesmith

mpkorstanje:
Lets merge once we have java and javascript done too. The release process releases them all in one go.

luke-hill:
So the code snippet for ruby you pasted is the current `master` I've altered this so it isn't always ran.

But I can add a test for the additional behaviour no problem. And see what happens. Give me a day or so I'll tack it on

mpkorstanje:
Ah sorry. I should have clarified.

1. The snippet from master for java passes both tests.
2. The snippet from master for ruby fails both tests.
3. I think that the snippet from ruby from this PR will not pass the second test.
4. I think the solution for master used by java is correct.

So everything together, the behavior was correctly implemented in java but not in ruby and the current PR doesn't quite fix it either.

luke-hill:
Right - Take 2.

@mpkorstanje @jaysonesmith @charlierudolph @olleolleolle - I "think" I've cracked it now. And I tried an additional trickier test which it passes plus some other local debugging. WDYT?

cucumber-expressions/ruby/lib/cucumber/cucumber_expressions/cucumber_expression.rb
```diff
@@ -51,7 +52,16 @@ def process_optional(expression)
...
+          # When using Parameter Types, the () characters are used to represent an optional
+          # item such as (a ) which would be equivalent to (?:a )? in regex
+          #
+          # You cannot have optional Parameter Types i.e. ({int}) as this causes
+          # problems during the conversion phase to regex. So we check for that here
+          #
+          # One exclusion to this rule is if you actually want the brackets i.e. you
+          # want to capture (3) then we still permit this as an individual rule
+          # See: https://github.com/cucumber/cucumber-ruby/issues/1337 for more info
+          check_no_parameter_type(g2, PARAMETER_TYPES_CANNOT_BE_OPTIONAL) unless brackets_wanted?(expression)
```
mpkorstanje: Not sure if this is the right way to write this in ruby but why write `unless $1 == DOUBLE_ESCAPE`? I can't come up with a test case that would make this fail.

luke-hill: Is that a strong enough test? And I've done something similar just by using another private method.

Don't we also need to check there is a parameter being captured in the middle? Or is that kind of redundant?

mpkorstanje: > Don't we also need to check there is a parameter being captured in the middle? Or is that kind of redundant?

I think so. We are escaping the brackets. So we don't have to check if there is a parameter type in the brackets because as far as Cucumber is concerned there are none (they are escaped).

So for example test case should also pass `some \(prefix {int} infix {int} suffix) thing` and this too `some \(prefix {int} infix {int} infix {int} infix {int} infix {int} suffix) thing`.

mpkorstanje: Maybe it'd be less confusing to write:
```
if $1 == DOUBLE_ESCAPE:
   return "\\(#{g2}\\)"
else 
  check_no_parameter_type(g2, PARAMETER_TYPES_CANNOT_BE_OPTIONAL)
  return "(?:#{g2})?"

```

luke-hill: :point_up: looks good. Let me take another attack later; I'll also write some more specs.

mpkorstanje:
> Lets merge once we have java and javascript done too. The release process releases them all in one go.

I've added the fixes for Java (just the test) and Javascript to @jaysonesmith PR.

https://github.com/cucumber/cucumber/pull/572

jaysonesmith:
Nice work @mpkorstanje! What's left for this?

mpkorstanje:
I was waiting for luke to finish. But as it's just a two line change I've merged this branch into #572 and applied the change myself.

@luke-hill please do feel free to create branches in this repo in the future. You're already a member and it makes everything easier.

mpkorstanje:
Build green! We're done. I've got a few more fixes piled up for cucumber expressions. I hope I've got some time tomorrow to run a release.

jaysonesmith:
Yeeeessss. Thank you @mpkorstanje !

PR #578, opened 2019-03-12
cucumber-messages: Add stricter gem restrictions for google-protobuf

luke-hill:
## Summary
As the comments elude to; this gem is a key runtime dependency. And on Ruby 2.6 this only works when using the latest version

## Details
Stricter gemspec restrictions as per the different ruby requirements

## Motivation and Context

For Running on JRuby and a couple of other environments, a weaker requirement is required (And not the latest), due to some issues with those environments.

Once the team maintaining this has done some additional work, we can probably relax this requirement. But for now this makes cucumber build properly again X-Ruby Versions

## Checklist:

### DIFF
--- a/cucumber-messages/ruby/Gemfile
+++ b/cucumber-messages/ruby/Gemfile
@@ -1,6 +1,3 @@
 # frozen_string_literal: true
 source "https://rubygems.org"
 gemspec
-
-# Pick a known-good version when running builds for cucumber-messages on JRuby.
-gem 'google-protobuf', '~> 3.2.0.2', platform: :jruby
--- a/cucumber-messages/ruby/Gemfile
+++ b/cucumber-messages/ruby/Gemfile
@@ -1,3 +1,7 @@
 # frozen_string_literal: true
 source "https://rubygems.org"
+
+# Pick a known-good version when running builds for cucumber-messages on JRuby.
+gem 'google-protobuf', '~> 3.2.0.2', platform: :jruby
+
 gemspec
--- a/cucumber-messages/ruby/cucumber-messages.gemspec
+++ b/cucumber-messages/ruby/cucumber-messages.gemspec
@@ -19,8 +19,8 @@ Gem::Specification.new do |s|
                     'source_code_uri'   => 'https://github.com/cucumber/cucumber/blob/master/cucumber-messages/ruby',
                   }
 
-  # As of this writing (28 June 2018), the latest version is
-  # 3.6.0, which doesn't works with JRuby. 
+  # As of this writing (12 March 2019), the latest version is
+  # 3.7.0, which doesn't work with JRuby.
   # See https://github.com/google/protobuf/issues/1594 
   # 3.1.0 works with JRuby, but fails with MRI 2.4.4 and above.
   #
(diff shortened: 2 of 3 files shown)


luke-hill:
I'm aware we could dive into this and maybe experiment more with restrictions, but I'd rather just get this in. As it's blocking a whole heap of other PR's which have failing CI

cucumber-messages/ruby/cucumber-messages.gemspec
```diff
@@ -29,8 +29,15 @@ Gem::Specification.new do |s|
...
-  # 3.6.0.
-  s.add_dependency 'google-protobuf', '~> 3.2'
+  # a version either around 3.2 (If they are on an old Ruby), or 3.7 if they
+  # are running on Ruby 2.6. This is due to some extensions failing to compile
+  # which has only been recently fixed and won't be backported
+  #
+  if RbConfig::CONFIG["MINOR"] == "6"
+    s.add_dependency('google-protobuf', '~> 3.7')
+  else
+    s.add_dependency('google-protobuf', ['>= 3.2', '< 3.6'])
```
mvz: I remember this kind of logic not being included in the gem once it's built. Does this work now?

luke-hill: I ran this locally on jruby 2.2 and 2.6 and it seemed to work.

That's not to say that's a rigorous test. Just what I'm capable of doing locally. I think we should merge this as is. And if it breaks things we can always revert it.

cucumber-messages/ruby/Gemfile
```diff
@@ -1,6 +1,3 @@
 # frozen_string_literal: true
 source "https://rubygems.org"
 gemspec
-
-# Pick a known-good version when running builds for cucumber-messages on JRuby.
-gem 'google-protobuf', '~> 3.2.0.2', platform: :jruby
```
mvz: I don't know why the build doesn't pick this up, or even if the build builds on any jruby, but this change makes the gem installation phase of the build fail locally for me with jRuby 9.2.5.0 and jRuby 9.2.6.0.

cucumber-messages/ruby/cucumber-messages.gemspec
mvz: I'm worried about what happens when a gem is build and pushed.

Can you try `gem build` please and see if the resulting gem file contains this logic or just picks one of the two `add_dependency` lines?

luke-hill: I've built it, but I cannot inspect it as it's all machine code. Is there a way I can directly test if this logic is here? Might need help to maybe add it to a gemfile or something

mvz: You can extract the data using, e.g.:
```
tar xvf  cucumber-messages-2.1.1.gem 
```

This will give you
```
metadata.gz
data.tar.gz
checksums.yaml.gz
```

Then you can look at the metadata:
```
gunzip metadata
more metadata
```

luke-hill: Nope, it's not pulling through the conditionals. Urm I'll revert the bit in the gemfile. See if that fixes things. After that I'm not sure.

Although (cynically), I'd rather get it working on 2.2-2.6 than on JRuby. So I'll give it a bit more of a whirl, but if I can't make any progress, I'd still like to get this merged in as is, and then maybe try fix the jruby stuff later?

luke-hill:
Oki, locally I've tested this again. Plus building the gem.

I can build on 2.2 - 2.6 and JRuby.
I can run and pass rspec on 2.2 - 2.6 and JRuby.

I'll try testing the meta-data using the same methods just now.

cucumber-messages/ruby/cucumber-messages.gemspec
mvz: Resigning to fixing JRuby later is probably the best course of action at the moment.

luke-hill:
I'm going to merge this then @mvz, and I'll raise a separate issue for tackling JRuby

take-cheeze:
Thanks for the fix! 👍 
May ask for a release plan for this fixes to `cucumber-messages` gem?

aslakhellesoy:
@luke-hill thanks for the fix. Unfortunately this won't work. As @mvz pointed out, the gemspec is only evaluated at build time, not at install time.

@tooky and I changed this in https://github.com/cucumber/cucumber/commit/3cba369c483c7e6b44cf81388140c2204035dc25#diff-e363e0e386df0084a887f93a1461d054

JRuby and MRI<2.6 users should add an explicit dependency on `google-protobuf=3.2` in their `Gemfile` instead. I think that will work.

aslakhellesoy:
> May ask for a release plan for this fixes to cucumber-messages gem?

@take-cheeze we don't do release plans because nobody is currently paid to work on Cucumber. It will happen when it happens.

PR #581, opened 2019-03-16
cucumber-expressions: Simplify java heuristic

mpkorstanje:
## Summary

Explaining when a string is interpreted as a cucumber expression should
be rather straight forward. In essence the explanation should be:

    If you surround a string with `^` and `$` it will be a regular
    expression. If not, it is a cucumber expression.

However the heuristic complicates this by checking if a capture group
contains regular expression symbols `[]+.*` with the assumption that
these are not used in normal writing.

This allows us to write little marvels like:

    this looks\( i.e: no regex symbols) like a cukexp
    a heavy storm forecast \(BF {int}+)
    the temperature is (\+){int} degrees celsius

Which look deceptively like cucumber expressions but are in fact regular
expressions. Simplifying the heuristic remove this ambiguity. This does
comes at the cost of being able to recognize obvious regular
expressions.

    this (.+) like a regexp
    this (\d+) like a regexp
    this ([a-z]+) like a regexp

But I think the simplicity is worth it.

## Motivation and Context
  * https://github.com/cucumber/cucumber-jvm/issues/1581
  * https://github.com/cucumber/cucumber/issues/515
  * https://github.com/cucumber/cucumber-jvm/issues/1581

## Types of changes

## Checklist:

### DIFF
--- a/cucumber-expressions/java/heuristics.adoc
+++ b/cucumber-expressions/java/heuristics.adoc
@@ -21,13 +21,4 @@ The table below describes the heuristics, along with some examples
 |RegularExpression
 |Forward slashes always assumes Regular Expression. The slashes themselves are removed.
 
-|this (.+) like a regexp
-|RegularExpression
-|If there are capture groups, assume Regular Expression.
-
-|this look(s) like a cukexp
-|CucumberExpression
-|Parenthesis containing only alphabetic characters are not considered Regular Expression capture groups,
-but Cucumber Expression optional text.
-
 |===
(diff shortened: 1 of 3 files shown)


aslakhellesoy:
I wonder if requiring both `^` and `$` is a little too strict. I think there might be a few regexps out there that only use one of the anchors.

How about saying it's a regular expression if it starts with `^` or ends with `$`?

aslakhellesoy:
Looking at the code, it looks like you've implemented my suggestion, but the text above says:

```
If you surround a string with `^` and `$` it will be a regular expression.
```

I think it should read:

```
If you start a string with `^` or end it with `$` it will be interpreted as a regular expression.
```

mpkorstanje:
Mmh. That's rather sloppy.

mpkorstanje:
This is a breaking change. The major version number should go up with the next release of Cucumber Expressions.

PR #612, opened 2019-05-07
Updated Documentation for Cucumber contributors

luke-hill:
## Summary

Updated the Contributing doc with some tweaks and removed an outdated doc (RELEASE_PROCESS), in favour of the more up to date one.

## Motivation and Context

After Cukenfest, we all got some learnings about how things work, so this is reflected here.

## How Has This Been Tested?

N/A

## Types of changes

## Checklist:


luke-hill:
I added a few people as reviewers from cukenfest. Apologies for the diff. I deleted one file and then renamed the other one as the original file, which I think it doesn't like (So there's 0 changes in the file)

mpkorstanje:
No plausible reason for the build to fail on this PR. I'm going to assume it's a fluke. If it isn't I'll roll it back

PR #613, opened 2019-05-09
add specs for TestRunStarted message

david1995:
## Summary

This is the start of writing specs for Cucumber messages as discussed on the Community Day on the CukenFest.

## Details

First scenarios for TestRunStarted message.

## Motivation and Context

All implementations of Cucumber should behave the same way.


cucumber-messages/specs/receiving/TestRunStarted.feature
```diff
@@ -0,0 +1,17 @@
+Feature: Receiving TestRunStarted messages
+
+Scenario: Handling a message queue with no TestRunStarted message
+  Given there is no TestRunStarted message in the message queue
+  When the messsage queue is processed
+  Then there is an error
+  And the queue processing is stopped
+
+Scenario: Handling a message queue with a singe TestRunStarted message
```
luke-hill: single*

cucumber-messages/specs/receiving/TestRunStarted.feature
```diff
@@ -0,0 +1,17 @@
...
+  Given there is a TestRunStarted message in the message queue
+  When the message queue is processed
+  Then the start of the test run was recognized
+
+Scenario: Handling a message queue with two TestRunStarted messages
+  Given there are two TestRunStarted messages in the message queue
+  When the message queue is processed
+  Then the second TestRunStarted message has been ignored
```
luke-hill: Should this check that the start of the run was recognized?

cucumber-messages/specs/receiving/TestRunStarted.feature
david1995: Yes, you're right. Already corrected.

cucumber-messages/specs/receiving/TestRunStarted.feature
david1995: Yes, this check should not be omitted. @SabotageAndi what do you think?

david1995: I've added the same Then statement as in the scenario above, which should clarify the behavior.

cucumber-messages/specs/receiving/TestRunStarted.feature
```diff
@@ -0,0 +1,17 @@
+Feature: Receiving TestRunStarted messages
+
+Scenario: Handling a message queue with no TestRunStarted message
+  Given there is no TestRunStarted message in the message queue
+  When the messsage queue is processed
+  Then there is an error
```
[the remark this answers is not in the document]
david1995: Your first suggestion sounds best to me. What do you think about `Then an error indicating that no TestRunStarted message could be found in the queue should be shown`?

mpkorstanje:
In Cucumber JVM the test run started event has a payload, a timestamp to indicate when the run started. Everty other event also has a timestamp. 

So we could add a ISO 8601 time stamp. The time should always be in UTC. 2019-05-18T07:18:58Z

This also raises a few questions:
 * Do we want to include that now or start with a truely minimal spec.
* Do we allow extensions to the spec?

I also think we are missing some principles about the channel. How can it be used? Currently only the channel can only be used by a single cucumber executor at a time. Is this intentional?

If we are to allow concurrent execution of pickles, why also not concurrent execution of executors?

Either way, adding an executor UUID to each event would allow us todistinguish the events from different executors on the same channel.

SabotageAndi:
@mpkorstanje We have another branch where we add a timestamp to the TestRunStarted message. We didn't send it yet as PR, because we wanted first to get feedback on the formulation of the specs, so we don't have to redo the specs for it multiple times.
We will send a PR for it in the next hours.

About the executor UUID: would be fine for me. Reason is, that because of how the SpecFlow+Runner is implemented, we will send multiple TestRunStarted messages (one for each Thread). With that, we could differentiate them.


@gasparnagy 
Yes, these scenarios are a little bit too abstract. The next PRs will be more concrete.

After discussion here, we removed the receiving side. We wrote it because of the behaviour of the SpecFlow+Runner (multiple messages). We solve this now with a different scenario.

About the term testsuite: that's one of the reason for this small PR. To get the first feedback to the problem domain language.

SabotageAndi:
PR for adding timestamp is here: https://github.com/cucumber/cucumber/pull/615

mattwynne:
I'm really glad to see us using feature files to specify these messages!

I'd like to see some concrete details about what is in the message. I would prefer to see something like #615 as the first iteration, rather than having this very SpecFlow-specific stuff in the second scenario about running in parallel.

mpkorstanje:
> About the executor UUID: would be fine for me. Reason is, that because of how the SpecFlow+Runner is implemented, we will send multiple TestRunStarted messages (one for each Thread). With that, we could differentiate them.

So now I'm a bit confused by your explanation. Say I have 12 pickles and I execute them in parallel on 3 threads. Will this mean that these are separate 3 test runs of 4 pickles.How would a listener to the event stream know that the original 12 pickles are all done?

This sounds like a rather significant difference. In Cucumber JVM one executor sends a single started/finished event and spins up several threads to execute the pickles in between these events. It is possible to have multiple executors running in parallel. They'd be executing different runs.

In a diagram:

```
System --- 1...n --> Cucumber Executors -- 1...n --> Threads
```

```
System
 |- Executor 1 executing pickles 1...12 tagged with @foo
 | |- Thread 1.1 Executing pickles 1...4
 | |- Thread 1.2 Executing pickles 5...8
 | |- Thread 1.3 Executing pickles 9...12 
 |- Executor 2 executing pickles 1...12 tagged with @bar
 | |- Thread 1.1 Executing pickles 1...4
 | |- Thread 1.2 Executing pickles 5...8
 | |- Thread 1.3 Executing pickles 9...12 
```

aslakhellesoy:
I think it's great to see progress towards a shared test suite for all Cucumber implementations.

Cucumber messages are fairly low level implementation details, and I am not convinced Gherkin is the best way to express how the messages should be constructed and processed. It feels to me like some sort of *approval test* suite is better suited. This is the approach taken for Gherkin implementations.

We also need to distinguish between the messages themselves and the Cucumber implementations producing/consuming these messages. This `.feature` file seems to be about validating the latter (a Cucumber implementation). The monorepo doesn't host any implementations (yet), so there is no way to actually *run* this feature file in the monorepo. Developers could easily introduce inconsistencies between the `.proto` files and the `.feature` files without noticing they have broken something. This violates an important principle in my book: *tests should be executable and fail when something is wrong*.

I think we need to take a step back and agree on what we want to achieve with a shared test suite *before* we start implementing it. 

* What kind of confidence are we looking for?
* What parts of a Cucumber implementation are we looking to test?
* What's out of scope of the test suite?
* etc...

SabotageAndi:
I created a new PR with all our current scenarios: https://github.com/cucumber/cucumber/pull/647

PR #616, opened 2019-05-20
cucumber-messages: add cucumber-implementation to TestRunStarted message

SabotageAndi:
PR is based on https://github.com/cucumber/cucumber/pull/615

## Summary

Ths PR adds a field `cucumberImplementation` to the `TestRunStarted` message. 
The reason for is, that you know on the receiving end who generated the cucumber-messages.

## Motivation and Context

Cucumber-formatters know which cucumber implementation generated the messages.

## Types of changes


cucumber-messages/specs/TestRunStarted.feature
```diff
@@ -0,0 +1,22 @@
+Feature: Sending TestRunStarted messages
```
aslakhellesoy: How is this feature file processed? I don't see anything running it in this repo.

I'm not convinced Cucumber is the right tool to test the messages. I'm not even sure the messages need to be "tested" as they have no logic, they are just structs.

aslakhellesoy:
As a general principle I would like us to adopt a consumer-driven contract mindset when evolving the protocol.

Do we have any consumers of this message yet? If we do, what will the consumer(s) do with the `cucumberImplementation` information? Are there other things we want to know about, such as version, OS, CPU etc?

aslakhellesoy:
This assumes that a test run executes all scenarios on the same platform. This would not be the case for a distributed execution with multiple agents. For example, some scenarios could be executed on SpecFlow and others on Cucumber.js.

I think it might make more sense to add this information to the `TestCaseStarted` message.

Moreover, I wonder if it might make more sense to wrap all metadata about the execution platform:

```proto
message TestCaseStarted {
  string pickleId = 1;
  google.protobuf.Timestamp timestamp = 2;
  Platform platform = 3;
}

message Platform {
  // The runner implementation. For example "SpecFlow", "Cucumber-JVM", "Behat" etc.
  string implementation = 1;
  // The version of the runner
  string version = 2;
  // The operating system
  string os = 3;
  // The CPU architecture
  string cpu = 4;
  // More fields here as needed...
}
```

WDYT?

aslakhellesoy:
This is superseded by #626

PR #625, opened 2019-06-04
Refactor/bugfix: Docker tweaks to default makefiles for Ruby

luke-hill:
## Summary

Try to alleviate some missing gem failures whilst running docker tasks

## Details

Use a double-check bundle before running rspec (gem), command

## Motivation and Context

When running Docker locally I often get an issue that rspec isn't installed, when I add `bundle` infront of the rspec command it fixes it.

This won't affect things if they're already installed, as `bundle install` or `bundle` only install missing gems and always adhere to the lock file.

## How Has This Been Tested?

Locally / CI

## Types of changes

## Checklist:

### DIFF
--- a/.templates/ruby/default.mk
+++ b/.templates/ruby/default.mk
@@ -18,7 +18,7 @@ Gemfile.lock: Gemfile $(GEMSPECS)
 	touch $@
 
 .tested: .deps $(RUBY_SOURCE_FILES)
-	bundle exec rspec --color
+	bundle install && bundle exec rspec --color
 	touch $@
 
 clean: clean-ruby
--- a/c21e/ruby/default.mk
+++ b/c21e/ruby/default.mk
@@ -18,7 +18,7 @@ Gemfile.lock: Gemfile $(GEMSPECS)
 	touch $@
 
 .tested: .deps $(RUBY_SOURCE_FILES)
-	bundle exec rspec --color
+	bundle install && bundle exec rspec --color
 	touch $@
 
 clean: clean-ruby
--- a/cucumber-expressions/ruby/default.mk
+++ b/cucumber-expressions/ruby/default.mk
@@ -18,7 +18,7 @@ Gemfile.lock: Gemfile $(GEMSPECS)
 	touch $@
 
 .tested: .deps $(RUBY_SOURCE_FILES)
-	bundle exec rspec --color
+	bundle install && bundle exec rspec --color
 	touch $@
 
 clean: clean-ruby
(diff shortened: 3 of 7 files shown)


tooky:
@luke-hill I'm not sure this is the right solution. In the `Makefile`s there is already a provision to run `bundle install` to create a `Gemfile.lock`. This step is a dependency of the step you've changed.

I wonder if it fails if you've run bundle install on your local machine, meaning the `Gemfile.lock` exists and the dependency is satisfied.

If you run `make clean` before you run `make` does it solve the problem?

luke-hill:
I don't quite understand how to read those makefiles. But when I run the single docker command it fails.

My understanding of docker is that inside the image it should be clean? But again, I'm not particularly strong on these things.

I have a separate issue when I just try running `make` which I'll make an issue ticket for later on when I've got time. Suffice it to say now I'm contributing more I need to run more tests locally and I'm hitting a brick wall big-time.

tooky:
@luke-hill let me have a stab at trying to explain `Makefile` - I'm by no means an expert :)

`make`'s job is to build files, and you need to tell it which files to build, you do this by specifying a _target_ (the file you want to build), its _dependencies_, and the _commands_ which `make` will execute to build that file.

So for example:

```
Gemfile.lock: Gemfile $(GEMSPECS)
	bundle install
	touch $@
```

Here the _target_ is `Gemfile.lock` and the dependencies are `Gemfile` and whatever is in `$(GEMSPECS)`.

`make` looks at the target and if it's missing or out of date with the dependencies (by comparing `mtime`, I think) it then executes the _command_, in this case:

```
bundle install
touch Gemfile.lock
```

Some of the make targets are not real file targets, e.g. `clean` and these are specified using `.PHONY` target. This means that make will execute those targets.

I think, for some reason, that when you are running `make` in the docker container it already thinks the Gemfile.lock is up to date so it is not "rebuilding it", e.g. it is not running `bundle install`.

I assume the docker container shares the project directory with your machine? It doesn't have a separate file system? If you have run `bundle install` in the directory on your machine (rather than in the container) then `make` will think the `Gemfile.lock` is up to date, and will not "rebuild" it, and therefore not run `bundle install`.

I think its probably safest to do `make clean` inside the container before you run `make` especially if you've been working on the project directly on your machine beforehand.

cc @aslakhellesoy

aslakhellesoy:
That's a great summary @tooky! The docker container mounts the host's file system as a volume, and sees the same files. So when you run `make` in the docker container, it is going to look at these files and assume there is nothing to do, because the timestamp of `Gemfile.lock` is newer than `Gemfile`.

So when you swtich between docker and the host OS you should always `make clean` to force the dependencies to be reinstalled.

luke-hill:
So shouldn't we add a `make clean` command the the docker script that gets ran? Again I'm only going off what I've previously understood but docker is always meant to be clean blank canvas right?#

I'll look into doing `make clean` locally and see if that works, but I'm having a lot of different failures when I try to run local tests, either on docker or not.

luke-hill:
Closing after conversations on slack. A documentation PR will be forthcoming.

PR #635, opened 2019-06-14
[cucumber-messages] .NET library

SabotageAndi:
## Summary

.NET library for cucumber-messages

## Motivation and Context

We need it for SpecFlow

## How Has This Been Tested?

Already used in SpecFlow

## Types of changes

## Checklist:

- [x] The change has been ported to .NET.


aslakhellesoy:
Now that Microsoft has released a Docker image with .NET, we can base our own image on that one. We no longer need a separate Docker image for .NET - one image to rule them all :-)

I also managed to get the prebuilt `protobuf` package for Alpine Linux working (with a small hack in `Dockerfile`) so we don't need to build `protobuf` from scratch every time we build a new image.

The `protoc` inside the nuget package is broken on Alpine, so the `Makefile` will now overwrite it with the official Alpine `protoc` if the build runs in Docker.

I'm now able to build `cucumber-messages/dotnet` inside Docker:

```
docker run -v "$(pwd)":/app -it cucumber/cucumber-build:53bdf15c8c2e8fab1fab780056642239
cd /app/cucumber-messages/dotnet
make clean
make
```

It looks like the `protoc` is being overwritten again, causing the build to fail: https://travis-ci.org/cucumber/cucumber/jobs/553286928#L1831

I've got an idea how to fix that...

SabotageAndi:
I think I solved the protoc problem.
I added an restore command before the build, so the NuGet packages are downloaded and only after that the protoc binary is replaced.

There is still one error left with .NET Framework 4.5.

SabotageAndi:
Thanks to @dasmulli it looks like the error for .NET Framework 4.5 is now also fixed.

PR #642, opened 2019-06-22
Protocol design and test strategy

aslakhellesoy:
This is a first cut of a design documentation and test strategy for Cucumber. It starts by outlining 3 main components in Cucumber. It then describes how these components communicate over a message protocol.

It goes on to describe 3 different kinds of tests, along with a build system that will generate approval test files as well as living documentation. This is done in a different way from conventional BDD. The test data and documentation are generated *from* the source code (by annotating the `.proto` file).

So far I have only implemented the test data generation. What I plan to do next is to implement the approval test suite and run it against @charlierudolph's `cucumber-engine`.

Then I hope we can start refactoring the native Cucumbers towards becoming front-ends. This should allow us to slim down the codebase considerably.

I am aware there is some reluctance to adopt a go-based engine. It would have to be installed separately, and could either run as a daemon and listen on a socket, or it could be spawned by the front-end and use `STDIN/STDOUT`.

This does sound a bit scary distribution/installation-wise, I admit that. But I also think it can be a great opportunity both for the users and contributors. Take docker for example. When the installer has run you have a little icon in your task bar where you can do all sorts of configuration. It updates itself when there is a new version. We could evolve it to be a provider of external plugins, such as custom reporters or even test parsers. That would make it really easy to support new formats such as Markdown, Excel or custom formats. This would allow users to truly take advantage of domain-specific models and connect them directly to acceptance tests.

Anyway, have a look and let me know what you think. We can discuss it next Tuesday at the weekly open source call.


aslakhellesoy:
Whoops I've also made some somewhat unrelated changes to `messages.proto`. Blame the boy-scout rule!

chassa:
Thanks for clarifying the testing strategy. I would like to better understand your intention for using Approval Tests.

Cucumer Messages on their own have no value or function that can be tested. They realize their value only when supporting concrete integrations between FrontEnd – Runner or Runner – Reporter. This is what IMHO the testing strategy for Cucumber Messages should describe. Ideally, there would be also a blue-print or specification to build a FrontEnds, Runners or Reporters, on how to integrate through Cucumber Messages with other FrontEnds, Runners or Reporters.

We are currently extending SpecFlow and SpecFlow+ as concrete Engines to support Cucumber Messages. To be more concrete: we want SpecFlow and SpecFlow+ to behave similarly as other Cucumber Engines (legacy and the cross-platform), when it comes to communicating with FrontEnds and Reporters through Cucumber Messages. 

This is why we can’t just “implement” Cucumber Messages into SpecFlow/SpecFlow+. We can implement behaviour towards a FrontEnd and/or a Reporter. As a first step, we’re focusing on SpecFlow/SpecFlow+ integrating as Engine to Reporters through Cucumber Messages.

Our decision was driven by what we have jointly agreed on in London at the Cucumber Community meeting: start with integrating Engines with Reporters to add immediate value to our users. Since no Reporter exists yet, we also need a first version of a Reporter that integrates with Engines through Cucumber Messages. To test the general ideas of cross-platform development in the community, we’ve started a cross-platform Reporter to consume Cucumber Messages.

Describing their behaviour of integration from the Engine (SpecFlow/SpecFlow+) and Reporter (first cross-platform version) perspective greatly helps us to move forward in focused way, document what we do, and also spot deviations in the different Engine implementations (e.g. SpecFlow/SpecFlow+).

The only thing that I'm not sure is, whether those specifications should be part of the abstract Cucumber Messages  project, or just concrete FrontEnds, Engines or Reporters that implement Cucumber Messages. Messages itself can never be implemented on their own. Their design emerges and changes from the different use cases we build around Engine-Reporter and FrontEnd-Engine integrations. Still, we could maintain a reference specification to help implement another FrontEnd, Engine or Reporter that integrates successfully with other Cucumber Family components.

My understanding of the purpose for Approval Tests (at least when I discussed the idea with Llewellyn 2012 in Dallas on the AAFTT Open Space) is that they are useful if it is hard to come up with abstract rules and a strict model to *describe* the assert. E.g. if there are large amounts of data to be described and/or when there is no clear set of rules or good description language on what to assert.

Instead of describing the assert as examples that satisfy rules, Approval Tests take a snapshot of the result which has been validated once as “correct” by a human in an exploratory way (without having a concrete set of rules to assert). An approval test tool can streamline the process of manual re-validation and provides tools for fuzzy result-set assertion.

Asserting graphic output is a good case, @Gojko gives a very good example with @AppraiseQA (see also: https://gojko.net/2017/10/09/introducing-appraise.html).

I currently don't see any of these challenges for describing asserts that test the proper integration of concrete FrontEnd-Engine and Engine-Reporter implementations. Instead, I think that describing the behaviour of those integrations properly in Gherkin is a perfect example for Specification-By-Example. 

You could write tests against a concrete Cucumber Engine that implements *already* Cucumber Messages, to assert the resulting message stream, assuming that this is the "correct" implementation. However, I wouldn't call this approval testing but rather an end-to-end testing safety net approach as described by @michaelfeathers in his book "Working Effectively with Legacy Code". This is if you need to keep a legacy system stable while you refactor it. But do we already have a concrete FrontEnd - Engine or Reporter - Engine combination implemented, that we want to define as "legacy reference" for all the other upcoming implementations? 

Approval Tests should be only applied to concrete end-to-end usage scenarios, eg. for an Engine and Reporter successfully producing a "proper" report for test runs. In that case you could assert manually (by looking at it) if the produced report looks "proper", and if so, approve the captured messages as successful result to assert against. Again, I currently would rather describe the behavior of concrete Engines and Reporters that use messages to produce something useful for the user (e.g. perform a report for a test run), as long as rules can be expressed concretely and examples can be described properly in Gherkin. But for certain (graphical) reports, it might be useful for particular Reporters to use approval tests later.

I'd be happy to discuss this in detail in our call on Monday.

mpkorstanje:
>  https://github.com/cucumber/cucumber/pull/642#issuecomment-504737279
> You could write tests against a concrete Cucumber Engine that implements already Cucumber Messages, to assert the resulting message stream, assuming that this is the "correct" implementation. However, I wouldn't call this approval testing but rather an end-to-end testing safety net approach as described by @michaelfeathers in his book "Working Effectively with Legacy Code". This is if you need to keep a legacy system stable while you refactor it. But do we already have a concrete FrontEnd - Engine or Reporter - Engine combination implemented, that we want to define as "legacy reference" for all the other upcoming implementations?

I'm going to second this. As a development strategy I don't believe that starting with the Cucumber Engine is the right approach. 

@aslakhellesoy one of your stated goals is to reduce the size of Cucumbers code base. The message protocol is a step towards achieving that goal. However I believe that the goal of implementing the message protocol itself is to provide a unified format for tooling to report on cucumbers execution. I also believe this the goal that we are trying to achieve now.

As such starting with the Cucumber Engine would not be beneficial to achieving that goal. It would add quite some lead time before we can validate the correctness of the message protocol. Additionally it would bet everything on a successful go-hybrid adoption. You'd metaphorically burning the ships.

mpkorstanje:
I am going to need some more time to process everything else that has been said.

mpkorstanje:
If the goal is to focus on the message protocol then I believe the following are the necessary and sufficient components to be successful. As mentioned at Cukenfest:
 
 0. A description of the message protocol
 1. An implementation of a message protocol message producer
 2. An implementation of a message protocol message consumer
 3. A state machine / state machine description
    * Used to validate the order of messages produced
    * Used to generate message in random / all possible valid orders

The state machine essentially describes the contract between producer and consumer.

gasparnagy:
@chassa I think the highlight on approval testing was a bit misleading in @aslakhellesoy 's summary. Actually what we need is to define a contract and a technique that can be used to verify whether a particular implementation properly handles that contract (this is often called _contract testing_). Here in our case, since the implementations use/need different platforms, implementing the assertions in the contract tests is problematic and this is what approval testing can support. (This is nothing new, BTW, this is how we have tested the Gherkin parser as well.) 

Besides all this, the clean separation of the intention/behaviour and the mechanics of the contract details is anyway useful, as you have said.

chassa:
I agree, @gasparnagy, we don’t need to argue how to call things, as long as we are aligned on our intentions.

My understanding is, that we want to have a reliable contract between different implementations of FrontEnds, Reporters and Engines in the Cucumber eco system that will evolve over time, as the eco system grows.

For that I think the following two questions are key to answer:
1.	What drives the design of the contract and how do we collaborate on that?
2.	How do we document and share the current design of the contract (which will evolve over time)?

For Q1 I think it is key to define the concrete scenarios for which the contract should emerge, e.g. integration Engine – Reporter. Defining Questions -> Rules -> Key Examples on how the integration should behave is IMHO a good approach for this.

Q2 should follow mostly on how we collaborate in Q1 to keep up with an always up-to-date documentation. I’d imagine Gherkin scenarios would be a natural choice for the Cucumber community.
My current concern is that we skip answering Q1 and discuss options (like state machines or protobuf definitions with inline comments) for Q2, which are completely disconnected from Q1.

SabotageAndi:
Perhaps the approval tests could be an addition to the specification in Gherkin. But so that somebody is able to implement when which message is created, I think approval testing isn't enough. I think it would feel for everybody as reverse engineering something.

I never liked the approval tests of the Gherkin parser.
 - it's not really easy for a human to check a 140 line json file if it looks correct (e.g. https://github.com/cucumber/cucumber/blob/master/gherkin/testdata/good/docstrings.feature.ast.ndjson)
 - I see what should be the result, but I don't know how I get to the result.
 - I never knew, if the actual result is a "better" result than the current 

From the little specs we already wrote for cucumber-messages, we got new insights, that we didn't saw only from the `messages.proto` file. I would miss them, if we didn't had them.

mpkorstanje:
@chassa
> Examples on how the integration should behave is IMHO a good approach for this.

I believe those are necessary but not sufficient. Consumer, Producer and State machine each cover different aspects in our process that the other two can't quite cover.

* The consumer informs us about the information that should be included in the event stream.
* The producer informs us about the events we can provide and their shape.
* The state-machine describes possible orders in which the events may be produced and could be consumed.

The examples would do quite well to describe behavior of the producer. They can't however capture the non-deterministic behavior inherent in concurrent execution. Nor can they enumerate all possible sequences of events.

In my experience not designing for concurrency from the ground up is painful.

mpkorstanje:
Design for the cucumber message system.

IMG_20190627_213158

The producers produces messages that are put into some channel. Multiple producers can use the same channel at the same time. Multiple consumers can listen to a channel. Each consumer will receive a copy of the message put into the channel.

A message consists of `producerId` which uniquely identifies the producer and an `event`.

IMG_20190627_213215

Same system again. But now focused on the interaction between producer and consumer. Both consumer and producer have the same state transition diagram. 

The consumer is a validating consumer. As the producer sends messages to the consumer the consumer transitions based on the message received. If the consumer can't transition the message order was invalid and it will be reported.

The producer enumerates all possible message sequences. It does by taking the state transition diagram and selecting an arbitrary transition and sending a message appropriate to this transition. 

IMG_20190627_213239

A class diagram of the minimal set of event types needed to represent everything that happens in a theoretical cucumber execution. I haven't fleshed the contents out much. That can come later on. Important attributes are the timestamp on all events and for all test case events a unique test case identifier.

IMG_20190627_213258

The state transition diagram as show in figure 2, now in detail.

This is a hierarchical state machine diagram. When entering a state machine state it is not possible to transition out of that state until the final state has been reached.

Also important to note is the e-transition. For clarity I've used a Non-deterministic finite automaton. The e-transition can be invoked arbitrarily without prompting.

In section A the execution of a producer is started by emitting the `TestRunStarted` event. Nothing need happen and it can immediately be followed by a `TestRunFinished` event. In between those two events the state machine is allowed to fork into _n_ state machines that execute in parallel. But it can not resume until all forks have been joined.

In section B we see the implementation of state machine B from section A.

Before any `TestCaseStarted` event 1 or more `TestSourceRead` events are emited. Once the test case is started it will be executed by machine C. When finished a `TestCaseFinished` event is emmited.

There is a little bit of contextuallity that isn't possible to capture with an NDFA. The test source that has been read must match the executed test case.

In section C we see the implementation of state machine C from section B.

It is a simple repetition of `TestStepStarted` and `TestStepFinished` events between any number of `Embed` and `Write` events can happen. Note: In this representation hooks are also considered steps.

A more accurate representation would have 5 copies of statemachine C connected head to tail to cover for before hooks, before step hooks, steps, after step hooks, and after hooks.

mpkorstanje:
I think the next step would be trying to validate if the different Cucumber implementations can be described by this system. If not we have a to have a discussion about the _correct_ implementation of Cucumber first.

Validation could be in the form of a reporter/plugin that outputs a set of minimal events and an implementation of the validating consumer. The reporter/plugin and will also provide us with a nice base system to start iterating further on.

As for the consumer to drive contents of the messages, I believe we specified a minimal example at Cukenfest. It was essentially a feature file augmented with `X`s and `V` behind each step and example. A `V` indicates a pass, an `X` a failure.

I.e:

```
# Comment
@tag
Feature: Eating too many cucumbers may not be good for you

  Eating too much of anything may not be good for you.

  Scenario: Eating a few is no problem
    Given Alice is hungry                              V 
    When she eats 3 cucumbers                  V
    Then she will be full                                X
```

aslakhellesoy:
Great insight and illustrations @mpkorstanje! I'm looking forward to studying them and listening to you explaining it :-)

I think the living documentation and executable specification for the
protocol should use a representation that is specific to the domain of protocols.

Protocols are state machines, and the Cucumber Protocol is no different.
There are many ways to document state machines. The most common representation is
[State diagrams](https://en.wikipedia.org/wiki/State_diagram). 
[Petri nets](https://en.wikipedia.org/wiki/Petri_net) can also be used, and they
are sometimes more concise, but since state diagrams are more widely understood
I think we should try that first.

Diagrams convey a lot of information in an way that makes a protocol quick and 
easy to understand compared to long-winded textual representations.

Doodling state diagrams on paper seems like a good start, and once we've agreed on the basics
we can formalise it with a [state transition table](https://en.wikipedia.org/wiki/State_transition_table).
This could be a CSV file.

We can then generate a state diagram from the CSV using a bit of scripting and [Graphviz](https://www.graphviz.org/)

@ciaranmcnulty explained this approach nicely in his [Doodle Driven Development](https://www.slideshare.net/CiaranMcNulty/doodle-driven-development) presentation at Cukenfest this year.

The transitions in the machine correspond to protocol messages (defined in the `.proto` file).
With a bit more scripting we can also generate documentation with examples from the CSV file
and the `.proto` file(s). The examples could be generated from annotations in the `.proto`
file as I have started to do in this PR.

In other words, the CSV and the `.proto` files would represent the single source of truth,
and we can use it to generate:

- code for the protocol handlers (using a simple parser generator)
- automated tests
- living documentation (with results from the tests)

This checks all the BDD boxes for me. It doesn't use Cucumber, but I think that is
absolutely fine. I believe in choosing the best tool for the job, and I think
Gherkin falls short in this domain, particularly because it is not visual.

We can still practice TDD with this approach. The first test would be generated from 
a simple embryonic transition table with two nodes and one transition. When this test
passes we add another node or transition and keep going until we've implemented the
whole protocol.

aslakhellesoy:
Hey @ciaranmcnulty, I'd love your insight on this.

Issue #643, opened 2019-06-25
cucumber-expressions: All regex lookarounds are incorrectly counted as a capturing group causing an iterator overflow and a NoSuchElement Exception.


benariss:
## Summary

When using any lookaround (positive lookahead '(?=..', negative lookahead '(?!..', negative lookbehind '(?<!..' or positive lookbehind '(?<=..') in a regular expression Cucumber will encounter a NoSuchElement exception when trying to parse a matching gherkin.  This is the same issue described in #481 and fixed for negative lookaheads in #593 and is being caused by Cucumber treating the lookaheads as a capturing group causing a mismatch between the number of groups to process and the iterator count causing it to overflow and trigger the exception.

## Expected Behavior

Cucumber should be able to successfully parse a regex match which contains any lookaround group.

## Current Behavior

Currently if you use any of the regex lookaround groups Cucumber treats them as a capture group, requiring you to add a parameter to the step definition for them.  
This also causes a discrepancy between the number of capture groups to be processed and the number of group iterations that are performed causing the iterator to overflow and the following exception stacktrace:

~~~
java.util.NoSuchElementException
	at io.cucumber.cucumberexpressions.TreeRegexp$IntRange.next(TreeRegexp.java:99)
	at io.cucumber.cucumberexpressions.TreeRegexp$IntRange.next(TreeRegexp.java:82)
	at io.cucumber.cucumberexpressions.GroupBuilder.build(GroupBuilder.java:18)
	at io.cucumber.cucumberexpressions.GroupBuilder.build(GroupBuilder.java:21)
	at io.cucumber.cucumberexpressions.TreeRegexp.match(TreeRegexp.java:75)
	at io.cucumber.cucumberexpressions.Argument.build(Argument.java:12)
	at io.cucumber.cucumberexpressions.RegularExpression.match(RegularExpression.java:54)
	at io.cucumber.stepexpression.StepExpression.match(StepExpression.java:22)
	at io.cucumber.stepexpression.ExpressionArgumentMatcher.argumentsFrom(ExpressionArgumentMatcher.java:21)
	at cucumber.runtime.java.JavaStepDefinition.matchedArguments(JavaStepDefinition.java:62)
	at cucumber.runner.Glue.stepDefinitionMatches(Glue.java:105)
	at cucumber.runner.Glue.stepDefinitionMatch(Glue.java:87)
	at cucumber.runner.Runner.addTestStepsForPickleSteps(Runner.java:64)
[...]

PR #644, opened 2019-07-02
cucumber-expressions - TreeRegexp: recognize lookaround as non-capturing groups

vincent-psarga:
## Summary

This should provide a solution for  issue #643

## Types of changes

## Checklist:


aslakhellesoy:
> Can we try keep the tests standardised

We already have `/cucumber-expressions/examples.txt` for shared tests, and I guess we should use that more. If you edit that file, sync it to the sub folders:

```
source ../../scripts/functions.sh && rsync_files
```
WDYT @luke-hill ?

vincent-psarga:
Just to ensure I understood @luke-hill comment correctly, on the Ruby example, you'd prefer me to rewrite:

```ruby
      it 'recognizes lookaround capture groups' do
        tr = TreeRegexp.new(/foo is a (?=[bar])(.*)$/)
        group = tr.match("foo is a big bar")
        expect(group.children[0].value).to eq('big bar')
      end
```

as 

```ruby
      it 'ignores `?=` as a non-capturing group' do
        tr = TreeRegexp.new(/a(?!b)(.+)/)
        group = tr.match('aBc')
        expect(group.value).to eq('aBc')
        expect(group.children[0].value).to eq('Bc')
        expect(group.children.length).to eq 1
      end
```

Is that right ? (and also for Java of course)

For the examples in examples.txt, I can add that:

```
/I have cukes in my (?=[b])(.*)$/
I have cukes in my belly
["belly"]
---
/I have cukes in my (?=[b])(.*)$/
I have cukes in my stomach
null
```

Although it is not 100% clear when reading that what we exactly try to match there, I like the idea of having clear tests describing what we are checking.
Having both of them makes more sense I think.

vincent-psarga:
Note about examples.txt: it may not be possible to add the examples there, as Go does not support some examples we want to exhibit here (typically ```(?=re)```, but maybe the other ones that were described in #643)

See: https://github.com/google/re2/wiki/Syntax

luke-hill:
Sounds like a plan RE standardised tests. But I've no idea how to implement them. I'm still reeling from the mass of work I had to do to get the last ones passing, so if it's easy to implement I'm for it.

cc/ @aslakhellesoy

luke-hill:
Also RE the re-writing of tests. Can we have everything inside one capture, not 2. As my understanding of look aheads e.t.c. is that they don't capture inside their group. So if you have a blank group it kinda negates it.

vincent-psarga:
> Can we have everything inside one capture, not 2

That sound doable :)

I'll try to complete the support of ```?<=``` before marking as ready :)

vincent-psarga:
I've added support for ```?<=``` and ```?<!```in Ruby.

That said I'm not 100% convinced by the way it's done (seems pretty brittle to me). If anyone as a better idea, it's much welcome :)

(I though about looking the last two characters but I get the exception about named groups getting raised before matching ```=``` or ```!```)

vincent-psarga:
@luke-hill is your change request still relevant after the changes I did or did I miss something else ?

cucumber-expressions/ruby/lib/cucumber/cucumber_expressions/tree_regexp.rb
```diff
@@ -55,6 +59,18 @@ def match(s)
         group_indices = (0..match.length).to_a.to_enum
         @group_builder.build(match, group_indices)
       end
+
+      private
+
+      def end_non_capturing_group
+        @stack.last.set_non_capturing!
+        end_group()
```
luke-hill: here and in a few others places we have redundant `()`

vincent-psarga:
Go does not support look around so that's why it's not checked here (well, the Go regex parser will fail when seeing lookarounds)
And for .NET, I guess the expressions are in Specflow, not this repo.

The redundant parenthesis should be removed now :)

aslakhellesoy:
> And for .NET, I guess the expressions are in Specflow, not this repo.

There is no cucumber-expressions implementation in .NET yet

luke-hill:
Genuine question: do we need something like this: https://github.com/cucumber/cucumber/pull/593/files#diff-992ea8ba4a1d5037dd6de00fe673baddR28

If not this is RtM (CI passing obviously)

mpkorstanje:
> Genuine question: do we need something like this: https://github.com/cucumber/cucumber/pull/593/files#diff-992ea8ba4a1d5037dd6de00fe673baddR28

Not really. Checks that don't try to create the regex are prone to false positives. I've removed it.

mpkorstanje:
I think we're good to go. 

If nobody has any remarks I'll merge this by the end of the week. But please feel free to do so earlier if you're think we're good to go.

aslakhellesoy:
> @aslakhellesoy looks like something external to this MR changed and broke gherkin-god and the dots-formatter

Yes, a time-bomb caused the build to break a week or so ago. I've fixed that on master, and merging master into this one should fix it.

PR #647, opened 2019-07-03
Specifications for the sender part of Cucumber-Messages

SabotageAndi:
## Summary

These are the scenarios we use to test the Cucumber-message implementation of SpecFlow.

## Motivation and Context

All implementations of Cucumber should behave the same way.

### DIFF
--- a/cucumber-messages/specs/TestCaseFinished.feature
+++ b/cucumber-messages/specs/TestCaseFinished.feature
@@ -4,7 +4,9 @@ Feature: Sending TestCaseFinished Messages
 
         Given there are <NumberOfScenarios> scenarios
         And all steps are bound and pass
+
         When the test suite is executed
+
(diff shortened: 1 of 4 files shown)


aslakhellesoy:
Since the monorepo doesn't contain any step definitions or scripts to run these scenarios, I'm wondering:

* Is there a CI process anywhere that runs these scenarios?
* Do these scenarios get copied into another repo where there are step definitions and a build script that runs them?
* If I change the `NumberOfTestCaseFinishedMessages` cell from `1` to `42`, will that cause a build failure anywhere? Where? How?

SabotageAndi:
In SpecFlow and SpecFlow+ we have Cucumber as a submodule included. In our build process, we copy the feature files into the location of our other scenarios.
There are also the step definitions for automation of the scenarions. They run in our normal CI and PR builds.

One thing we could do is, to automate the submodule update when there is a change in the monorepo. Because of this change a CI- build would be triggered.

Issue #645, opened 2019-07-02
Remove dist folders


aslakhellesoy:
We recently translated all the JavaScript code to TypeScript. Some modules depend on other modules, and they declare npm dependencies on git repositories rather than packages releasded on NPM.

For example: https://github.com/cucumber/cucumber/blob/e01e40fa851d45414bcff29ef4bfd981adb46efb/gherkin/javascript/package.json#L45-L46

We pin them to released versions prior to a release, and unpin them afterwards.

This means the transpiled TypeScript must be added to git - npm does not run `tsc` when you install a package - that's supposed to be done when the package is published.

Having generated code in git is a well-known headache - it causes annoying merge conflicts and noisy diffs, but we did it for a reason (as described above).

Now that we're more active developers I want to explore a different approach. I want to `git rm` all the `dist` directories. Instead, I want to publish an npm package for *every* build. We don't want to do that to the public npm repository, but to a special repository only used for builds.

It looks like [verdaccio](https://verdaccio.org/) might be a good way to do this. Let's give it a try:

- [ ] Set up a local verdaccio to test the waters
- [ ] Add `prepublish` to `package.json` so `tsc` runs when the module is released
- [ ] Try to publish to the local verdaccio, and verify that it's idempotent (overwriting should work).
- [ ] Set up a hosted verdaccio on AWS - see https://verdaccio.org/docs/en/amazon
- [ ] Tweak CI settings so regular builds publish there
- [ ] Delete all the `dist` directories

aslakhellesoy:
This is now fixed on master

PR #648, opened 2019-07-03
Migrate to CircleCI

aslakhellesoy:
Fix for #645


aslakhellesoy:
This PR started to address #645, but then evolved into a migration to CircleCI:

* Much faster builds (starts faster, runs in parallel)
* No more need to synchronise to read-only sub repos (we'll archive those)
* Pull request workflow works again since the build doesn't rely on other builds or published artifacts
* More secure. Releases will require manual approval by a release manager
* Automated update of dependencies

aslakhellesoy:
More changes in this PR:

* Use git-crypt to encrypt most secrets used for deployment (some are still in env vars)

PR #650, opened 2019-07-05
Fix #646 by checking containment within extraRows, so that cardinalit…

### LINKED ISSUE #646: DataTable.unorderedDiff is not symmetric and does not take row count into account
## Summary

`DataTable.unorderedDiff` semantics are ill-defined and arguable wrong

## Expected Behavior

I was expecting that `unorderedDiff`:

 * is symmetric, i.e. `a.unorderedDiff(b)` behaves the same as `b.unorderedDiff(a)` except that +/- signs are swapped in the output when a and b are different,
 * either behaves as an actual set difference, or (much better) as a multiset difference, i.e. given the current name, I'd expect it to throw if one datatable has 10 equal records and the other datatable has 1 record, also equal to those 10 records.
 * outputs diffs that include the correct number of +/- lines.

## Current Behavior

Currently, it is neither symmetric, nor behaves like set difference nor multiset difference.

Specifically, for `a.unorderedDiff(b)`, if b has more equal records than a, it throws, but if a has more records than b, it passes. Thus, it's not symmetric and only behaves like set difference in one direction.

~~~
~~~

## Possible Solution

I may come up with something later, but I think at least someone ought to decide on what the semantics should be, and then the method can be fixed to stick to those semantics.

I think making it a symmetric multiset difference reduces surprises the most, given the name, but obviously that makes it stricter than the current behavior, so it would be a breaking change. 

Alternatively, it may be an idea to deprecate this method and add new methods with more well defined semantics, e.g. `setDiff` and `multisetDiff`, the first disregarding both order and count, and the second only disregarding order.

This code shows some of the inconsistencies:

```java
  private static final DataTable DT1 = DataTable.create(ImmutableList.of(
      ImmutableList.of("foo"),
      ImmutableList.of("bar")
  ));

  private static final DataTable DT2 = DataTable.create(ImmutableList.of(
      ImmutableList.of("foo"),
      ImmutableList.of("foo"),
      ImmutableList.of("bar"),
      ImmutableList.of("bar")
  ));

  private static final DataTable DT3 = DataTable.create(ImmutableList.of(
      ImmutableList.of("foo"),
      ImmutableList.of("bar"),
      ImmutableList.of("bar"),
      ImmutableList.of("bar")
  ));

  @Test public void testDirection1() {
    DT1.unorderedDiff(DT2); // throws as expected
  }

  @Test public void testDirection2() {
    DT2.unorderedDiff(DT1); // passes despite the tables being different
  }

  @Test public void isItEvenASetDifference() {
    DT2.unorderedDiff(DT3); // throws despite Set(DT2) == Set(DT3)
    // also prints a diff that seems to suggest one of the tables should have 5 records?!?
  }
```

## Your Environment

* Version used: datatable-1.1.12 / cucumber 4.4.0

tvo:
See issue #646

### DIFF
--- a/datatable/java/datatable/src/main/java/io/cucumber/datatable/TableDiffer.java
+++ b/datatable/java/datatable/src/main/java/io/cucumber/datatable/TableDiffer.java
@@ -44,13 +44,12 @@ public class TableDiffer {
         extraRows.addAll(to.cells());
 
         for (List<String> row : from.cells()) {
-            if (!to.cells().contains(row)) {
+            if (!extraRows.remove(row)) {
                 diffTableRows.add(
                         new SimpleEntry<>(row, DiffType.DELETE));
             } else {
                 diffTableRows.add(
                         new SimpleEntry<>(row, DiffType.NONE));
-                extraRows.remove(row);
             }
         }
 
--- a/datatable/java/datatable/src/test/java/io/cucumber/datatable/TableDifferTest.java
+++ b/datatable/java/datatable/src/test/java/io/cucumber/datatable/TableDifferTest.java
@@ -262,17 +262,17 @@ public class TableDifferTest {
     }
 
     @Test
-    public void diff_set_with_itself() {
+    public void unordered_diff_with_itself() {
         assertTrue(new TableDiffer(table(), table()).calculateUnorderedDiffs().isEmpty());
     }
 
     @Test
-    public void diff_set_with_itself_in_different_order() {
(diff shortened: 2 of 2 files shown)


mpkorstanje:
Cheers. This looks al right but I'll try and work out the logic for my self as soon as possible. I don't think we have to consider this a breaking change but can in fact consider it a bug fix.

aslakhellesoy:
Hi @tvo,

Aslak Hellesøy
Creator of Cucumber

PR #656, opened 2019-07-18
Cucumber protocol petri net

aslakhellesoy:
This is a draft design document for the Cucumber Protocol.


aslakhellesoy:
I’ve implemented a simple Petri Net simulator. You load a Petri Net into it (`PetriNet.from_pnml(xml)`) and then fire transitions. If a transition is non-fireable, an error is raised. We can use this as a basis for validating a Cucumber implementation.

We’ll record the protobuf messages exchanged between front-end and back-end, transform those to a series of fired transitions, and pass them through the `PetriNet` validator. If an exception is raised, the implementation does not conform to the protocol.

Currently there is one failing test:
https://github.com/cucumber/cucumber/blob/f3f8022a2fbabbf2e8d26c33b68e612ddedb1435/cucumber-protocol/spec/petri_net_spec.rb#L47

So we have more work to do on the protocol definition (which we're editing in the PIPE editor). But at least this is starting to take shape.

aslakhellesoy:
Obsolete!

PR #659, opened 2019-07-19
cucumber-expressions: Prefer Java type hint over parameter type

mpkorstanje:
## Summary

When there is a conflict between the type hint from the regular expression
and the method prefer the the parameter type associated with the regular
expression. This ensures we will use the internal/user registered parameter
transformer rather then the default.

Unless the parameter type indicates it is the stronger type hint.

## Details

Reasoning:
1. Pure cucumber expression users will not notice this in either scenario.
2. Pure regular expression users will benefit because 
   `BuiltInParameterTransformer` can now seamlessly transform any captured
   values. (For all built in types `useRegexpMatchAsStrongTypeHint` is
   explicitly set to false.)
2. Regular expression users that define a default transformer have little
   need to define parameter types. The default transformer should be
   sufficiently powerful to meet their needs and will often allow users to
   add custom creation methods e.g. Jacksons @JsonFactory.
3. Users who mix regular and cucumber expressions may run into conflicts
   when a registered cucumber expression and unregistered happens to
   collide. However this was the situation before this flag was added.
4. Regular expression users who define custom parameter types do so with
   the expectation that the parameter will be matched. Subverting this
   expectation when the method signature does not match may result in a
   parameter transformer that is unable to convert to the desired type.
   Leaving the user puzzled as to why his transform was ignored.

## Motivation and Context   

Fixes: https://github.com/cucumber/cucumber/issues/658

## Types of changes

## Checklist:

### DIFF
--- a/cucumber-expressions/CHANGELOG.md
+++ b/cucumber-expressions/CHANGELOG.md
@@ -20,6 +20,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   ([#515](https://github.com/cucumber/cucumber/issues/515)
    [#581](https://github.com/cucumber/cucumber/pull/581)
    [mpkorstanje])
+* [Java/Go] cucumber-expressions: Prefer language type hint over parameter type
+  ([#658](https://github.com/cucumber/cucumber/pull/658)
+   [#659](https://github.com/cucumber/cucumber/pull/659)
+   [mpkorstanje])
 
 ### Deprecated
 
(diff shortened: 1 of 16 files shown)


aslakhellesoy:
Tip @mpkorstanje - rather than creating a PR prefixed with `WIP: `, you can mark the PR as a `Draft`. See https://github.blog/2019-02-14-introducing-draft-pull-requests/

mpkorstanje:
Cool. That's new. Thanks.

mpkorstanje:
Implemented Go too.

mpkorstanje:
I'll merge this by the next of next week unless there are any more comments. Then we can bundle it up for the next release of cucumber expressions.

Issue #657, opened 2019-07-18
Only suggest parameter types when text is surrounded by whitespace


aslakhellesoy:
Imagine the following parameter type is defined:

```ruby
ParameterType(
  name: 'direction',
  regexp: /(up|down)/,
  transformer: ->(s) { s }
)
```

This could be used with the following step definition:

```ruby
When("the sun goes {direction}") do |direction|
end
```

However, if we add two new undefined Gherkin steps:

```gherkin
Given the sun has gone down
When Aslak downloads a picture
```

That should suggest:

```ruby
Given("the sun has gone {direction}") do |direction|
end
```

But it should *not* suggest:

```ruby
When("Aslak {direction}loads a picture") do |direction|
end
```

Instead, it should simply suggest:

```ruby
When("Aslak downloads a picture") do |direction|
end
```

Parameters should only be suggested when the text is surrounded by whitespace, or when it's at the beginning or end of the step. That's what users would expect.

sebrose:
Perhaps an unintended consequence, but now numerical parameters with units are not recognised by the snippet generator:
Given Sean is 15m from Lucy -> "Sean is 15m from Lucy", but should be "Sean is {int}m from Lucy"

vincent-psarga:
Maybe it could be interesting to have some built-in "units" type (`\d+(d|h|m|s`, `\d+(m|km)`, `\d+('|")`) but I think it would be pretty hard to have something easily translatable/portable for multiple languages.

PR #661, opened 2019-07-23
cucumber-expressions: Fix suggestions with in-word parameters

vincent-psarga:
## Summary

Do not suggest in-word parameters for cucumber expressions. For example, we should not suggest:

```
 I use gem I{int}n
```

for the sentence:

```
I use gem i18n
```

See issue #657

## Details

I think we'll need discussions about this one as some examples used in the test show the tool working the other way around.
For example in Ruby, we have this test that clearly shows that we may want to suggest in-word parameters:

```ruby
        expression = @generator.generate_expressions("I reach Stage4: 1st flight-1st hotel")[0]
        # While you would expect this to be `I reach Stage{int}: {optional-flight}-{optional-hotel}`
        # the `-1` causes {int} to match just before {optional-hotel}.
        expect(expression.source).to eq("I reach Stage{int}: {optional-flight}{int}st hotel")
```

What I did so far (for compatibility):
 - when creating a custom parameter type, we have an option "force_full_word" which is true by default. So in-word parameters won't be suggested
 - default types (int, float etc) have the "force_full_word" set to false (so the tests don't fail and we have the same behavior).

Before applying the same changes to Go, Java etc, I'd like to be sure this approach is correct (or should I remove this flag and never propose in-word parameters ? )

## Checklist:


cucumber-expressions/ruby/spec/cucumber/cucumber_expressions/cucumber_expression_generator_spec.rb
```diff
@@ -162,6 +164,67 @@ class Currency
...
+            expect(@generator.generate_expression("When I watch the muppet show").source).not_to eq("When I watch the m{direction}pet show")
+            expect(@generator.generate_expression("When I watch the muppet show").source).to eq("When I watch the muppet show")
+          end
+
+          it "at the end of a word" do
+            expect(@generator.generate_expression("When I create a group").source).not_to eq("When I create a gro{direction}")
+            expect(@generator.generate_expression("When I create a group").source).to eq("When I create a group")
+          end
+
+          it "a full word" do
```
luke-hill: this feels a separate context but not of big importance?

vincent-psarga: I wanted to be sure not breaking anything there, but yes they are not that important checks indeed.

Maybe only keep the last one (which ensure that full words at the beginning and end of the sentence are matched)

cucumber-expressions/ruby/lib/cucumber/cucumber_expressions/parameter_type.rb
```diff
@@ -6,7 +6,7 @@ class ParameterType
       ILLEGAL_PARAMETER_NAME_PATTERN = /([\[\]()$.|?*+])/
       UNESCAPE_PATTERN = /(\\([\[$.|?*+\]]))/
 
-      attr_reader :name, :type, :regexps
+      attr_reader :name, :type, :regexps, :force_full_word
```
mpkorstanje: I hope I'm commenting on the right line.

I don't think this should be a flag. I would consider the old behavior to be just wrong so no need to enable it.

cucumber-expressions/ruby/lib/cucumber/cucumber_expressions/parameter_type_matcher.rb
```diff
@@ -3,26 +3,30 @@ module CucumberExpressions
...
+        @parameter_type, @regexp, @text, @force_full_word = parameter_type, regexp, text, force_full_word
         @match = @regexp.match(@text, match_position)
       end
 
       def advance_to(new_match_position)
         (new_match_position...@text.length).each {|advancedPos|
-          matcher = self.class.new(parameter_type, @regexp, @text, advancedPos)
-          if matcher.find
+          matcher = self.class.new(parameter_type, @regexp, @text, advancedPos, @force_full_word)
+          if matcher.find && matcher.full_word?
```
mpkorstanje: I would have expected this flag to be used in the Cucumber Expression Generator. What is the effect of using it here?

mpkorstanje: I get the impression this actually changes how the matching works.

mpkorstanje: Owh wait. I get it now. It's matcher. Not the parameter type.

mpkorstanje:
>  For example in Ruby, we have this test that clearly shows that we may want to suggest in-word parameters:

I would consider a `-` a valid word boundary so it'd match like this:

```
        expect(expression.source).to eq("I reach Stage4: {optional-flight}-{optional-hotel}")
```

mpkorstanje:
I think the following characters are valid boundaries ` `, `-`, `,`, `!`, `;`, `:`. They're all punctuation that normally follows a word.

mpkorstanje:
I think that with possibly a few exceptions every punctuation mark should count as a boundary. There are some Unicode classes we could use for that.

https://en.wikipedia.org/wiki/Punctuation

vincent-psarga:
> I think that with possibly a few exceptions every punctuation mark should count as a boundary. There are some Unicode classes we could use for that.
> 
> https://en.wikipedia.org/wiki/Punctuation

Definitely, that's something I didn't think about thanks. That should be taken into account now.

cucumber-expressions/ruby/lib/cucumber/cucumber_expressions/parameter_type.rb
vincent-psarga: I'm wondering about (which is the reason why so far only ruby is implemented).

I think it could be helpful for advanced users who may want to finely tweak the way suggestions are done.
But in the other hand those people would be mature enough to know how to tweak the step definitions and not blindly copy-paste what has been suggested by the tool.

Well, I'm a bit lost here :D

cucumber-expressions/ruby/spec/cucumber/cucumber_expressions/cucumber_expression_generator_spec.rb
luke-hill: I think I may have been mis-interpreted.

The checks are fine. But I would wrap them as thus

```
context 'when set to true, does suggest parameter when a match is'
```

cucumber-expressions/ruby/lib/cucumber/cucumber_expressions/parameter_type.rb
mpkorstanje: In Java this would mean yet another constructor variation. This makes the API more difficult to use because (when your IDE isn't perfectly) configured you'd have to guess between `....boolean, boolean)` and `...boolean, boolean, boolean)`. So I'd prefer not to have an extra one. 

Additionally the cucumber expression generator is already prone to producing combinatorial explosions. So anything that prunes the suggestions to include only the most likely ones is good.

cucumber-expressions/ruby/spec/cucumber/cucumber_expressions/cucumber_expression_generator_spec.rb
vincent-psarga: ahh ok I got it now sorry. Yes I can do that :)

cucumber-expressions/ruby/lib/cucumber/cucumber_expressions/parameter_type_matcher.rb
```diff
@@ -47,12 +47,12 @@ def <=>(other)
 
       def space_before_match_or_sentence_start
         match_begin = @match.begin(0)
-        match_begin == 0 || @text[match_begin - 1].match(/\s/)
+        match_begin == 0 || @text[match_begin - 1].match(/\s|[[:punct:]]/)
```
mpkorstanje: As far as I know `\s` only matches `[ \t\r\n\]`. However there are other spacing marks in use. In Hindi, Arabic and Pashto for example. (I don't have my reference examples around at the moment). So I think the unicode class `/\p{Zs}/ ` might be more appropriate here.

I believe `|[[:punct:]]` is a ruby specific implementation. To keep the implementation consistent between implementations I think it might be better to use Unicode character classes. All languages are able to understand these. For punctuation that would be `/\p{P}/`.

aslakhellesoy:
> I remember when I raised that original issue. The problem was originally an infinite loop which aslak did fix up a while back.
> 
> I think the usage was because often when you want an alternation in normal text you would right `1 or 2` as `1/2`, but this isn't possible using cuke-exp, because `/` is alternation. So I opted for `1-2` but this then got parsed as a very complex combination of integers.

@luke-hill I think the issue referring to is this: https://github.com/cucumber/cucumber/issues/307

cucumber-expressions/ruby/lib/cucumber/cucumber_expressions/parameter_type.rb
aslakhellesoy: I agree with @mpkorstanje - we should not make this a configurable flag. YAGNI, KISS etc.

cucumber-expressions/ruby/lib/cucumber/cucumber_expressions/parameter_type_matcher.rb
aslakhellesoy: For the Java implementation I went for `[\s\p{P}]` to determine if a character is whitespace or punctuation. I didn't look into Hindi/Arabic/Pashto spacing. I think we can defer supporting that until it becomes an issue. I prefer simplicity over completeness for this functionality - it's not execution, it's just snippet suggestions.

cucumber-expressions/java/src/test/java/io/cucumber/cucumberexpressions/CucumberExpressionGeneratorTest.java
```diff
@@ -234,43 +273,6 @@ public void exposes_transforms_in_generated_expression() {
...
-                true,
-                false
-        );
-
-        parameterTypeRegistry.defineParameterType(optionalFlight);
-        parameterTypeRegistry.defineParameterType(optionalHotel);
-        // While you would expect this to be `I reach Stage{int}: {optional-flight}-{optional-hotel}` the `-1` causes
-        // {int} to match just before {optional-hotel}.
-        List<GeneratedExpression> generatedExpressions = generator.generateExpressions("I reach Stage4: 1st flight-1st hotel");
-        assertEquals("I reach Stage{int}: {optional-flight}{int}st hotel", generatedExpressions.get(0).getSource());
```
mpkorstanje: I'm would not expect this to be deleted. This should now match `"I reach Stage{int}: {optional-flight}{optional-hotel}" because the `-` is punctuation.

cucumber-expressions/ruby/lib/cucumber/cucumber_expressions/parameter_type_matcher.rb
mpkorstanje: Good enough for now I guess. I'll add the extra character classes + tests when I get the time. It doesn't add much complexity. It's all in the Unicode standard.

cucumber-expressions/java/src/test/java/io/cucumber/cucumberexpressions/CucumberExpressionGeneratorTest.java
aslakhellesoy: You're right @mpkorstanje - deleting it was perhaps a bit sloppy of me. The thing is, I couldn't make the test verify the behaviour it was initially written to test, namely that the optional group had a 0-length match. We could keep it, tweak it and make it pass as you suggested, but it would be testing behaviour we're already testing I think...

cucumber-expressions/ruby/lib/cucumber/cucumber_expressions/parameter_type_matcher.rb
vincent-psarga: Thanks for pointing that out, the Ruby implementation now uses \p{P} instead of [[:punct]]

cucumber-expressions/java/src/test/java/io/cucumber/cucumberexpressions/CucumberExpressionGeneratorTest.java
vincent-psarga: ON the Ruby implementation, I've had to update this test so it still reflects the expected behaviour (matching the int before the optional capture group).
The sentence does not make that much sense now, hopefully the comment below still explain correctly why we don't match (optional-hotel)

cucumber-expressions/ruby/lib/cucumber/cucumber_expressions/parameter_type.rb
vincent-psarga: that's been removed.

cucumber-expressions/ruby/lib/cucumber/cucumber_expressions/parameter_type_matcher.rb
```diff
@@ -4,14 +4,14 @@ class ParameterTypeMatcher
       attr_reader :parameter_type
 
       def initialize(parameter_type, regexp, text, match_position=0)
-        @parameter_type, @regexp, @text = parameter_type, regexp, text
+        @parameter_type, @regexp, @text, @force_full_word = parameter_type, regexp, text
```
luke-hill: Think this can be removed

cucumber-expressions/ruby/lib/cucumber/cucumber_expressions/parameter_type_matcher.rb
```diff
@@ -38,6 +42,18 @@ def <=>(other)
         return length_comparison if length_comparison != 0
         0
       end
+
+      private
+
+      def space_before_match_or_sentence_start
```
luke-hill: Think this can be a predicate - so `space_before_match_or_sentence_start?`

cucumber-expressions/ruby/lib/cucumber/cucumber_expressions/parameter_type_matcher.rb
```diff
@@ -38,6 +42,18 @@ def <=>(other)
...
+        match_begin = @match.begin(0)
+        match_begin == 0 || @text[match_begin - 1].match(/\s|\p{P}/)
+      end
+
+      def space_after_match_or_sentence_end
```
luke-hill: This should probably end in `?` too.

cucumber-expressions/ruby/spec/cucumber/cucumber_expressions/cucumber_expression_generator_spec.rb
```diff
@@ -162,6 +162,58 @@ class Currency
         expect(expressions[1].source).to eq("{zero-or-more} {zero-or-more} {zero-or-more}")
       end
 
+      context "does not suggest parameter when match is" do
+        let!(:direction_parameter_type) {
```
luke-hill: not of huge importance, but do these need to be let bangs? It'll cause some issues for local runs potentially. It'll also annoy rubocop.

These aren't blockers. And if needs to remain, I'm :cool: with that. But if we can change them just to `let` that would be better

vincent-psarga: It's done :)

For the let! part, it was useful in the previous version to be able to use let(:force_full_word) and override it in sub-context but that does not make sense anymore indeed :)

mpkorstanje:
I wouldn't mind porting this one over to java and I'll review the go and javascript implementations afterwards.

aslakhellesoy:
> I wouldn't mind porting this one over to java and I'll review the go and javascript implementations afterwards.

@mpkorstanje - I've already ported this to Java (in this PR)

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/ParameterTypeRegistry.java
```diff
@@ -19,7 +19,7 @@
     // Pattern.compile(...).pattern() is not necessary, but it helps us take advantage of the IntelliJ's regexp validation,
     // which detects unneeded escapes.
     private static final List<String> INTEGER_REGEXPS = asList(Pattern.compile("-?\\d+").pattern(), Pattern.compile("\\d+").pattern());
-    private static final List<String> FLOAT_REGEXPS = singletonList(Pattern.compile("-?\\d*(?:[.,]\\d+)?").pattern());
+    private static final List<String> FLOAT_REGEXPS = singletonList(Pattern.compile("[+-]?(?:[0-9]*[.,])?[0-9]+").pattern());
```
mpkorstanje: @aslakhellesoy why did you change this? And why only for Java?

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/ParameterTypeMatcher.java
```diff
@@ -1,30 +1,46 @@
...
+                    char before = text.charAt(matcher.start() - 1);
+                    if (!isWhitespaceOrPunctuation(before)) {
+                        return false;
+                    }
+                }
+
+                if (matcher.end() < text.length()) {
+                    char after = text.charAt(matcher.end());
+                    return isWhitespaceOrPunctuation(after);
+                }
```
mpkorstanje: I find this section really hard to follow. Perhaps we extract a method here that makes it clear we should have a match and that the match should be a full word.

mpkorstanje: So something like 
```java
             if (!group().isEmpty() && matchIsFullWord()) { 
```

mpkorstanje:
Right. I even commented on it.

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/ParameterTypeRegistry.java
aslakhellesoy: When I fiddled with the test I removed I discovered that an undefined step def with `-` would suggest {float}. That's why I changed it. In retrospect I realise that should be done in a different PR, so I'll remove it from this one.

mpkorstanje:
I'd say this is ready to go. I'll merge it at the end of next week and bundle it up with the other cucumber expression mrs for a release.  Please do feel free to merge earlier though.

PR #668, opened 2019-08-01
[release process] Do not try to push old releases of packages

vincent-psarga:
## Summary

Just a few hours ago, when I tried to release cucumber-messages, I've ran into issues when running ``NEW_VERSION=3.0.1 make publish`` as the Make target was also trying to push old releases (the packages were still there).

This ensures that we just try to push the newest package.

Surprisingly, the old .nupkg should have been removed with make clean before, I don't know why it did not get deleted.

I also added the removal of the old gem in ruby/make clean


luke-hill:
I'll hold my hands up to that. I did try a couple of weeks ago to do a build, which half worked. For cucumber-expressions v8, and messages 2.1.2

c21e/ruby/default.mk
```diff
@@ -31,12 +31,12 @@ endif
...
-	gem push $$(find . -name "*.gem")
+	gem push $$(find . -name "*$(NEW_VERSION).gem")
 .PHONY: publish
 
 clean: clean-ruby
 .PHONY: clean
 
 clean-ruby:
-	rm -f .deps .linked .tested Gemfile.lock
+	rm -f .deps .linked .tested Gemfile.lock .gem
```
aslakhellesoy: Did you mean `*.gem`?

aslakhellesoy: If so, should this be done in the `.template/ruby/default.mk` file instead?

aslakhellesoy: Never mind the last comment, I see you added it to `.template/ruby/default.mk`, but I still think it should be `*.gem`

PR #672, opened 2019-08-02
cucumber-expressions: Improve decimal number parsing

### LINKED ISSUE #669: cucumber-expressions: `-` sign withouth numbers matches as `{double}`
This in `cucumber-jvm` now fails with the latest `cucumber-expressions` snapshot:

```java
    @Test
    public void generatesCopyPasteReadySnippetWhenStepHasIllegalJavaIdentifierChars() {
        String expected = "" +
            "@Given(\"I have {int} cukes in: my {string} red-belly!\")\n" +
            "public void i_have_cukes_in_my_red_belly(Integer int1, String string) {\n" +
            "    // Write code here that turns the phrase above into concrete actions\n" +
            "    throw new io.cucumber.java.PendingException();\n" +
            "}\n";
        assertEquals(expected, snippetFor("I have 4 cukes in: my \"big\" red-belly!"));
    }
```

With the snippet being:

```
@Given("I have {int} cukes in: my {string} red{double}belly!")
public void i_have_cukes_in_my_red_belly(Integer int1, String string, Double double1) {
    // Write code here that turns the phrase above into concrete actions
    throw new io.cucumber.java.PendingException();
}
```

Looks like the `-` is matched as a `double`.

mpkorstanje:
## Summary

Properly match decimal numbers taking into account that a decimal has:
 * at least one number
 * optional negative/positive sign
 * a number preceding/succeeding the thousands separator
 * may have multiple thousands separators
 * may have a single decimal separator
 * may start with a decimal separator
 * an optional fractional part
 * an optional scientific exponent notation
 * localized decimal and grouping characters. See: [International Language Environments Guide - Decimal and Thousands Separators](https://docs.oracle.com/cd/E19455-01/806-0169/overview-9/index.html)

## Details

Required partially re implementing https://github.com/cucumber/cucumber/pull/644 

## Motivation and Context

Looks like https://github.com/cucumber/cucumber/pull/605 made the expression a bit too eager. Fixes #669.

## Types of changes

## Checklist:

### DIFF
--- a/cucumber-expressions/java/pom.xml
+++ b/cucumber-expressions/java/pom.xml
@@ -36,7 +36,7 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <version>4.13-beta-3</version>
+            <version>4.12</version>
             <scope>test</scope>
         </dependency>
 
@@ -46,11 +46,20 @@
             <version>2.8.5</version>
             <scope>test</scope>
         </dependency>
-
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-core</artifactId>
+            <version>2.1</version>
+        </dependency>
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-library</artifactId>
+            <version>2.1</version>
+        </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
-            <version>2.9.9.1</version>
+            <version>2.9.9.2</version>
             <scope>test</scope>
         </dependency>
 
--- a/cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/ParameterTypeRegistry.java
+++ b/cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/ParameterTypeRegistry.java
@@ -4,6 +4,7 @@ import org.apiguardian.api.API;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.text.DecimalFormatSymbols;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
@@ -21,10 +22,24 @@ import static java.util.Collections.singletonList;
 public final class ParameterTypeRegistry {
     // Pattern.compile(...).pattern() is not necessary, but it helps us take advantage of the IntelliJ's regexp validation,
     // which detects unneeded escapes.
-    private static final List<String> INTEGER_REGEXPS = asList(Pattern.compile("-?\\d+").pattern(), Pattern.compile("\\d+").pattern());
-    private static final List<String> FLOAT_REGEXPS = singletonList(Pattern.compile("-?\\d*(?:[.,]\\d+)?").pattern());
-    private static final List<String> WORD_REGEXPS = singletonList(Pattern.compile("[^\\s]+").pattern());
-    private static final List<String> STRING_REGEXPS = singletonList(Pattern.compile("\"([^\"\\\\]*(\\\\.[^\"\\\\]*)*)\"|'([^'\\\\]*(\\\\.[^'\\\\]*)*)'").pattern());
+    private static final List<String> INTEGER_REGEXPS = asList(
+            Pattern.compile("-?\\d+").pattern(),
+            Pattern.compile("\\d+").pattern()
+    );
+    private static final String SIGN = "[-+]?";
+    private static final String MUST_CONTAIN_NUMBER = "(?=.*[0-9].*)";
+    private static final String SCIENTIFIC_NUMBER = "(?:[0-9]+[{expnt}][-+]?[0-9]+)?";
+    private static final String DECIMAL_FRACTION = "(?:[{decimal}](?=[0-9].*))?[0-9]*";
+    private static final String INTEGER = "(?:[0-9]+(?:[{group}]?[0-9]+)*)*";
+    private static final String FLOAT_REGEXPS =
+            Pattern.compile(MUST_CONTAIN_NUMBER + SIGN + INTEGER + DECIMAL_FRACTION + SCIENTIFIC_NUMBER).pattern();
+    private static final List<String> WORD_REGEXPS = singletonList(
+            Pattern.compile("[^\\s]+").pattern()
+    );
+    private static final List<String> STRING_REGEXPS = asList(
+            Pattern.compile("\"([^\"\\\\]*(\\\\.[^\"\\\\]*)*)\"").pattern(),
+            Pattern.compile("'([^'\\\\]*(\\\\.[^'\\\\]*)*)'").pattern()
+    );
     private static final String ANONYMOUS_REGEX = Pattern.compile(".*").pattern();
     private final Map<String, ParameterType<?>> parameterTypeByName = new HashMap<>();
     private final Map<String, SortedSet<ParameterType<?>>> parameterTypesByRegexp = new HashMap<>();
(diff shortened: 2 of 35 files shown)


mpkorstanje:
@luke-hill would you like to do the Ruby on this one?

cucumber-expressions/java/pom.xml
```diff
@@ -46,11 +47,20 @@
             <version>2.8.5</version>
             <scope>test</scope>
         </dependency>
-
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-core</artifactId>
+            <version>${hamcrest.version}</version>
```
aslakhellesoy: Could you change this to use a literal value rather than a variable? The `update-dependencies` make target only works when the versions are literal:

https://github.com/cucumber/cucumber/blob/0e63c88e101b29da3f69d95b395197a486d25963/cucumber-expressions/java/default.mk#L14

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/ParameterTypeRegistry.java
```diff
@@ -4,13 +4,8 @@
...
 import java.math.BigInteger;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.SortedSet;
-import java.util.TreeSet;
+import java.text.DecimalFormatSymbols;
+import java.util.*;
```
aslakhellesoy: I've configured IDEA to not use star imports. Could you do the same?

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/ParameterTypeRegistry.java
```diff
@@ -21,10 +16,24 @@
...
-    private static final List<String> FLOAT_REGEXPS = singletonList(Pattern.compile("-?\\d*(?:[.,]\\d+)?").pattern());
-    private static final List<String> WORD_REGEXPS = singletonList(Pattern.compile("[^\\s]+").pattern());
-    private static final List<String> STRING_REGEXPS = singletonList(Pattern.compile("\"([^\"\\\\]*(\\\\.[^\"\\\\]*)*)\"|'([^'\\\\]*(\\\\.[^'\\\\]*)*)'").pattern());
+    private static final List<String> INTEGER_REGEXPS = asList(
+            Pattern.compile("-?\\d+").pattern(),
+            Pattern.compile("\\d+").pattern()
+    );
+    private static final String SIGN = "[-+]?";
+    private static final String MUST_CONTAIN_NUMBER = "(?=.*[0-9].*)";
+    private static final String SCIENTIFIC_NUMBER = "(?:[0-9]+[{expnt}][-+]?[0-9]+)?";
```
aslakhellesoy: Is `{expnt}` standard regexp that is easily portable to other languages? If not, I don't think we should use it, as it will lead to inconsistent behaviour.

aslakhellesoy:
I would like to keep the option open to use a non-java runtime in the future for matching cucumber expressions. If we're relying on Java-only features now, we may not be able to do this in the future. So while this is an improvement, I'm worried that it will lock us in.

cucumber-expressions/java/pom.xml
mpkorstanje: Sure.

Though I'm not a fan of automatically updating dependencies. Maven does not understand the concept of pre-releases so will happily jump from 1.0.0 to 2.0.0-M1. This is usually not desirable as these releases can be quite buggy.

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/ParameterTypeRegistry.java
mpkorstanje: I'll add an enforcer rule in the pom.

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/ParameterTypeRegistry.java
mpkorstanje: It is a template variable so we can localize the the regular expression. It is substituted out a few lines down for the localized value. It just has the weird spelling so there are no repeat letters which trips up IDEA's regex inspection.

aslakhellesoy: Aha! I missed that. In that case - please disregard my other comments. Very cool!

mpkorstanje:
This feature uses the same system a user would to define custom parameter types -- they are provided by cucumber only because it is convenient. So unless you intend to remove that ability lock-in won't be a problem.

cucumber-expressions/java/pom.xml
mpkorstanje: > The update-dependencies make target only works when the versions are literal

You can use `versions:update-properties` for this:

http://www.mojohaus.org/versions-maven-plugin/update-properties-mojo.html

mpkorstanje:
I'll try to merge this after #644 is merged. It will probably fail due to conflicts.

luke-hill:
For my own benefit, is `\d` not as usable as `[0-9]` and also, the regex has now got a lot more complex. Didn't we have pretty much all use cases covered before (I've not been through rigorously, but does it need to be this complex?)

mpkorstanje:
Unfortunatly so. It has to match numbers with and without fraction, with and without a interger part, positive and negative. But at the same time it should not match empty, only the a minus sign or a comma.

The scientific notation and seperator matching was easy to add in and is usefull.

luke-hill:
Oki, that kinda doesn't answer my question though. Like doesn't `\d` mean the same as `[0-9]` I was always under the impression it did.

The scientific notation I can see the use for. But I'm pretty sure the previous stuff caught all the ones you mentioned.

mpkorstanje:
Ah. You want to replace the `[0-9]` with `\d`. That is reasonable.

Initially the pattern worked fine but it also matched a single minus sign. However fixing that allowed the empty string to match. Fixing that ... Fixing that ... So I eventually started from scratch with the list of requirements I wrote on top of this ticket.

These are all the tests I wrote for the pattern. But I see I didn't duplicate them to the other languages. The first group is stuff that shouldn't be matched. 

```java
        assertThat(expression.match(""), nullValue());
        assertThat(expression.match("."), nullValue());
        assertThat(expression.match(","), nullValue());
        assertThat(expression.match("-"), nullValue());
        assertThat(expression.match("E"), nullValue());
        assertThat(expression.match("1,"), nullValue());
        assertThat(expression.match(",1"), nullValue());
        assertThat(expression.match("1."), nullValue());
        assertThat(expression.match("E1"), nullValue());

        assertThat(expression.match("1").get(0).getValue(), is(BigDecimal.ONE));
        assertThat(expression.match("-1").get(0).getValue(), is(new BigDecimal("-1")));
        assertThat(expression.match("1.1").get(0).getValue(), is(new BigDecimal("1.1")));
        assertThat(expression.match("1,000").get(0).getValue(), is(new BigDecimal("1000")));
        assertThat(expression.match("1,000,0").get(0).getValue(), is(new BigDecimal("10000")));
        assertThat(expression.match("1,000.1").get(0).getValue(), is(new BigDecimal("1000.1")));
        assertThat(expression.match("1,000,10").get(0).getValue(), is(new BigDecimal("100010")));
        assertThat(expression.match("1,0.1").get(0).getValue(), is(new BigDecimal("10.1")));
        assertThat(expression.match("1,000,000.1").get(0).getValue(), is(new BigDecimal("1000000.1")));
        assertThat(expression.match("-1.1").get(0).getValue(), is(new BigDecimal("-1.1")));

        assertThat(expression.match(".1").get(0).getValue(), is(new BigDecimal("0.1")));
        assertThat(expression.match("-.1").get(0).getValue(), is(new BigDecimal("-0.1")));
        assertThat(expression.match("-.10000001").get(0).getValue(), is(new BigDecimal("-0.10000001")));
        assertThat(expression.match("1E1").get(0).getValue(), is(new BigDecimal("1E1"))); // precision 1 with scale -1, can not be expressed as a decimal
        assertThat(expression.match(".1E1").get(0).getValue(), is(new BigDecimal("1")));
        assertThat(expression.match("-.1E-1").get(0).getValue(), is(new BigDecimal("-0.01")));
        assertThat(expression.match("-.1E+1").get(0).getValue(), is(new BigDecimal("-0.1")));
        assertThat(expression.match("-.1E1").get(0).getValue(), is(new BigDecimal("-1")));
```

PR #673, opened 2019-08-04
Take advantage of Java8 language features. Misc cleanup.

aslakhellesoy:
## Summary

Use Java8 language features

## Details

We've recently upgraded to Java8 (by bumping `cucumber-parent`)

## Motivation and Context

The code is more terse

## Types of changes

## Checklist:


cucumber-expressions/java/pom.xml
```diff
@@ -36,7 +36,7 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <version>4.12</version>
+            <version>4.13-beta-3</version>
```
mpkorstanje: We could use JUnit Jupiter instead. I think that is preferable to using a beta version.

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/TreeRegexp.java
```diff
@@ -74,38 +75,10 @@ Pattern pattern() {
...
         if (!matcher.matches()) return null;
-        return groupBuilder.build(matcher, new IntRange(0, matcher.groupCount() + 1));
+        return groupBuilder.build(matcher, IntStream.rangeClosed(0, matcher.groupCount()).iterator());
     }
 
     public GroupBuilder getGroupBuilder() {
         return groupBuilder;
     }
-
-    private static class IntRange implements Iterator<Integer> {
```
mpkorstanje: Good riddance.

cucumber-expressions/java/pom.xml
aslakhellesoy: Eh JUnit Jupiter - is that a new API I have to learn?

aslakhellesoy: I don't think there is much to risk using a beta version for a test dependency

mpkorstanje: > Eh JUnit Jupiter - is that a new API I have to learn?

Also known as JUnit 5. For the sort of tests we are using, you'd have to use some new imports, that is it.

I'll update it to Junit 5 on the next pass through. It's not so difficult.

Issue #663, opened 2019-07-27
Disable go mod replace directive for releases


aslakhellesoy:
When we build a release of a go module, all declared dependencies should be versioned.

To ensure this, all `replace` directives should be commented out during the build, and uncommented as part of the post-release process.

For example, when we build and tag a release of gherkin-go, we should *not* build against cucumber-messages-go using the relative path, but rather the tagged release of cucumber-messages-go.

This poses a new challenge. Go modules requires tags to be of the format `vMAJOR.MINOR.PATCH`. However, for the monorepo we have adopted a git tag naming convention of `COMPONENT/vMAJOR.MINOR.PATCH` to avoid collisions of tags between components.

I think we have 2 options here:

a) Adopt a global version number for all components in the monorepo
b) Use [splitsh-lite](https://github.com/splitsh/lite) to sync to read-only "sub" repos

We have been using splitsh-lite in the past (before we migrated from TravisCI to CircleCI), so that would be relatively easy. We should still consider option a) as I think it might simplify other aspects of the build.

This is somewhat related to #662

aslakhellesoy:
Fixed

PR #680, opened 2019-08-13
Subrepo sync

aslakhellesoy:
## Summary

Restore subrepo sync with splitsh. We used to have this before the CircleCI migration.

## Motivation and Context

Go modules must be versioned using `vX.Y.Z` git tags (semver). The tagging scheme used in the monorepo is incompatible with go modules. Also see #663.


aslakhellesoy:
This PR also removes some unused bash functions

PR #684, opened 2019-08-21
HTML formatter

aslakhellesoy:
## Summary

New cross-platform HTML formatter, based on cucumber-messages and React

## Details

This PR implements a command-line tool `cucumber-html-formatter` which reads binary protobuf messages from `STDIN` and writes a HTML report to `STDOUT`.

Cucumber implementations will need a `protobuf` formatter that writes messages to a file. A build script can then generate a HTML report:

    cat messages.bin | cucumber-html-formatter > report.html

The report is 100% standalone and doesn't require any additional CSS, JavaScript or JSON files. It's a static SPA that allows users to navigate between documents.

Currently there is no support for attachments (screenshots) - that can be added later. It may not be practical to embed attachments in the HTML (because they tend to be too big), so in future the formatter may require an `--attachments DIR` option as well as an optional `--attachment-base-url` option to override the `` links in the generated HTML. (This will be useful if the report and assets will be hosted on a HTTP server).

The formatter currently lives inside the `cucumber-react` package, and will be moved out to a `html-formatter` package later. This is because we want to be able to use the `cucumber-react` component without the baggage of the command-line formatter in other web-based tools.

## Motivation and Context

There are two motivations for implementing this.

The first one is that the built-in Ruby and Java HTML formatters are severely outdated and difficult to maintain. Cucumber.js doesn't even have one. This one will work for all implementations (as soon as they implement the `protobuf` formatter).

The other motivation is to gauge the community's appetite for installing a separate executable in addition to a "native" Cucumber installed via Rubygems/Maven/Grandle/NPM/NuGet etc. If a big enough part of the community is willing to do this, we may move more parts of Cucumber into a share runtime such as [cucumber-engine](https://github.com/cucumber/cucumber-engine).

## How Has This Been Tested?

With some automated tests, but also visually, using Storybook

Screenshot 2019-08-21 at 10 55 58

## Types of changes

## Checklist:


mpkorstanje:
Looks pretty. Does it escape HTML?

aslakhellesoy:
> Looks pretty. 
Thanks, but I haven't really spent much effort styling it yet

> Does it escape HTML?
Yes. This is handled by React.

mattwynne:
Great, Aslak! I have lots o lots of ideas and suggestions but nothing that can't wait until this is merged and we can do a second iteration.

PR #694, opened 2019-09-01
gherkin/perl: Update perl implementation

ehuelsmann:
## Summary

This PR updates the Gherkin Perl implementation to Gherkin v7 by updating the generated Perl modules and adjusting the hand-coded modules accordingly.

## Details

Because since the last time the Perl implementation was updated, `Scenario` and `Scenario Outline` have semantically been merged, code changes were required to adapt the hand-crafted code (the token matcher) to behave the same way.

## Motivation and Context

Ideally, Perl's Cucumber implementation (Test::BDD::Cucumber -- of which I'm the primary maintainer) moves to using the official Cucumber parser (based on the Berp grammar). However, in its current state, that's impossible because this implementation is too far behind on Gherkin releases.

## How Has This Been Tested?

There currently are hardly any tests for this implementation. I wrote a test locally to test this change against the `testdata/good/` body of test scenarios. Obviously a number of additional tests need to be developed to introduce sufficient assurance for this implementation. I was planning to do that in a separate PR: it's impossible to build passing tests for the code as it is on `master` now (using the `testdata/` body of tests).

Building a body of tests however is independent of this change: this change is about updating the implementation, not about building an extensive test suite.

Let me know if adding the creation of a test suite should be included into this PR.

## Types of changes

No idea; it's an update to the implementation...

## Checklist:

None of the above apply: It's an update to the existing Perl implementation (no updates/ports required). No documentation needs to be added (beyond the obvious lack of documentation of the module in general). No additional tests are required beyond what should have been in place (but already isn't).

Let me know how to proceed!

### DIFF
--- a/gherkin/Makefile
+++ b/gherkin/Makefile
@@ -1,2 +1,2 @@
-MAKEFILES = go/Makefile javascript/Makefile ruby/Makefile c/Makefile java/Makefile
+MAKEFILES = go/Makefile javascript/Makefile ruby/Makefile c/Makefile java/Makefile perl/Makefile
 include default.mk
--- a/gherkin/perl/bin/gherkin-generate-ast
+++ b/gherkin/perl/bin/gherkin-generate-ast
@@ -12,18 +12,13 @@ package App::GherkinGenerateAst;
 sub run {
     my ($class, $fh, @file_list) = @_;
 
-    my $json = Cpanel::JSON::XS->new->utf8->canonical;
+    my $json = Cpanel::JSON::XS->new->utf8->convert_blessed->canonical->pretty;
 
     my $parser = Gherkin::Parser->new();
 
     for my $file (@file_list) {
         my $gherkin_document = $parser->parse($file);
-        my $event = {
-            type => 'gherkin-document',
-            uri => $file,
-            document => $gherkin_document
-        };
-        print $fh $json->encode( $event );
+        print $fh $json->encode( $gherkin_document );
         print "\n"
     }
 
--- a/gherkin/perl/bin/gherkin-generate-pickles
+++ b/gherkin/perl/bin/gherkin-generate-pickles
@@ -1,15 +1,16 @@
 #!/usr/bin/env perl
 
+package App::GherkinGeneratePickles;
+
 use strict;
 use warnings;
 use lib 'lib';
 
 use Cpanel::JSON::XS;
+use Path::Class;
 
 use Gherkin::Parser;
 use Gherkin::Pickles::Compiler;
-
-package App::GherkinGeneratePickles;
 
 sub run {
     my ( $class, $fh, @file_list ) = @_;
@@ -18,16 +19,12 @@ sub run {
     my $parser = Gherkin::Parser->new();
 
     for my $file (@file_list) {
+        my $source = file($file)->slurp;
         my $pickles = Gherkin::Pickles::Compiler->compile(
-            $parser->parse($file)
+            $parser->parse($file), $source
         );
         for my $pickle (@$pickles) {
-            my $event = {
-              type => 'pickle',
-              uri => $file,
-              pickle => $pickle
-            };
-            print $fh $json->encode($event);
+            print $fh $json->encode($pickle);
             print "\n"
         }
     }
--- a/gherkin/perl/bin/gherkin-generate-pickles
+++ b/gherkin/perl/bin/gherkin-generate-pickles
@@ -14,7 +14,7 @@ use Gherkin::Pickles::Compiler;
 
 sub run {
     my ( $class, $fh, @file_list ) = @_;
-    my $json = Cpanel::JSON::XS->new->utf8->canonical;
+    my $json = Cpanel::JSON::XS->new->utf8->pretty->canonical;
 
     my $parser = Gherkin::Parser->new();
 
@@ -25,7 +25,6 @@ sub run {
         );
         for my $pickle (@$pickles) {
             print $fh $json->encode($pickle);
-            print "\n"
         }
     }
 }
--- a/gherkin/perl/bin/gherkin-generate-tokens
+++ b/gherkin/perl/bin/gherkin-generate-tokens
@@ -3,6 +3,7 @@
 use strict;
 use warnings;
 use lib 'lib';
+use utf8;
 
 use Gherkin::Parser;
 use Gherkin::TokenFormatterBuilder;
@@ -20,4 +21,5 @@ sub run {
 
 }
 
+binmode STDOUT, ':utf8';
 __PACKAGE__->run(\*STDOUT, @ARGV) unless caller;
--- a/gherkin/perl/gherkin-perl.razor
+++ b/gherkin/perl/gherkin-perl.razor
@@ -60,6 +60,8 @@ our %states_to_match_names = (
 
 sub parse {
     my ( $self, $token_scanner, $token_matcher ) = @@_;
+    my $additional = {};
+    $additional->{'uri'} = $token_scanner unless ref $token_scanner;
 
     $token_matcher ||= Gherkin::TokenMatcher->new();
     $token_scanner = Gherkin::TokenScanner->new($token_scanner)
@@ -87,7 +89,7 @@ sub parse {
         last if $token->is_eof();
     }
 
-    $self->_end_rule( $context, '@Model.RuleSet.StartRule.Name' );
+    $self->_end_rule( $context, '@Model.RuleSet.StartRule.Name', $additional );
 
     if ( my @@errors = $context->errors ) {
         Gherkin::Exceptions::CompositeParser->throw(@@errors);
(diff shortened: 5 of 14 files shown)


aslakhellesoy:
Hi @ehuelsmann - thanks for tasking the time to update this!

There is a common acceptance test suite, which is based on `make`, `diff` and a bunch of "golden master" expected results.

In order to verify that the Perl implementation passes those tests, they have to be re-enabled in [gherkin/Makefile](https://github.com/cucumber/cucumber/blob/master/gherkin/Makefile).

Could you do that please?

Cheers,
Aslak

ehuelsmann:
@aslakhellesoy sure! No problem.

However, running the tests manually, I'm having a UTF-8 encoding issue:

```diff
--- testdata/good/escaped_pipes.feature.tokens	2019-08-29 12:31:40.522288882 +0200
+++ acceptance/testdata/good/escaped_pipes.feature.tokens	2019-09-01 22:44:59.348372922 +0200
@@ -4,11 +4,11 @@
 (4:1)Other://
 (5:3)ScenarioLine:Scenario/They are the future/
 (6:5)StepLine:Given /they have arrived/
-(7:7)TableRow://9:æ,13:o
-(8:7)TableRow://9:a,13:ø
+(7:7)TableRow://9:�,13:o
+(8:7)TableRow://9:a,13:�
 (9:5)StepLine:Given /they have arrived/
-(10:7)TableRow://9:|æ\n,22:\o
+(10:7)TableRow://9:|�\n,22:\o
 o\
-(11:7)TableRow://9:\|a\\n,22:ø\
-ø\
+(11:7)TableRow://9:\|a\\n,22:�\
+�\
 EOF
```

Any hints on how to solve this?

ehuelsmann:
Actually, I found the solution. I needed to make sure `gherkin-generate-tokens` return UTF-8 its STDOUT.

ehuelsmann:
If you're ok with it, I'll enable Perl in the central makefile once I have all tests fixed locally. I'm right now ready to push fixed ASTs. This PR already has the tokens.

However, there's a bit of work to be done on the pickles.

aslakhellesoy:
Sure, sounds good to me, but don't worry if the PR build is failing. We won't merge to master until it's all green anyway.

I usually recommend [draft PRs](https://github.blog/2019-02-14-introducing-draft-pull-requests/) for bigger chunks of work like this one.

aslakhellesoy:
Another thing that will need updating is the build scripts so that the core team will be able to publish Perl releases. The release process had changed quite a bit in the past few months. You can find more information about this in https://github.com/cucumber/cucumber/blob/master/RELEASE_PROCESS.md

We use a `cukebot` "robot" account with the various package managers, so we'd have to set one up on CPAN or Pause or whatever it is, and also get publish access to the existing one: https://metacpan.org/pod/Gherkin

ehuelsmann:
> I usually recommend draft PRs for bigger chunks of work like this one.

Actually, I had in mind to do this in smaller steps. However, with the smaller steps, it's tough to show you it's a step in the right direction, because the tests still fail. That's when it kind-of happened that I continued to implement the rest of the required changes in this single PR. There's a number of commits to monitor what I've actually done, for those interested.

ehuelsmann:
> We use a cukebot "robot" account with the various package managers, so we'd have to set one up on CPAN or Pause or whatever it is, and also get publish access to the existing one: https://metacpan.org/pod/Gherkin

No problem. I'm in contact with Peter (although he's extremely busy these days). I'll get the correct access rights assigned and set up once I fix the rest of the tests.

ehuelsmann:
@aslakhellesoy I think I have resolved most issues, but I can't get the tests to succeed because of 1 thing: I know the pickle->id value is a sha1sum over something (I got that from the C sources); however:

```diff
diff --unified <(jq "." testdata/good/docstrings.feature.pickles.ndjson) <(jq "." acceptance/testdata/good/docstrings.feature.pickles.ndjson)
--- /dev/fd/63	2019-09-02 00:27:53.635257411 +0200
+++ /dev/fd/62	2019-09-02 00:27:53.635257411 +0200
@@ -1,6 +1,6 @@
 {
   "pickle": {
-    "id": "7f6957a28baf5ecf2af10c7ebcbb2febb1719be4",
+    "id": "8d5a672f16eb3f03761a6c13d17a5dcd068eb189",
     "language": "en",
     "locations": [
       {
```

but:
```bash
$ sha1sum testdata/good/docstrings.feature
8d5a672f16eb3f03761a6c13d17a5dcd068eb189  testdata/good/docstrings.feature
$ sha1sum -b testdata/good/docstrings.feature
8d5a672f16eb3f03761a6c13d17a5dcd068eb189 *testdata/good/docstrings.feature
```

Concluding: I'm getting the exact same sha1 from `sha1sum` as from the latest state of my source code. However, that's not the expected sha1sum...

Could you please tell me what it is that's the source for the sha1? Thanks!

ehuelsmann:
Hmm, actually, I have a bit more work on the pickles, but that should all be "just work". For the pickle IDs, I lack information. (But maybe I missed the docs?)

ehuelsmann:
@aslakhellesoy This PR now has complete compliance with your test suite (token, ast and pickles) **_except_** for the pickle `id` field...

ehuelsmann:
The build currently fails because `cpanm` fails from the build image. It's Perl's de-facto dependency installer. Could you install that on the images? (I'd rather not include commands in the Makefile to do it, because it doesn't apply outside the build environment...)

aslakhellesoy:
I had a once-over and noticed gherkin/perl doesn't use protocol buffers. Does that mean it can only emit messages as JSON (as required by the tests), but not binary protobuf?

How hard would it be to use https://metacpan.org/pod/Google::ProtocolBuffers?

I wonder if we should start using `.proto.bin` instead of `.ndjson`. The only benefit I see with `.ndjson` is that the diff is human-readable. We could print the ndjson diff when the binary diff fails.

WDYT?

ehuelsmann:
@aslakhellesoy The reason the current implementation doesn't output protobuf is because it wasn't there before and I wasn't aware that other implementations did. I don't have a problem supporting protobuf output. In fact, I immediately checked whether Google::ProtocolBuffers can work with Cucumber's `messages.proto` tonight. Unfortunately, it doesn't understand the input.

What about we use this PR to update what's currently there to be 7.0.x compliant and I create separate PRs for what should be there? In order to create those PRs, I have studied several other implementations, which raised the following questions:
 - does an implementation need `bin/gherkin-generate-*` or `bin/gherkin` (with `--with-ast`/`--no-ast` etc options)?
 - does an implementation need to generate protobuf output? (or is JSON the minimum requirement?)
 - does an implementation need other formatters than protobuf or JSON?

Maybe the underlying question is: which language subdirectory should I be looking at to see a normative language implementation?

ehuelsmann:
> I wonder if we should start using `.proto.bin` instead of `.ndjson`. The only benefit I see with `.ndjson` is that the diff is human-readable. We could print the ndjson diff when the binary diff fails.
> 
> WDYT?

In general, I think this is a very good idea, assuming that ProtoBuf output is the norm these days. It'll help bring that requirement to the attention of authors of various language implementations.

ehuelsmann:
Just finished checking https://metacpan.org/pod/Google::ProtocolBuffers::Dynamic ; however, that doesn't support the protobuf well known types, which `messages.proto` does use (Timestamp). So that's not a short term solution either.

aslakhellesoy:
That `Timestamp` type has caused nothing but headaches! Just look at the hack I had to add to the `Dockerfile`.

I think we'd be fine if we changed all timestamp fields to use `int64` - see https://stackoverflow.com/questions/3574716/date-and-time-type-for-use-with-protobuf#3627777

Would that simplify the use of https://metacpan.org/pod/Google::ProtocolBuffers::Dynamic @ehuelsmann?

> It'll help bring that requirement to the attention of authors of various language implementations.

Exactly my thought - I realised it must not have been apparent to you at all that we were using protobuf.

ehuelsmann:
Yes, that works. However, what also works is adding

```protobuf
message Timestamp {
  int64 seconds = 1;
  int32 nanos = 2;
}
```

to the end of the `messages.proto` file and removing the timestamp include.

Basically, this *is* the definition of that type; depending on some external source (causing problems at that) for these 4 lines seems... unnecessary.

ehuelsmann:
> We use a cukebot "robot" account with the various package managers, so we'd have to set one up on CPAN or Pause or whatever it is, and also get publish access to the existing one: https://metacpan.org/pod/Gherkin

After I contacted @pjlsergeant about the access rights, it turns out the primary rights are on the account HELLESOY , presumably the same person as @aslakhellesoy  :-) Could you revive the account and provide me upload rights to at least upload the updated version in its current version (which is already a *lot* better than what's there); further updates can then be uploaded when we succeed to add the protobufs and maybe other improvements (tag expressions, cucumber expressions, etc)

ehuelsmann:
(If you need help getting upload rights, we could set up a slack chat or a phone call)

aslakhellesoy:
I have requested a new account for `cukebot` at https://pause.perl.org/pause/query?ACTION=request_id.

The credentials for this account will be in 1Password once it's approved. To get access to that, please create a free 1Password account and send me your email address in a private message in Slack, @ehuelsmann.

aslakhellesoy:
I'm closing this PR in favour of #711. It has all your commits and some additional CI-related ones from me. Let's continue over there.

PR #696, opened 2019-09-04
Standalone json formatter

aslakhellesoy:
A standalone executable `cucumber-json-formatter` that will produce output similar to the legacy `--format json` formatters that are built-into cucumber-{ruby,js,java}. This one reads cucumber messages from STDIN.

The purpose of this formatter is to produce identical output to the various built-in JSON formatters in cucumber-ruby/cucumber-js/cucumber-jvm/specflow.

These implementations should remove their native JSON formatter implementations when they are capable of producing cucumber-messages output.


json-formatter/go/json.go
```diff
@@ -143,7 +143,7 @@ func (formatter *Formatter) ProcessMessages(stdin io.Reader, stdout io.Writer) (
 					Name:    pickleStep.Text,
 					// The match field defaults to the feature itself for some curious reason
 					Match: &jsonStepMatch{
-						Location: fmt.Sprintf("%s:%d", pickle.Uri, step.Location.Line),
+						Location: fmt.Sprintf("%s:%d", pickle.Uri, pickleStep.Locations[len(pickleStep.Locations)-1].Line),
```
mpkorstanje: This is a pattern that keeps showing up. Perhaps it'd be useful to have separate properties. It will also allow us to more accurately name the locations. For instance `pickleLocation` and `scenarioLocation`. If they're equal the scenario isn't a scenario outline.

aslakhellesoy: Yeah I suppose we could do that. I chose to model it as a stack for a couple of reasons:

- I don't know a priori how many levels of stack we may need with other formats
- A stack lends itself well to printing stack traces

I'm not too bothered by the way things are now.

PR #706, opened 2019-09-23
Use our own timestamp and duration message

vincent-psarga:
## Summary

Do not rely on Google's Timestamp message but user own format.
The duration will be expression with the same message (although named Duration for clarity)

## Types of changes

Not rely applying. Not a bug fix, neither a new feature. The duration field __might__ break previous version.

## Checklist:


cucumber-messages/messages.proto
```diff
@@ -433,31 +458,32 @@ message TestStepMatched {
...
   TestResult testResult = 2;
-  google.protobuf.Timestamp timestamp = 3;
+  Timestamp timestamp = 3;
 }
 
 message TestResult {
   Status status = 1;
   string message = 2;
-  uint64 durationNanoseconds = 3;
+  uint64 durationNanoseconds = 3 [deprecated=true];
```
aslakhellesoy: Up until now we have not used protobuf's deprecated feature. Instead we are simply replacing the field with `reserved`. See the comment on line 7.

Once we have stuff in production we can consider `deprecated`, but for now I think it is best to remove it (more thorough cleanup, and no risks).

aslakhellesoy:
I think we should include an explnanation about _why_ we are replacing the stdlib messages with our own.

cucumber-messages/messages.proto
vincent-psarga: Works for me. I wasn't sure about Specflow handling of messages, but I think we can discuss it in the coming meeting with them.
I thought the deprecation would make it easier on their side :)

aslakhellesoy: I think deprecation will only make it easier to keep using the old field, which we want to avoid.

vincent-psarga: that's removed

cucumber-messages/go/go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/cucumber/cucumber-messages-go/v5
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/gogo/protobuf v1.2.2-0.20190723190241-65acae22fc9d
+	github.com/golang/protobuf v1.3.2 // indirect
```
aslakhellesoy: Hmm I wonder where this comes from. We're using `gogo/protobuf`, not `golang/protobuf`...

aslakhellesoy:
The reason the build is failing is that you have deleted `cucumber-messages/go/messages.pb.go` from git.

This file needs to be checked into git, so that consumers of the package can access the file. (Consumers will not run protobuf to generate it)

PR #715, opened 2019-09-27
Generate lib/cucumber/messages_pb.rb on install

vincent-psarga:
##  Summary

gem cucumber-messages can not be installed from git as the file lib/cucumber/messages_pb.rb is not checked-out.

This ensures that the file is created.

Still TO-DO:
 - [x] Ensure the file is not regenerated (when installing via Bundler for example, the released gem contain the missing file)
 - [x] Add a target in the makefile to ensure this is tested (basically running: ```rake build && gem install pkg/cucumber-messages-*.gem``` and ensure it does not fail)

## Motivation and Context

This should allow us to reference the gem from the GIT repo when working on ```cucumber-ruby```or ```cucumber-ruby-core```.

## How Has This Been Tested?

Badly for now, see TODO section.

## Types of changes

Not applicable.

## Checklist:


vincent-psarga:
> Is this `require` still needed? It stands for "make makefile", but we're creating it manually...

From what I've need, yes, it will provide the ```find_executable```method.

vincent-psarga:
The idea was to avoid having to load the shell (for example for people using the gem on Windows). But maybe a simpler solution would be to use:

```
cd ../.. && protoc -I. --ruby_out lib/cucumber messages.proto && cd -
```

instead of:

```
pushd ../.. && protoc -I. --ruby_out lib/cucumber messages.proto && popd
```

This way we don't need bash at all. That said I'm not sure this would work at all when installing the gem on windows.
I'll have to try that out.

aslakhellesoy:
This is related to #813

aslakhellesoy:
Will this approach work if the user doesn't have `make` installed? (Thinking about Windows users)

PR #719, opened 2019-10-02
Do not update dependencies with non-final versions

vincent-psarga:
##  Summary

During release process, ```mvn versions:use-latest-versions``` may pick non-final version (rc1 etc). 
This fixes it.

## Details

There's a new template file called ```maven-versions-rules.xml``` which specifies which releases to exclude when running ```use-latest-versions```.

Maybe this should go to cucumber-parent instead, so less files to manage.

Currently the list is pretty empty (tested with the ones that get picked when running ```use-latest-versions``` for datatable), it could be easily made bigger.

## Motivation and Context

Avoid having to revert to final versions during the release process.

## Types of changes

## Checklist:

### DIFF
--- a/.templates/java/default.mk
+++ b/.templates/java/default.mk
@@ -13,7 +13,7 @@ default: .tested
 
 update-dependencies:
 	mvn versions:force-releases
-	mvn versions:use-latest-versions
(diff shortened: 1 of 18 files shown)


mpkorstanje:
I'll add this to the cucumber parent pom.

PR #720, opened 2019-10-02
Use cucumber-messages 6 for Ruby libraries

vincent-psarga:
## Summary

After Cucumber-messages 6 release, two libraries started to fail on master: gherkin and dots-formatter.
This should fix it.

The other languages do not seem to be impacted the same way. We may have to check the release process somehow.

## Checklist:


aslakhellesoy:
Looks good to me, but I would have expected a single PR that upgrades other platforms (java, javascript, go) as well.

vincent-psarga:
The master build is not failing for the other languages, that was a temporary fix until is release the other tools tomorrow :)

PR #737, opened 2019-10-07
Simplify go releases

vincent-psarga:
## Summary

There are still some manual changes that need to be done when releasing Go modules. This PR tries to automate them.
For example, `cucumber-messages 6.0.0` can not be used due to a missing manual step.

## Details

- [x] remove the `replace` lines in `go.mod` before the release
- [x] add back the `replace` lines in `go.mod` after the release
- [x] change major version in `go.mod` before release
- [x] change major version in imports in `*.go`files

## Motivation and Context

The less manual steps, the less errors (logically)

## Checklist:

### DIFF
--- a/.templates/go/default.mk
+++ b/.templates/go/default.mk
@@ -97,11 +97,11 @@ clean-go:
 	rm -rf .deps .tested .go-tested .linted dist/ .dist-compressed dist_compressed/ acceptance/
 .PHONY: clean-go
 
-remove_replaces:
+remove-replaces:
 	sed -i '/^replace/d' go.mod
 	sed -i 'N;/^\n$$/D;P;D;' go.mod
 .PHONY: remove_replaces
 
-add_replaces:
+add-replaces:
 	sed -i '/^go .*/i $(REPLACEMENTS)\n' go.mod
 .PHONY: add_replaces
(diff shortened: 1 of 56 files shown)


vincent-psarga:
We have some extra targets now in `default.mk` that allow us to simplify the release process for the go modules.
That said, I'm not 100% sure on how to correctly plug them.

In the RELEASE_PROCESS file, we tell to remove the `replace` directives in `go.mod` before running `make update-dependencies`.
Do you think I should make the `remove-replaces` a requirement for the `update-dependencies` target ? (I'm not really convinced by that, we may want to run `update-dependencies` without remove the `replace` directives.

Another solution I would see would be to add a `pre-release` target. For most languages, it would be:

```makefile
pre-release: update-dependencies clean default
.PHONY: pre-release
```

and for go:

```makefile
pre-release: remove-replaces update-dependencies clean default
.PHONY: pre-release
```

What would be your thoughts about that solution ?

**Note:** I'm not sure we can reference default as simply in Makefiles

.templates/go/default.mk
```diff
@@ -48,6 +51,9 @@ update-dependencies:
 	cd $(MOD_DIR) && go get -u && go mod tidy
 .PHONY: update-dependencies
 
+pre-release: update-major update-dependencies clean default
```
vincent-psarga: should be:

```makefile
pre-release: remove-replaces update-major update-dependencies clean default
```

Just saw that, I finish the release of another package and fix that.

PR #742, opened 2019-10-08
tag-expressions: convert from junit v4 to junit v5

nhojpatrick:
## Summary

JUnit v5 jupiter update only

## Details

Change from junit v4 to junit v5.
Also fix hamcrest meta project dependency, changing to org.hamcrest:hamcrest:2.1.

## How Has This Been Tested?

`mvn clean install` for using both jdk 8 and 11

## Types of changes

## Checklist:


nhojpatrick:
regarding assertThrows i've broken most into using testMethod to try and break up potential huge blocks of code into hopefully easier to read chunks

i've also change some lambdas into old style because the classname appears in error messages

i've used hamcrest is(equalTo(xxx)) so the full error message is check, as some of the previous message checking was doing a contains type check, so want to make sure the whole message was checked.

PR #750, opened 2019-10-11
Gherkin pure java

aslakhellesoy:
Restores the pure Java Gherkin parser from v5. It's been updated with the following:

- Use cucumber-messages
- Add pickle IDs
- Use `java.util.stream`

This is left to do:

- [x] Shade all dependencies in cucumber-messages (there are a few transitive ones that aren't shaded yet)
- [x] Replace `Iterable<Envelope>` with `Stream<Envelope>` in the public API (`Gherkin` class)
- [x] Get completely rid of C21e - I don't think we're going to need it.
- [x] Update the docs/examples


gherkin/java/pom.xml
```diff
@@ -51,27 +57,34 @@
...
                         <phase>package</phase>
                         <goals>
-                            <goal>single</goal>
+                            <goal>shade</goal>
                         </goals>
+                        <configuration>
+                            <relocations>
+                                <relocation>
+                                    <pattern>com.eclipsesource.json</pattern>
+                                    <shadedPattern>io.cucumber.gherkin.com.eclipsesource.json</shadedPattern>
```
mpkorstanje: Using `io.cucumber.gherkin.com.internal.eclipsesource.json` will make it clear that this is an internal package and shouldn't be reused by down stream

aslakhellesoy: Right, so use `internal` as a prefix for all shaded packages? Makes sense. I've also added a script that will check for the presence of unshaded classes in jar across the monorepo.

Issue #734, opened 2019-10-07
Document how to coordinate releases


aslakhellesoy:
Lots of other tools depend on Cucumber's APIs and outputs. While it's important to improve Cucumber, it is also important to reduce the amount of friction a new release causes for our users.

We need to give tool vendors a chance to update their products to upcoming changes that break backwards compatibility. This is what I propose:

* All major releases (incompatible API changes) MUST be preceded by at least one pre-release
* At least 4 weeks MUST elapse between a major prerelease and a major release
* APIs SHOULD be deprecated before being removed/changed in an incompatible manner
* Deprecation notices SHOULD suggest what users can do to migrate to the new API

This is not an exhaustive or final list - feedback is very welcome.

SabotageAndi:
Is this only for Cucumber or also for libraries (Gherkin) and protocol specifications (cucumber-messages)?

vincent-psarga:
I'm not sure this is needed for libraries too, as it is always possible to pin previous versions of libraries (although this could be interesting too) in tools. It's a bit more difficult with integrations of other tools.

aslakhellesoy:
> Is this only for Cucumber or also for libraries (Gherkin) and protocol specifications (cucumber-messages)?

This is only for the packages that users install directly (cucumber gem/npm/maven). Cucumber's dependencies can still be released "whenever", since this won't affect end users.

The release cycle for SpecFlow is up to the SpecFlow team of course, but if you want to avoid breaking consumers who might be relying on an API or format you're exposing it's probably a good idea to do something similar for SpecFlow.

SabotageAndi:
We have no written rules in SpecFlow about backwards compability, but we really try to not add breaking changes. Also in Major/Minor version updates.
If we do, we are approximately doing the suggested steps. We could be better with deprecation notes.

luke-hill:
Left-wing thought.

Have a mailing list of large (definition TBD), people / companies who use our products and might benefit from knowledge of this.

Initial thoughts are:
Ruby: Rails / Capybara
JS: NPM / wdio
Java: Spring
Then maybe a generic Jetbrains one? Maybe VS for C# people?

I've come back from a week or so off and I'm a bit flued up atm, so maybe this is a bad idea. Just throwing it out there.

mpkorstanje:
A release is a PR opportunity. Shouldn't be missed. Currently all I do is write a blog post (when I feel like it) and tweet at `@cucumberbdd` (when I feel like it).

[...]

PR #760, opened 2019-10-15
First try for the code-names document

vincent-psarga:
Related to #734 - this is an example of how we could format / unify releases of Cucumber across multiple platforms.


vincent-psarga:
To be followed here: https://github.com/cucumber/docs.cucumber.io/pull/402

PR #766, opened 2019-10-18
cucumber-html-formatter docker image

aslakhellesoy:
The HTML formatter is a Node.js application. Since it's meant to be used with any Cucumber implementation, the users may not be familiar with Node.js, or have it installed.

The same might be true for Docker, but having more ways to distribute the tool increases the number of ways it can be installed.

This PR adds scripts for packaging and publishing [cucumber-html-formatter](https://cloud.docker.com/u/cucumber/repository/docker/cucumber/cucumber-html-formatter). Version 2.0.1 has been published already. We need a little more documentation to the release process how to use this. We need to document that it cannot be published from the docker container we use to release everything else - it has to be built and released from the host OS.


vincent-psarga:
> We need to document that it cannot be published from the docker container we use to release everything else - it has to be built and released from the host OS.

We should avoid that I think. Instead of pushing to DockerHub, let's use the pull functionality. Every time a tag and/or commit is pushed, the image will be built.

On `hiptest-publisher`, we have the following rules set on DockerHub:

Screenshot 2019-10-21 at 10 01 58

This way:
 - each tag has a corresponding image built
 - `latest` points to the latest tag image
 - `master` allows using the master version (not sure it would be useful here)

So I'd deactivate the `publish` target in the Makefile to let DockerHub handle that (less manual steps, less risks of forgetting it :) )

aslakhellesoy:
@vincent-psarga I think that requires setting up a read-only mirror on GitHub where the `Dockerfile` is at the root.

PR #769, opened 2019-10-23
gherkin: Do not trim spaces inside a table cell content

### LINKED ISSUE #290: gherkin: Leading and trailing "\n" are trimmed in DataTable cells

vincent-psarga:
## Summary

If your datatable contains this:

```
| column 1       | column 2                                 | column 3                                            |
| some data     |                    some other data | some\n               other\n                  data  |
```

After parsing, you should get those values:
 - `some data`(white spaces after removed)
 - `some other data`(white spaces before removed)
 -  `some\n               other\n                  data ` (white spaces inside left as it)

Currently, we "may" get `some\nother\ndata` instead/

**UPDATE:** that's not linked to issue #290, I misread the issue ... I'll remove the references in the commits before squashing to avoid having wrong references in the issue.

## TODO

- [x] Add generic example in `gherkin/testdata/good`

## How Has This Been Tested?

With unit tests (and later on acceptance one, but I think the acceptance ones are important to really understand what happens)

## Types of changes

## Checklist:


gherkin/javascript/test/GherkinLineTest.ts
```diff
@@ -0,0 +1,20 @@
...
+import GherkinLine from '../src/GherkinLine'
+
+describe('GherkinLine', function() {
+  function getCellsText(line: string) {
+    const gl = new GherkinLine(line, 1)
+
+    return gl.getTableCells().map(span => span.text)
+  }
+
+  it('getTableCells correctly trims cells content', function() {
```
aslakhellesoy: Nitpick: Create a nested `describe("#getTableCells()")` block and use one test per assertion. Instead of saying "correctly" (which is open to interpretation), be explicit, such as `it("trims whitespace after")` etc.

gherkin/javascript/test/GherkinLineTest.ts
```diff
@@ -0,0 +1,20 @@
+import * as assert from 'assert'
+import GherkinLine from '../src/GherkinLine'
+
+describe('GherkinLine', function() {
```
aslakhellesoy: We use arrow functions `() => {}` in mocha tests

gherkin/javascript/test/GherkinLineTest.ts
vincent-psarga: Oh nice I did not know we were able to have nested `describe` in mocha, that's great :)

gherkin/javascript/test/GherkinLineTest.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    it('trims white spaces after cell content', () => {
+      assert.deepEqual(getCellsText("|spaces after   |"), ['spaces after'])
+    })
+
+    it('trims white spaces around cell content', () => {
+      assert.deepEqual(getCellsText("|   \t spaces everywhere   \t|"), ['spaces everywhere'])
+    })
+
+    it('does not delete white spaces inside a cell', () => {
```
luke-hill: not super important. But you could maybe rename this to 

`it does not delete white spaces in the middle of cell content`

mpkorstanje:
So only Go left?

mpkorstanje:
> So only Go and the acceptance tests left?

Looks like I already added acceptances while fixing https://github.com/cucumber/cucumber/pull/891

PR #770, opened 2019-10-23
Added expected regex for when an optional word may be needed in front…

spatel1009:
… of alternative group

cucumber-expression-pattern-test

## Summary

Added a test for the expected regex pattern for when an optional word may be needed in front of alternative non-capture group.

## Types of changes

**No changes or new code has been made towards the app**

## Checklist:

### DIFF
--- a/cucumber-expressions/java/src/test/java/io/cucumber/cucumberexpressions/CucumberExpressionPatternTest.java
+++ b/cucumber-expressions/java/src/test/java/io/cucumber/cucumberexpressions/CucumberExpressionPatternTest.java
@@ -39,7 +39,7 @@ public class CucumberExpressionPatternTest {
     @Test
     public void translates_alternation_with_optional_words() {
         assertPattern(
-            "the (test )chat/call/email interactions are visible,
+            "the (test )chat/call/email interactions are visible",
             "^the (?:test )?(?:chat|call|email) interactions are visible$"
         );
     }
--- a/cucumber-expressions/java/src/test/java/io/cucumber/cucumberexpressions/CucumberExpressionPatternTest.java
+++ b/cucumber-expressions/java/src/test/java/io/cucumber/cucumberexpressions/CucumberExpressionPatternTest.java
@@ -35,6 +35,14 @@ public class CucumberExpressionPatternTest {
                 "^I said (?:Alpha1|Beta1)$"
         );
     }
+    
+    @Test
+    public void translates_alternation_with_optional_words() {
+        assertPattern(
+            "the (test )chat/call/email interactions are visible,
+            "^the (?:test )?(?:chat|call|email) interactions are visible$"
+        );
+    }
 
     @Test
     public void translates_parameters() {


cucumber-expressions/java/src/test/java/io/cucumber/cucumberexpressions/CucumberExpressionPatternTest.java
```diff
@@ -35,6 +35,14 @@ public void translates_alternation_with_non_alpha() {
                 "^I said (?:Alpha1|Beta1)$"
         );
     }
+    
+    @Test
+    public void translates_alternation_with_optional_words() {
+        assertPattern(
+            "the (test )chat/call/email interactions are visible,
```
luke-hill: we need a `"` at the end of this line, so there's no compilation error.

luke-hill:
The error does look legit. But what's interesting is where it's tripping up is where we half expected it to.

Can you move the space one place back so instead of `(capture )` you have `( capture)` this should ensure a space between the right `)` and the start of the 3 alternated words. This would be a good one to see the output of.

spatel1009:
> The error does look legit. But what's interesting is where it's tripping up is where we half expected it to.
> 
> Can you move the space one place back so instead of `(capture )` you have `( capture)` this should ensure a space between the right `)` and the start of the 3 alternated words. This would be a good one to see the output of.

Change it to `the( test) chat/call/email interactions are visible`?

mpkorstanje:
It's ok. I found the problem already.

Given any expression of the form `(a )c/d`.

We first rewrite the optionals to regex so we get `(?:a )?c/d`.

Then we rewrite the alternatives so we get `(?:a (?:(?:)?c)|(?:d))` which happens to be a valid regex.

As you can see after rewriting the optionals we rewrite the alternatives. And because the rewritten optionals ends with ` )?` those last two characters are then rewritten into the group for the alternatives.

luke-hill:
This is really where colours in markdown would help. Took me 15mins to figure out that regex. But yeh @spatel1009 that is right what you said. That should mitigate (But the original post is still causing some friction I agree)

spatel1009:
Doing it that way with the( test) call/chat/email will only read a Gherkin written as `Then the test call/chat/email....` but not any of those singular like `Then the test call....` etc.  Also "test" is no longer be optional as steps like `Then the chat....` will not be read either.

luke-hill:
I'm not sure I understand

`Then the( test) call/chat/email` will match

`Then the call`
`Then the chat`
`Then the test email`

spatel1009:
Yes I thought so too but I get "cannot find declaration to go to" when using that above.  It reads none of the examples you have listed and the only one it reads is:
`Then xyza the test call/chat/email interactions page successfully loads with an interaction` 
Written exactly like that with all 3 alternative text options written with the /.

luke-hill:
This feels like we're finding more edge cases as we go.

Ok noted! I'll have a little dabble and liaise up with @mpkorstanje to see if we can add tests for this also.

mpkorstanje:
Superseded by #771

Issue #767, opened 2019-10-22
Optional text not working as expected


spatel1009:
## Summary

When using optional text in the stepdef, it is not being read by the Gherkin as expected

## Expected Behavior

Would like to use optional word where it is applicable to all alternative words following and not just the first one listed.  Currently the third option below is what we are using since in that scenario only call needs the word test in front of it, but for future use may be instances were multiple alternative text have that optional word in front and would like to have possibility of being interchangeable.

## Current Behavior

`the( test) chat/email/call interaction is not visible` 
only reads: **"And the test chat/email/call interaction is not visible"** 
Will not read: **"And the chat interaction is not visible"** 
Nor singular: **"And the test chat interaction is not visible"**  
Summary:  Will only read where "test" is used and not optional with all 3 alternative words also included in the Gherkin

`the (test) chat/email/call interaction is not visible` 
Will read **"And the test call interaction is not visible"** 
But not **"And the call interaction is not visible"**.  
Summary:  Will read only when test is put in front of the 3 options, not without.  Basically renders "test" not optional

`"the (test )call/email/chat interaction is not visible"` 
Will read **"And the test call interaction is not visible"** 
And also:` "And the chat interaction is not visible"` 
But not `"And the test chat interaction is not visible"`
Summary:  Will read the use of test when in front of call only, not for email or chat.

## Steps to Reproduce (for bugs)

Unable to provide company code, however, creating a stepdef that needs optional words in front of 2 out of 3 alternative text directly following it will help recreate issue.

## Context & Motivation

This issue is for non-capture groups so this allows for more clear readable gherkin linking to same test for multiple portions of app.

## Your Environment

Java

luke-hill:
Just for benefit (Although you have half done it), you "could" reproduce the failing code. As you could create a feature that just printed `nil` to stdout for each step def or a blank string.

I'll try reproduce this issue, but as I'm a rubyist, this may be a java-centric problem. I don't know without a full repro scenario.

luke-hill:
The best type

aslakhellesoy:
@luke-hill if this really is a bug, it's likely present in all implementations (including Ruby). The easiest way to reproduce it is to write a failing test in `CucumberExpressionTest`

spatel1009:
Hey not sure if this is what you meant but I added a couple tests in how we are trying to use it?

https://github.com/cucumber/cucumber/compare/master...spatel1009:patch-1

luke-hill:
[...]

Issue #601, opened 2019-04-28
Incorrect Match Step Definitions if Optional text past on Alternative text


sgrillon14:
Hi,
I find a case with a bug:

`"I wait {int} second(s)\\./\\?"` not match this gherkin: `I wait 2 second?`

    
## Summary

I create lot of sample and I find a case with a bug if Optional text past on Alternative text

```
import java.util.ArrayList;
import java.util.List;
import java.util.Locale;

import com.google.gson.Gson;

import io.cucumber.cucumberexpressions.Argument;
import io.cucumber.cucumberexpressions.CucumberExpression;
import io.cucumber.cucumberexpressions.Expression;
import io.cucumber.cucumberexpressions.ParameterTypeRegistry;

public class Poc {


    public static void main(String[] args) {
        String res = new Gson().toJson(match("I wait {int} second(s)", "I wait 2 seconds"));
        System.out.println(res);
        res = new Gson().toJson(match("I wait {int} second(s)", "I wait 1 second"));
        System.out.println(res);
        res = new Gson().toJson(match("I wait {int} second(s).", "I wait 2 seconds."));
        System.out.println(res);
        res = new Gson().toJson(match("I wait {int} second(s) \\./\\?", "I wait 2 seconds ."));
        System.out.println(res);
        res = new Gson().toJson(match("I wait {int} second(s) \\./\\?", "I wait 2 seconds ?"));
        System.out.println(res);
        res = new Gson().toJson(match("I wait {int} second(s) \\./\\?", "I wait 2 seconds x"));
        System.out.println(res);
        res = new Gson().toJson(match("I wait {int} second(s)\\./\\?", "I wait 2 seconds."));
        System.out.println(res);
        res = new Gson().toJson(match("I wait {int} second(s)\\./\\?", "I wait 2 secondsx"));
        System.out.println(res);
        res = new Gson().toJson(match("I wait {int} second(s)\\./\\?", "I wait 2 seconds?"));
        System.out.println(res);
    }

    private static List<?> match(String expressionString, String text) {
        Expression expression;
        ParameterTypeRegistry parameterTypeRegistry = new ParameterTypeRegistry(Locale.ENGLISH);
        expression = new CucumberExpression(expressionString, parameterTypeRegistry);
        List<Argument<?>> args = expression.match(text);
        if (args == null) {
            return null;
        } else {
            List<Object> list = new ArrayList<>();
            for (Argument<?> arg : args) {
                Object value = arg.getValue();
[...]

PR #771, opened 2019-10-24
cucumber-expressions: Use a parser to parse Cucumber Expressions

mpkorstanje:
## Summary

A parser to parse Cucumber Expressions.

Cucumber Expressions currently uses a string rewriting algorithm
to turn Cucumber Expression Strings into regular expressions. This algorithm
has three major problems.

 1. Rewriting optionals interferes with rewriting alternatives as seen in #767.
 2. It is possible to inject regex expressions into Cucumber Expressions as seen in #601.
 3. It is not possible use alternatives containing a space.
 4. Escaping characters is generally inconsistent.

By using a parser to generate the AST of Cucumber expression and rewriting this
AST to a regular expression we can avoid all these problems. For example:

```
{int} blind\ rat(s)/mouse/mice
```

Matches:
* 3 blind rats
* 1 blind rat
* 3 mice
* 1 mouse

Fixes: #767
Fixes: #601
Fixes: #726
Closes: #770

## Details

Grammar

```
cucumber-expression :=  ( alternation | optional | parameter | text )*
alternation := (?<=left-boundary) + alternative* + ( '/' + alternative* )+ + (?=right-boundary)
left-boundary := whitespace | } | ^
right-boundary := whitespace | { | $
alternative: = optional | parameter | text 
optional := '(' + option* + ')'
option := optional | parameter | text
parameter := '{' + name* + '}'
name := whitespace | .
text := whitespace | ')' | '}' | .
```

The AST is constructed from the following tokens:
```
escape := '\'
token := whitespace | '(' | ')' | '{' | '}' | '/' | .
. := any non-reserved codepoint
```

Note:
 * While `parameter` is allowed to appear as part of `alternative` and 
  `option` in the AST, such an AST is not a valid a Cucumber Expression.
 * While `optional` is allowed to appear as part of `option` in the AST,
   such an AST is not a valid a Cucumber Expression.
 * ASTs with empty alternatives or alternatives that only
   contain an optional are valid ASTs but invalid Cucumber Expressions.
 * All escaped tokens (tokens starting with a backslash) are rewritten to their
   unescaped equivalent after parsing.

## Types of changes

## Checklist:


mpkorstanje:
I've implemented the parser in Go and Java. I'm fairly proficient with both, before I continue with JavaScript and Ruby I'd like to have some feedback.

mpkorstanje:
Nevermind. Found the flaws.

Cases not yet covered:

```
aaa \( bbb (cccc)
aaa (bbb \)ccc)
aaa \(bbb\)  ccc
```

Ect. 

Currently can't catch these because this parser rewrites tokens collected by regex rather implementing a proper tokenizer + LR(1) parser.

cucumber-expressions/README.md
```diff
@@ -1,9 +1,27 @@
 See [website docs](https://cucumber.io/docs/cucumber/cucumber-expressions/) for more details.
 
+
+## Grammar ##
+
+```
+cucumber-expression :=  [ optional | escaped-optional | other-then-optional ]*
```
luke-hill: Is this an indicative order of the algorithm?

mpkorstanje: Yes. A precedence order parser first parses the highest precedence operator, then the lower precedence operators.

mpkorstanje:
> Personal pref, I don't like this (4 escaping \), type of cuke exp. When you've got 4 escaping characters it's starting to get beyond the point that we should support (imo)

We support two escapes at most currently. But if the parser is implemented properly it doesn't matter.

cucumber-expressions/go/cucumber_expression.go
```diff
@@ -78,83 +94,140 @@ func (c *CucumberExpression) Source() string {
 	return c.source
 }
 
-func (c *CucumberExpression) processEscapes(expression string) string {
-	return ESCAPE_REGEXP.ReplaceAllString(expression, `\$1`)
+func (c *CucumberExpression) escapeRegexAndJoin(expression []token, prefix string, suffix string) string {
+	var builder strings.Builder
```
ankitpokhrel: `strings.Builder` is only available since golang 1.10, don't we support earlier versions?

luke-hill: golang is a new language and it's not widely used. So I think it makes sense to only support new(er), versions. But I don't fully use it yet, so maybe I'm not the best person to have an opinion

ankitpokhrel: In that case, maybe we should specify a min supported version somewhere on the docs (if there isn't already any).

aslakhellesoy: @luke-hill Go is 10 years old and I'd argue it's [widely used](https://www.tiobe.com/tiobe-index/)

That said I think we should be explicit about what Go versions are supported, and I think it's fine to say >= 1.10 until somebody complains about it.

mpkorstanje:
Okay. I've changed the implementation to a regular recursive descent parser. While not strictly necessary it makes the implementation much cleaner and behaves like all the other parsers people are used too (esp w.r.t  escapes). 

Additionally I've changed the precedence order. Alternation is now the highest precedent operator. This means that any thing bordering a `/` without spaces is part of the alternation (it used to be bound by optionals, but that was broken).

To make cucumber expressions easier to validate I've also extended the grammar to allow parameters to be nested in optionals and alternatives and optionals to be nested in parameters. These are not valid cucumber expressions and will throw an error. 

For now I'm going to let this percolate a bit. I'll add some examples (for the docs) later on.

laeubi:
@mpkorstanje thats great :+1: 
This is something we could definately use/require to support better IDE integration of cucumber (see https://github.com/cucumber/cucumber-jvm/issues/1711) and will greatly improve the further feature development of cucumber-expressions.
As there are ots of changes I'm not completely sure but will it be possible to use acess the different parts of an expression via public API?
What would be required is

- Get parameters/alternatives and there positions in the origial string
- Get type of parameters
- Get alternative values

For your example ({int} blind\ rat(s)/mouse/mice) I would expect the following data:

- one parameter at (0, 5) of type int
- one alternative at (13, 30) with values [rat, rats, mouse, mice]

Is this already possible? If not can you add it?

mpkorstanje:
> Is this already possible? If not can you add it?

Feel free to send a PR against the `tokenize-cucumber-expression` branch. The java implementation is pretty much  done.

You'll probably want to add start and end indexes to `Token` and then reuse that information in the `Node` class to work out the bounds. Once that works we can hammer out the public API for it.

cucumber-expressions/java/src/test/java/io/cucumber/cucumberexpressions/CucumberExpressionTest.java
```diff
@@ -37,6 +36,56 @@ public void matches_word() {
...
+
+    @Test
+    public void matches_optional_in_alternation() {
+        assertEquals(singletonList(3), match("{int} rat(s)/mouse/mice", "3 rats"));
+        assertEquals(singletonList(2), match("{int} rat(s)/mouse/mice", "2 mice"));
+        assertEquals(singletonList(1), match("{int} rat(s)/mouse/mice", "1 mouse"));
+    }
+
+    @Test
+    public void matches_optional_before_alternation_with_regex_characters() {
```
mpkorstanje: This test should be added to `examples.txt`

cucumber-expressions/java/src/test/java/io/cucumber/cucumberexpressions/CucumberExpressionTest.java
```diff
@@ -37,6 +36,56 @@ public void matches_word() {
...
+    }
+
+    @Test
+    public void matches_optional_before_alternation() {
+        assertEquals(emptyList(), match("three (brown )mice/rats", "three brown mice"));
+        assertEquals(emptyList(), match("three (brown )mice/rats", "three rats"));
+    }
+
+    @Test
+    public void matches_optional_in_alternation() {
```
mpkorstanje: This test should be added to `examples.txt`

mpkorstanje:
And that is the go implementation of the recursive descent parser done.

aslakhellesoy:
Great work @mpkorstanje! Looking forward to porting this to TypoeScript. Maybe this cleanup will make it easier to port to C# too? /cc @SabotageAndi @gasparnagy

aslakhellesoy:
Can't believe I just wrote TypoeScript!

cucumber-expressions/go/cucumber_expression_parser_test.go
```diff
@@ -0,0 +1,344 @@
...
+			expressionNode,
+			[]astNode{
+				{textNode, []astNode{}, token{"{", beginParameter}},
+			},
+			nullToken,
+		})
+	})
+
+	t.Run("unfinished parameter", func(t *testing.T) {
+		assertAst(t, "{string", astNode{
```
Ajwah: I am surprised this would not raise an exception instead. Surely if someone meant to `{string` they would escape it: `\\{string`?  (I am genuinely curious as for CucumberExpressions I am implementing I am raising an error for this exact scenario.)

mpkorstanje: Good point.

gasparnagy:
I know it is too late and I don't want to screw up the plan, but did you consider using BERP for parsing? That would allow and easier port to other languages.

mpkorstanje:
I don't believe Berp allows look behind `(?<=boundary) ` and using a windows exectuable was a bit of a barrier to entry for me. :smile:

gasparnagy:
@mpkorstanje Well, it is not a Windows executable, but a .NET executable and can also be run on any platform using mono, but I can understand. It does not support lookbehind, but it supports rule-level lookahead and that is enough, I think. (For the first look, the `alternative*` inside the `alternation` rule sounds strange to me.) Berp is used to parse the [berp grammar](https://github.com/gasparnagy/berp/blob/master/src/Berp/BerpGrammar/BerpGrammar.berp) itself, that also has alternation.

Are you otherwise ready with the Java implementation? Shall I check how berp could support this and discuss it in the tomorrows Cucumber meeting?

mpkorstanje:
I don't think you'll get there with just look ahead. Perhaps you could do it by making the boundary part of the `alternation` node but then you'd have to rewrite the AST afterwards to move the boundry out of the `alternation` node again.

The challenge with alternation were things like this:

```
pre a/(b c)/d post
pre (a b)/c/d post
pre a/{b/c/}d post
```

mpkorstanje:
> Are you otherwise ready with the Java implementation? Shall I check how berp could support this and discuss it in the tomorrows Cucumber meeting?

There is a still the open question what we'll do with incomplete optional and parameter types. E.g: `"{string"` and `"(string"`

>  Shall I check how berp could support this and discuss it in the tomorrows Cucumber meeting?

I'm not quite convinced that generating the parser will make the code easier to understand. But lets discuss it tomorrow anyway.

gasparnagy:
@mpkorstanje OK. I'll think about it.

Just to make sure I understand the tricky cases:

* `pre a/(b c)/d post` -- this matches to `pre a post` `pre b c post` `pre post` and `pre d post`
* `pre (a b)/c/d post` -- this matches to `pre a b post` `pre post` `pre c post` and `pre d post`
* `pre a/{b/c/}d post` -- for me this sounds like invalid, because `b/c/` is not a type name. But even if it would be `pre a/{int}d post`, I would still expect that to be invalid because it would make the parameter optional that would mix the matching to the actual parameters of the method. I think I would not allow parameters in alternation and optional at all...

mpkorstanje:
> Just to make sure I understand the tricky cases:

I don't think you understood them. They are all valid cucumber expression ASTs with one alternation node in the parse tree.

>  I would still expect that to be invalid because it would make the parameter optional that would mix the matching to the actual parameters of the method. I think I would not allow parameters in alternation and optional at all...

While using a parameter inside an alternative does not result in valid cucumber expression, this should yield a valid AST. Not treating it as parse error makes it easier to report what the user did wrong because we have the whole AST to work out what is actually wrong.

Using an optional inside alternatives are entirely fine.

It's what allows the variations of this expression to work:

```
{int} blind\ rat(s)/mouse/mice
```

e.g: 

```
    3 blind rats
    1 blind rat
    3 mice
    1 mouse
```

gasparnagy:
My question was about whether the expressions are valid and what steps would they match to. (So basically as a cucumber user what should I expect by looking at the expression.) Could you confirm that on this level the potential matches I listed are correct?

> While using a parameter inside an alternative does not result in valid cucumber expression, it should yield a valid AST. Not treating it as parse error makes it easier to report what the user did wrong because we have the whole AST to work out what is actually going on rather then a stack in the parser.

Well... I think this depends on the error reporting capabilities of the parser... Because creating an AST for something that is obviously senseless makes the code more complex.

> Using an optional inside alternatives are entirely fine. 

I had no concerns about that. What I said is using parameters inside alternatives (or optional) is the problem.

(Although I think we could even forbid optional inside alternation for the sake of simplicity, because the very few cases where it would make sense can be also done with an empty alternate case. E.g. `a/(b)/c` with `a/b//c`...)

mpkorstanje:
> My question was about whether the expressions are valid and what steps would they match to

> pre a/(b c)/d post -- this matches to pre a post pre b c post pre post and pre d post

Correct.

> pre (a b)/c/d post -- this matches to pre a b post pre post pre c post and pre d post

That would be `pre  post` with two spaces.

Note: We also disallow empty alternatives so an optional that spans the entire alternative is also forbidden.

> pre a/{b/c/}d post -- ...

If we'd allows parameter types in alternatives and if `b/c/` would have the pattern `.+` then the `pre a/{b/c/}d post` matches `pre a post`, `pre .+ post` and `pre d post`. 

> I think we could even forbid optional inside alternation for the sake of simplicity

I tried that initially but the syntax becomes much easier to understand when it can be used consistently. 

It would also break backwards compatibility.

mpkorstanje:
Note; We should update the cucumber expression generator to generate properly escaped cucumber expressions.

gasparnagy:
@mpkorstanje I published the demo to https://github.com/gasparnagy/CucumberExpressionParserDemo

I have made a new release of BERP, where I back-ported the tweaked razor files from the Gherkin project. (The original idea was that the razor files should not be project dependent. Now I made them project-independent again, so we could generate both Gherkin and Cucumber Expressions with the same razor files -- with different grammar, of course :))

The new release is on NuGet, but can also be downloaded as a simple zip from: https://www.nuget.org/api/v2/package/Berp/1.1.1. 

The ideal would be to make the cucumber build in a way that it wgets BERP and runs the parser generations from there. (So we don't need to commit the binaries.) I would make this for the cucumber expressions and then fix the Gherkin parser generation as well. (I am happy to help.)

Play with it and let me know if something is unclear.

mpkorstanje:
Cheers! I'll have a look at putting berp into the docker image.

gasparnagy:
@mpkorstanje Sure. I have another tool with a similar approach, for that I documented how to use it through docker. If it helps... https://specsolutions.gitbook.io/specsync/important-concepts/using-specsync-on-osxlinux-page#dockerfile-for-mono-execution

mpkorstanje:
@gasparnagy given these rules below I get a parse error while parsing `(three )rats/(some )cats`:

```
[
	Tokens -> #TEXT,#WHITE_SPACE,#ALTERNATION,#BEGIN_OPTIONAL,#END_OPTIONAL,#BEGIN_PARAMETER,#END_PARAMETER
	ClassName -> CucumberExpressionParser
	ParserClassName -> CucumberExpressionParser
	JAVA_Package -> io.cucumber.cucumberexpressions
	SimpleTokenMatcher -> true
	MaxCollectedError -> 1
]


CucumberExpression! := (Separator | Parameter | Alternation )*
Separator! := #WHITE_SPACE
Alternation! := Alternate (#ALTERNATION Alternate)*
Alternate! [->#TEXT] := (Optional | Parameter | AlternateText )+
AlternateText! := #TEXT
Optional! := #BEGIN_OPTIONAL (Text)* #END_OPTIONAL
Parameter! := #BEGIN_PARAMETER (Text)* #END_PARAMETER
Text! := (#TEXT | #WHITE_SPACE)
```

This is rather curious because the `#TEXT` token is expected here according to the diagnostics generated by Berp:
 
```
token = {Token@1710} "Token{text='rats', type=TEXT}"
 text = "rats"
 tokenType = {CucumberExpressionParser$TokenType@1681} "TEXT"
expectedTokens = {Arrays$ArrayList@1723}  size = 6
 0 = "#EOF"
 1 = "#BEGIN_OPTIONAL"
 2 = "#BEGIN_PARAMETER"
 3 = "#TEXT"
 4 = "#ALTERNATION"
 5 = "#WHITE_SPACE"
stateComment = "State: 7 - CucumberExpression:0>__alt0:2>Alternation:0>Alternate:1>__alt2:0>Optional:2>#END_OPTIONAL:0"
```

Debugin the generated code it looks like the `lookahead_0` requires that the next node is a `#TEXT` nodes while the next token is an `#ALTERNATION` . 

```
        if (match_TEXT(context, token))
        {
            if (lookahead_0(context, token))
            {
                endRule(context, RuleType.Optional);
                startRule(context, RuleType.AlternateText);
                build(context, token);
            return 11;
            }
        }
        if (match_TEXT(context, token))
        {
            if (lookahead_0(context, token))
            {
                endRule(context, RuleType.Optional);
                endRule(context, RuleType.Alternate);
                endRule(context, RuleType.Alternation);
                startRule(context, RuleType.Alternation);
                startRule(context, RuleType.Alternate);
                startRule(context, RuleType.AlternateText);
                build(context, token);
            return 11;
            }
        }
```

So trying to include `#ALTERNATION` in the lookahead:

```
Alternate! [->#TEXT | #ALTERNATION] := (Optional | Parameter | AlternateText )+
```
I run into a parse error in Berp:

```
Generating parser for grammar 'CucumberExpressionGrammar.berp' using template 'Java.razor'.
Loading grammar...

Generation failed.
Parser errors:
(14:35): expected: #AlternateOp, #Arrow, got ']'
(14:37): expected: #AlternateOp, #Arrow, got ':='
(14:40): expected: #AlternateOp, #Arrow, got '('
(14:41): expected: #AlternateOp, #Arrow, got 'Optional'
(14:52): expected: #Token, got 'Parameter'
(14:62): expected: #Token, got '|'
(14:64): expected: #Token, got 'AlternateText'
(14:78): expected: #Token, got ')'
(14:79): expected: #Token, got '+'
(14:79): expected: #Token, got ''
make: *** [Makefile:4: src/main/java/io/cucumber/cucumberexpressions/CucumberExpressionParser.java] Error 1
```

And looking at the berp grammar I don't think I'm too far off:

```
LookAhead! := #LBracket LookAheadTokenList? #Arrow LookAheadTokenList #RBracket
LookAheadTokenList! := #Token (#AlternateOp #Token)*
```

What am I missing here?

gasparnagy:
@mpkorstanje Interestingly I've run into the same problem. I agree, that the grammar should be valid. I will take a look, in the meanwhile please revert to the previous grammar. (I can only check it on Monday).

gasparnagy:
@mpkorstanje I've got the fix for this. The multiple lookahead error was a clear bug in BERP. Fixed.

But... Actually I had to realize that we use the lookaheads here in a non-designed way. BERP lookaheads can help if you need to choose from two branches based on the tokens that come after the first token (hence "ahead"). But in our case we don't really want to look ahead, but we always want to keep the parser continuing the Alternate sequence and do not fall up to the CucumberExpression sequence (and come back to the Alternate rule). So basically we want to have a simple "priority". To be able to support that, I needed to tweak BERP a bit. To keep the things simple, the priority has to be specified as a special, empty lookahead rule `[->]`. 

So basically what you need to do:
* Upgrade to BERP 1.2.0 (https://www.nuget.org/packages/Berp/1.2.0)
* Change the lookahead for rule `Alternate` to `[->]` (remove `#TEXT`)

I've [tested it](https://github.com/gasparnagy/CucumberExpressionParserDemo/commit/42485b11e30b765092982414390926f9f4c4782b) and with that it was able to parse `(three )rats/(some )cats` as

```
<CucumberExpression>: []
  <Alternation>: []
      <Alternate>: []
          <Optional>: [three ]
          <Text>: [rats]
      <Alternate>: []
          <Optional>: [some ]
          <Text>: [cats]
```

mpkorstanje:
Cheers that works!

@gasparnagy now I'm trying to parse an empty alternative `a//b`. Or more generally speaking `//`. This is not a valid cucumber expression but it should provide a valid parse tree. This allows us to create a clear error message to explain that empty alternatives are not allowed an highlight common mistakes such as using urls (`https://github.com/`) as part of a step definition.

```java
    @Test
    void emptyAlternative() {
        assertThat(astOf("//"), equalTo(
                new AstNode(RuleType.CucumberExpression,
                        new AstNode(RuleType.Alternation,
                                new AstNode(RuleType.Alternate),
                                new AstNode(RuleType.Alternate),
                                new AstNode(RuleType.Alternate)
                        )
                )
        ));
    }
```

To do this I made the Alternate text have 0...n occurances of text tokens using `AlternateText! := #TEXT*`. Giving me the following berp:

```
[
	Tokens -> #TEXT,#WHITE_SPACE,#ALTERNATION,#BEGIN_OPTIONAL,#END_OPTIONAL,#BEGIN_PARAMETER,#END_PARAMETER
	ClassName -> CucumberExpressionParser
	ParserClassName -> CucumberExpressionParser
	JAVA_Package -> io.cucumber.cucumberexpressions
	SimpleTokenMatcher -> true
	MaxCollectedError -> 1
]


CucumberExpression! := (Separator | Parameter | Alternation )*
Separator! := #WHITE_SPACE
Alternation! [->] := Alternate (#ALTERNATION Alternate)*
Alternate! [->] := (Optional | Parameter | AlternateText )+
AlternateText! := #TEXT*
Optional! := #BEGIN_OPTIONAL (Text)* #END_OPTIONAL
Parameter! := #BEGIN_PARAMETER (Text)* #END_PARAMETER
Text! := (#TEXT | #WHITE_SPACE)
```

I had to add the priority to `Alternation` to the error below:

```
Generation failed.
Ambiguous rules. Provide more look-ahead hints. Branches: 
ALTERNATION -> CucumberExpression:0>__alt0:2>Alternation:1>__grp1:0>#ALTERNATION:0
ALTERNATION -> CucumberExpression:0>__alt0:2>Alternation:1>__grp1:0>#ALTERNATION:0
```

And using this grammar I run into the problem that the empty alternative does not trigger a `startRule` or `endRule` invocation on the builder. 

```
Start rule: CucumberExpression
Start rule: Alternation
Start rule: Alternate
Start rule: AlternateText
Token: a
End rule: AlternateText
End rule: Alternate
Token: /
Token: /
Start rule: Alternate
Start rule: AlternateText
Token: b
End rule: AlternateText
End rule: Alternate
End rule: Alternation
Token: 
End rule: CucumberExpression
```

I would expect it to trigger an begin/end rule because we are entering an empty alternate.

gasparnagy:
@mpkorstanje Thinking it over the today's session, sorry if I was too nonconstructive. I think the key thing I learned was that I should have been better focusing on your expectations in terms of the ast result and not so much on what I think is good. Anyway probably where we ended (that the expectations should be better preserved as tests) is good.

Also it was a good input, that you don't know how BERP works. I don't know how this could be solved the best, but here is what I have in my mind now: feel free to park it for the time when you get back to it. (I flash it out to myself partly.)

So BERP builds up a finite state machine by discovering the unique branches of the rules. The state transitions are triggered by the tokens: from a particular state you can get to another states depending on the next token that comes. This is represented with the big switch in the generated code, but when calling the parser generator with the `-d` option, it also displays it. 

This is for example the state 1 summary:
```
1: CucumberExpression:0>__alt0:0>Separator:0>#Separator:0
    EOF -> 14
    Separator -> 1
    LCurl -> 2
    LParen -> 5
    Word -> 8
    Slash -> 9

    EOF -> CucumberExpression:1>#EOF:0
        [E/Separator,P/#TOKEN]
    LCurl -> CucumberExpression:0>__alt0:1>Parameter:0>#LCurl:0
        [E/Separator,S/Parameter,P/#TOKEN]
    LParen -> CucumberExpression:0>__alt0:2>Alternation:0>Alternate:0>__alt2:0>Optional:0>#LParen:0
        [E/Separator,S/Alternation,S/Alternate,S/Optional,P/#TOKEN]
    Separator -> CucumberExpression:0>__alt0:0>Separator:0>#Separator:0
        [E/Separator,S/Separator,P/#TOKEN]
    Slash -> CucumberExpression:0>__alt0:2>Alternation:1>__grp1:0>#Slash:0
        [E/Separator,S/Alternation,P/#TOKEN]
    Word -> CucumberExpression:0>__alt0:2>Alternation:0>Alternate:0>__alt2:1>Text:0>#Word:0
        [E/Separator,S/Alternation,S/Alternate,S/Text,P/#TOKEN]
```

From that what you can see, that from state 1 the parser can get to states 1, 2,5,8,9,14 depending on the next token. (State 1 was the state where the parser got after processing a single separator.) For example when reading EOF, it gets to state 14 and during this transition it will end the `Separator` rule and process the EOF token (this you can get from the line above the transition `[E/Separator,P/#TOKEN]`). 

Getting back to our problem: if an alternate was empty, there were no tokens processed, so there will be no state transition related to that rule. This is why it does not make an empty start/end rule for that case. Of course BERP knew that the empty rule had to be considered, so probably it would be possible to convince it to generate the empty rule (this would be the change we need in BERP). (Probably we could even workaround it by reacting on the `#ALTERNATION` token in the `build` method of the AST builder. But let's not go into this before the expectations are finalized.)

mpkorstanje:
I've fixed the tokenizer:

- Throws when EOL is escaped
- Throws when regular characters are escaped

This might seem annoying but it will avoid surprising people with expressions that don't match.

I've fixed the parser:

- Mismatched braces are not ignored.
- All exceptions thrown now have the form.

```
This Cucumber Expression has problem at column 22:

three blind mice/rats(
                     ^
The '(' does not have a matching ')'. 
If you did not intended to use optional text you can use '\(' to escape the optional text
```

Still have to consider:
- Allowing non-text tokens to be treated as text tokens is nothing else matches

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/CucumberExpressionException.java
```diff
@@ -1,14 +1,115 @@
...
+    static CucumberExpressionException createMissingEndTokenException(Type beginToken, Type endToken,
+            List<Token> tokens, int current) {
+        int index = currentTokenIndex(tokens, current);
+        return new CucumberExpressionException("" +
+                thisCucumberExpressionHasAProblemAt(index) +
+                "\n" +
+                expressionOf(tokens) + "\n" +
+                pointAt(index) + "\n" +
+                "The '" + beginToken.symbol() + "' does not have a matching '" + endToken.symbol() + "'. " + "\n" +
+                "If you did not intended to use " + beginToken.purpose() + " you can use '\\" + beginToken
```
marnen, line 62: "did not intend"

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/CucumberExpressionException.java
```diff
@@ -1,14 +1,115 @@
...
+    static CucumberExpressionException createAlternativeIsEmpty(String expression, AstNode node) {
+        return new CucumberExpressionException(thisCucumberExpressionHasAProblemAt(node.start()) +
+                "\n" +
+                expression + "\n" +
+                pointAt(node.start()) + "\n" +
+                "Alternative may not be empty. If you did not mean to use an alternative you can use '\\\\' to escape the the '\\'");
+    }
+
+    private static String thisCucumberExpressionHasAProblemAt(int index) {
+        return "This Cucumber Expression has problem at column " + (index+1) + ":" + "\n";
```
mpkorstanje, line 104: `has a problem`

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/CucumberExpressionException.java
```diff
@@ -1,14 +1,115 @@
...
+                thisCucumberExpressionHasAProblemAt(index) +
+                "\n" +
+                expressionOf(tokens) + "\n" +
+                pointAt(index) + "\n" +
+                "The '" + beginToken.symbol() + "' does not have a matching '" + endToken.symbol() + "'. " + "\n" +
+                "If you did not intended to use " + beginToken.purpose() + " you can use '\\" + beginToken
+                .symbol() + "' to escape the " + beginToken.purpose() + "\n");
+    }
+
+    private static String thisPRo(int i) {
```
mpkorstanje, line 66: Junk

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/CucumberExpressionException.java
```diff
@@ -1,14 +1,115 @@
...
+            String value = argGroup.getValue();
+            list.add(value);
+        }
+        return list;
+    }
+
+    static CucumberExpressionException createMissingEndTokenException(Type beginToken, Type endToken,
+            List<Token> tokens, int current) {
+        int index = currentTokenIndex(tokens, current);
+        return new CucumberExpressionException("" +
```
mpkorstanje, line 56: This follow the pattern of pre-amble, new line, expression, pointer, problem description, suggested solution.

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/CucumberExpressionException.java
```diff
@@ -1,14 +1,115 @@
...
     }
 
     CucumberExpressionException(String message, Throwable cause) {
         super(message, cause);
     }
+
+    static CucumberExpressionException createCaptureGroupParameterTypeMisMatch(TreeRegexp treeRegexp,
+            List<ParameterType<?>> parameterTypes, List<Group> argGroups) {
+        return new CucumberExpressionException(
+                String.format("Expression /%s/ has %s capture groups (%s), but there were %s parameter types (%s)",
```
mpkorstanje, line 26: This isn't in the right format yet.

mpkorstanje:
@gasparnagy I think I've got all the parsing exceptions that I'd like to cover covered. They're all listed in `CucumberExpressionException`. The Java version should have all the tests pass. I'm currently updating the go version.

I think now would be a good time to try to implement a generated parser. It would probably be best to do this from a branch though.

gasparnagy:
@mpkorstanje super. could you maybe help with creating the branch and setting up BERP generation to the build pipeline? Don't worry if the generated code does not compile, I can take it over from that point on.

mpkorstanje:
Note to self: interesting use case of alternative here:

https://stackoverflow.com/questions/63792710/alternative-parameter-type-in-cucumber

mpkorstanje:
That was surprisingly easy. And it removed an entire error message too. 

It is now possible to use: `user makes a request for the {int}st/nd/rd/th item` .

However `{int}/none`  will still fail.

While `a/i{int}n/y` (to match i18n or a11y) is valid.

mpkorstanje:
@gasparnagy I'll give it a try. I might extract some tests out into a file first so testing the go implemetnation might become a bit easier.

mpkorstanje:
> Nice shared approval test! But why YAML and not JSON?

Better readability for exception messages:

```
expression: \
exception: |-
  This Cucumber Expression has a problem at column 2:

  \
  ^
  The end of line can not be escaped.
  You can use "\\" to escape the the "\"
```

And there is json for data structures

```
expression: mice/rats
expected: |-
  {type: "EXPRESSION_NODE", start: "0", end: "9", nodes: [
    {type: "ALTERNATION_NODE", start: "0", end: "9", nodes: [
      {type: "ALTERNATIVE_NODE", start: "0", end: "4", nodes: [
        {type: "TEXT_NODE", start: "0", end: "4", token: "mice"}
      ]},
      {type: "ALTERNATIVE_NODE", start: "5", end: "9", nodes: [
        {type: "TEXT_NODE", start: "5", end: "9", token: "rats"}
      ]}
    ]}
  ]}
```

mpkorstanje:
I dropped the stupid single quote stuff. Yaml doesn't care in a multi-line text thingy.

mpkorstanje:
Currently all the comparisons go via `toString` because I had the whole rendering already available for debugging.

mpkorstanje:
Still gotta extract the non-java specific stuff from `CucumberExpressionTest` and `CucumberExpressionPatternTest`.

aslakhellesoy:
I think we can get decent messages by diffing JSON. There are decent libraries for that. Diffing pretty-printed JSON with sorted keys also works surprisingly well.

mpkorstanje:
> I think we can get decent messages by diffing JSON. There are decent libraries for that. Diffing pretty-printed JSON with sorted keys also works surprisingly well.

The JSON representation of multi-line text blocks is atrocious. Pretty much impossible to read. 

Anyway I'm using a slightly different setup from Gherkin. Each file contains the cucumber expression and the expected output. This will allow us to run the tests as unit tests rather then integration tests. This is much easier to debug and IDEA can show me the diff. It also bypasses the whole problem of having of having to compare the files in a terminal in the first place.

mpkorstanje:
@vincent-psarga I've got no idea why `react/javascript` is now failing. I didn't change anything that relates to it as far as I can tell. And react doesn't depend on cucumber expressions at all.

mpkorstanje:
@aslakhellesoy Typoscript is done. You can do Ruby.

aslakhellesoy:
> @aslakhellesoy Typoscript is done. You can do Ruby.

Typoscript 🤣

mpkorstanje:
Note to self: Have a critical look at  the remaining tests for `CucumberExpression`. Looks like Java and Ruby both contain tests that weren't copied over to Java/Go but aren't Ruby specific. These could be added to the suite.

mpkorstanje:
Note to self: Also check the escapes in the Cucumber expression generator.

mpkorstanje:
image

mpkorstanje:
Still need to address:

> Allowing non-text tokens to be treated as text tokens if nothing else matches. Probably not a good idea because it might confuse people. However it does make escaping some patterns rather cumbersome. E.g. \(hello) vs \(hello\).

mpkorstanje:
And also need to address:
```
ILLEGAL_PARAMETER_NAME_PATTERN = Pattern.compile("([\\[\\]()$.|?*+])");
```

Parameter names are part of a cucumber expression and should should not allow reserved characters from that domain. Currently the regex domain is not allowed.

mpkorstanje:
Design choices made:

# Make the parser complain about ambiguous structures #

Because cucumber expressions are an unknown language implicit behaviour will be harder to understand. 

For example given: `three (brown/black) mice` could depending match either:

```
three brown/black mice
three mice
```

Or:

```
three mice
three brown mice
three black mice
```

The latter makes sense but cucumber expressions don't support nesting alternatives in an optional. To make sure we can add this in the future and to make sure users understand that this is not a valid Cucumber expression the parser will complain about this.

# But don't complain about shorthand

This however is perfectly fine:

```
three \(brown) mice
```

mpkorstanje:
Implementation is done. Before releasing this it would be prudent to release the other unreleased fixes first. This release should result in a new major version because the semantics of Cucumber Expressions have subtly changed.

The documentation here would need a small update w.r.t to escaping:

https://cucumber.io/docs/cucumber/cucumber-expressions/

cucumber-expressions/ruby/lib/cucumber/cucumber_expressions/cucumber_expression_generator.rb
```diff
@@ -85,19 +83,17 @@ def create_parameter_type_matchers(text)
...
+          ParameterTypeMatcher.new(parameter_type, regexp, text, 0)
         end
-        result
       end
 
       def escape(s)
         s.gsub(/%/, '%%')
         .gsub(/\(/, '\\(')
-        .gsub(/\{/, '\\{')
+        .gsub(/{/, '\\{')
```
mpkorstanje, line 96: Is `{` not a reserved character?

aslakhellesoy: It is, but whether it needs to be escaped or not depends on the context. IDEA told me escaping was redundant here, and I trust it :-)

aslakhellesoy (approved):
What a tremendous piece of work this is @mpkorstanje. Thanks a million.

Really looking forward to integrating this into Cucumbers.

laeubi:
@mpkorstanje is there any example how to use the parser "standalone"? e.g. I have an expression as a string and like to parse it withe the parser to get the different parts?

mpkorstanje:
Not really. I'd rather not expose that as an API. I'd have to polish the AST even more before I'm confident I won't change it after the first bug we find.

laeubi:
@mpkorstanje thanks for the feedback, I would be happy even with a "preliminary" API just to test and maybe give some feedback on this. I think this would is a really great addition to support tools in not firing their own (incomplete or differing) parsing algorithms.

PR #776, opened 2019-10-28
tag-expressions-go: Slice rune to string conversion is expensive

ankitpokhrel:
## Summary

Currently, we are saving token in a slice of rune and converting it to a string which is expensive. This PR replaces this with `bytes.Buffer` which is much faster.

**Before**
```
goos: darwin
goarch: amd64
expression: not a\\(\\) or b and not c or not d or e and f

BenchmarkTokenize-4   	  740218	      1571 ns/op
```

**After**
```
goos: darwin
goarch: amd64
expression: not a\\(\\) or b and not c or not d or e and f

BenchmarkTokenize-4   	 1227254	       978 ns/op
```

## Types of changes

- [x] Performance enhancement (non-breaking change)

## Checklist:

### DIFF
--- a/tag-expressions/go/parser.go
+++ b/tag-expressions/go/parser.go
@@ -1,6 +1,7 @@
 package tagexpressions
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"strings"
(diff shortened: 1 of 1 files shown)


mpkorstanje:
Looks good to me. Cheers!

PR #777, opened 2019-10-29
gherkin: Add acceptance test that fails on Ruby 2.4

aslakhellesoy:
This fails on Ruby 2.4 because protobuf on that version does not allow setting `nil` fields.

Two possible fixes:
* Upgrade protobuf
* Only set a rule's `description` if it's not `nil` in `ast_builder.rb` (we do this conditionally already for other elements)

I'd love to make this fail in CI before fixing it though!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -740,9 +740,6 @@ workflows:
       - gherkin-ruby-26:
           requires:
             - cucumber-messages-ruby
-      - gherkin-ruby-27:
-          requires:
-            - cucumber-messages-ruby
       - gherkin-jruby-92:
           requires:
             - cucumber-messages-ruby
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,11 +51,6 @@ executors:
     docker:
       - image: circleci/ruby:2.6
     working_directory: ~/cucumber
-  # JRuby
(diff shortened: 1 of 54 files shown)


aslakhellesoy:
We're currently overriding the protobuf version in the `Gemfile` for older rubies, because the latest version didn't work on them. This might have changed - something to look into.

luke-hill:
The last I checked protobuf 3.8.0 worked on Ruby 2.3 - 2.6 (Without the writer tweaks). But I'm a long way off checking where they're up to now with their specification. I know their big area at the moment is trying to find / convince someone to support their JRuby implementation as this is quite flaky.

aslakhellesoy:
I think you're right @luke-hill - we should only have to override the protobuf version for JRuby now as done here:

https://github.com/cucumber/cucumber/blob/a640a01f44bf5c2ef2ceb129d167e91416462e1e/cucumber-messages/ruby/Gemfile#L5

We should stop doing this:

https://github.com/cucumber/cucumber/blob/a640a01f44bf5c2ef2ceb129d167e91416462e1e/gherkin/ruby/Gemfile#L7-L9

We do this in cucumber-ruby and cucumber-ruby-core too IIRC, and that should go.

PR #788, opened 2019-11-04
Handle doc strings in JSON-formatter

vincent-psarga:
## Summary

The JSON-formatter is not be complete yet, the goad is to make it better (or at least more compliant with the current one in `cucumber-ruby`)

## Details

Discovered for the moment, the `doc strings` are not generated in the JSON report. I guess the `datatable` will cause the same kind of issues.

This will need more tests, but there are strong chances that messages showing steps coming from `before` and `after` will also cause issues. Still need investigations.

## Types of changes

## Checklist:

### DIFF
--- a/json-formatter/ruby/features/test_doc_strings.feature
+++ b/json-formatter/ruby/features/test_doc_strings.feature
@@ -1,0 +1,11 @@
+Feature: doc strings
+
+  Scenario: a step can have docstrings
+    When a passed step
+      """
+      This is a doc string
+      """
+    Then a failing step
+      """
+      Can also have doc strings
+      """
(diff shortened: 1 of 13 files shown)


aslakhellesoy:
Please remember to tag issues :-)

json-formatter/ruby/features/test_doc_strings.feature
```diff
@@ -0,0 +1,11 @@
+Feature: doc strings
+
+  Scenario: a step can have docstrings
+    Given a passed step with
+      """
+      This is a doc string
+      """
+    Then a failed step
```
aslakhellesoy: We need to update the stepdef to take an argument here, no?

vincent-psarga:
Humm, we also need to support the tags which are not supported in the output yet.

PR #801, opened 2019-11-22
Generate more realistic TestResult data in fake-cucumber

vincent-psarga:
## Summary

This follows #793 : as we use "real" `cucumber-expressions` to execute the steps in `fake-cucumber`, why not generate real failure messages (based on the execution stack trace) and execution time.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue).
- [X] New feature (non-breaking change which adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).

## Checklist:

- [ ] The change has been ported to Java.
- [ ] The change has been ported to Ruby.
- [x] The change has been ported to JavaScript.
- [ ] The change has been ported to Go.
- [ ] The change has been ported to .NET.
- [ ] I've added tests for my code.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #804, opened 2019-11-25
Support doc strings and data tables

aslakhellesoy:
## Summary

Support `DocString` and `DataTable` in `fake-cucumber`

## Details

For now, we just grab the `DocString#value` and the `DataTable`is transformed into a `string[][]`. This can be made pluggable later so users can provide their own converters (as Cucumber-JVM has)


vincent-psarga:
This PR has been merged into PR #801

PR #810, opened 2019-11-28
json-formatter: handle hooks

vincent-psarga:
## Summary

This is a refactoring of `json-formatter` so it outputs steps coming from hooks.

The current approach (building the JSON from the AST) does not work here, as Hooks are not present in the AST.
The idea here will be to produce data from the `testStep...` messages, then from `testCase`.

## How Has This Been Tested?

I've used the [GinkGo](http://github.com/onsi/ginkgo) framework for testing. It provides capabilities close to `rspec` like imbricating, allowing readable tests.

## Types of changes

## Checklist:


json-formatter/go/json.go
```diff
@@ -172,7 +115,7 @@ func (self *Formatter) ProcessMessages(reader gio.ReadCloser, stdout io.Writer)
 				}
 				if self.isBackgroundStep(step.Id) {
 					backgroundJsonSteps = append(backgroundJsonSteps, jsonStep)
-					background = self.lookup.LookupBrackgroundByStepId(step.Id)
+					background = self.lookup.LookupBackgroundByStepID(step.Id)
```
aslakhellesoy: I see you dropped the `ByID` suffix for some other methods. I think we can drop it here too. If we're not explicit about the key, assume it's by id.

vincent-psarga: that was the idea (assuming default look up is by ID)
This one is a bit different, we lookup by `StepID`, so that's why I kept it there

aslakhellesoy: Oh I missed that - sounds good.

aslakhellesoy:
Is there anything left to do before we can merge this @vincent-psarga ?

PR #813, opened 2019-11-29
cucumber-messages: Use pure-Ruby protobuf implementation (experimental)

mvz:
## Summary

Replace `google-protobuf` gem with `protobuf`, a pure-Ruby implementation that should work on all Ruby platforms.

## Details

- Replace `google-protobuf` dependency with `protobuf`.
- Use `protobuf` gem's custom generator.
- Update `messages.proto` to make the generator emit `Cucumber::Messages` instead of `Io::Cucumber::Messages`. This is a bit of a hack. The `protobuf` gem should use the `ruby_package` option.

## Motivation and Context

JRuby support is lacking in `google-protobuf`, and the signs are not encouraging that this will get better soon. 

## How Has This Been Tested?

I ran the specs, and I have backported this change to `cucumber-messages` 6.0.x and tested it with cucumber `4.0.0.pre.3`.

## Types of changes

## Checklist:

## Still to do

- [ ] Update the `Ndjson` specs to actually check the generated json instead of just round-tripping.
- [ ] Update `protobuf` to honor the `ruby_package` option.
- [ ] Fix the build on CircleCI.


cucumber-messages/ruby/Rakefile
```diff
@@ -1,6 +1,7 @@
 # encoding: utf-8
 require 'rubygems'
 require 'bundler'
+load 'protobuf/tasks/compile.rake'
```
aslakhellesoy: This file should not be modified. The build copies it from `.templates/ruby/Rakefile`. We may need some dynamic loading here, for examople load all files in `./rake/*.rb`

cucumber-messages/ruby/messages.proto
```diff
@@ -1,7 +1,5 @@
 syntax = "proto3";
-package io.cucumber.messages;
```
aslakhellesoy: This file should not be modified. It is copied from `../messages.proto`. Change that file instead and run `source scripts/functions.sh && rsync_files`.

Changing the `package` value will break the Java implementation because it needs generated files to go into `io.cucumber.messages`. Maybe we can use `java_package` to set that specifically for Java - I haven't tried. See:
* https://developers.google.com/protocol-buffers/docs/proto3

aslakhellesoy:
This bug must be fixed before we can use the native ruby `protobuf` library: https://github.com/ruby-protobuf/protobuf/issues/408

aslakhellesoy:
I've submitted a PR to fix the issue in ruby-protobuf: https://github.com/ruby-protobuf/protobuf/issues/408

aslakhellesoy:
I've also submitted https://github.com/ruby-protobuf/protobuf/pull/411 - another fix that is required

mvz:
Thanks for working on this, @aslakhellesoy.

PR #819, opened 2019-12-04
cucumber-messages/gherkin: Fix shaded dependencies

mpkorstanje:
## Summary

Fixes incorrectly shaded dependencies and adds a script to ensure all classes are within the module path of the jar file.

## Details

`cucumber-message` did not relocate the `google.protobuf` package.
`gherkin` shaded the `cucumber-messages` dependency.

## Types of changes

### DIFF
--- a/.templates/java/scripts/check-jar.sh
+++ b/.templates/java/scripts/check-jar.sh
@@ -7,8 +7,10 @@ set -uf -o pipefail
 
 check_jar() {
   jar="$1"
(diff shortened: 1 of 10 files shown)


aslakhellesoy:
Good stuff!

I stiull think we should run `check-jar.sh` as part of the CI build (it's currently only running during release). We can fix that in a separate commit.

mpkorstanje:
```
default: .tested
.PHONY: default

.tested: pom.xml $(JAVA_SOURCE_FILES) .deps
	mvn install
	./scripts/check-jar.sh
	touch $@
```

The `default.mk` file suggests it runs during make.

aslakhellesoy:
Ah you're right.

PR #821, opened 2019-12-09
Add targets to build the mono-repo only for a given language

vincent-psarga:
## Summary

Running `make` at the root is long and boring. This allows to build the library only for a given language, example `make javascript` will build all the libraries, but only for Javascript.

## Types of changes

- [x] Build

## Checklist:


aslakhellesoy:
Running `make javascript` fails because there is no `datatable/javascript` directory. I'm experimenting with a different approach.

PR #833, opened 2019-12-19
Add option to specify the id generator instance to use

vincent-psarga:
## Summary

Gherkin has an option `predictable-ids` to generate incremental ids instead of UUIDs (the default one). That's fine for testing Gherkin itself, but when integrating with a cucumber implementation, we need the capability to have an instance of ID generator that will be passed along all messages generation (in case of incremental, that ensures that the increment is kept all along the message generation and not restarting at zero).

## Details

You now have an `id_generator` field that can be set in the option.

## How Has This Been Tested?

Unit tests

## Types of changes

## Checklist:


gherkin/ruby/lib/gherkin/stream/parser_message_stream.rb
```diff
@@ -11,7 +11,7 @@ def initialize(paths, sources, options)
         @sources = sources
         @options = options
 
-        id_generator = id_generator_class.new
+        id_generator = options[:id_generator] || id_generator_class.new
```
aslakhellesoy: I agree it's better to pass an `id_generator` object to the constructor. With that ability, I think we should remove the `:predictable_ids` option.

Offering multiple ways to do the same thing makes the library harder to maintain and use.

vincent-psarga: makes sense, and only the gherkin binary used for the tests could instantiate that. I'll change this :)

PR #835, opened 2019-12-20
Output parameter types in message stream

aslakhellesoy:
## Summary

Output `ParameterType` messages (only for user-defined ones, not built-in ones)

## Motivation and Context

Can be used by reporting tools to build a glossary

## How Has This Been Tested?

Manually actually. The cck has a `parameter-types` feature, and I inspected the NDJSON manually


mpkorstanje:
```
message ParameterType {
  // The name is unique, so we don't need an id.
  string name = 1;
  repeated string regular_expressions = 2;
  bool prefer_for_regular_expression_match = 3;
  bool use_for_snippets = 4;
}
```

In Cucumber-JVM the parameter type is part of the glue. This means that it can be defined and redefined for each scenario (think scenario scoped glue code). This means that the assumption that the name is unique doesn't hold.

PR #845, opened 2020-01-07
Add gherkin query in Ruby

vincent-psarga:
## Summary

Like this was done in [Gherkin-JS](https://github.com/cucumber/cucumber/blob/master/gherkin/javascript/src/GherkinQuery.ts), this PR adds a simple GherkinQuery object that provides the `Location` objects of various AST nodes found in `GherkinDocument` message after parsing.

Also, it raises an exception when we iterate more than once on the messages. If we don't do so, we may end up with multiple AST node objects representing the same part of the `Gherkin` feature but with different IDs.

This object will likely grow in the future (for example for finding the original `PickleStep` location), but this should already be enough to be used in `cucumber-ruby(-core)`.
Maybe more will be added to this PR or in other ones when needed.

## How Has This Been Tested?

With Unit tests

## Types of changes

## Checklist:


gherkin/ruby/lib/gherkin/gherkin_query.rb
```diff
@@ -0,0 +1,54 @@
...
+  class Query
+    def initialize
+      @ast_node_locations = {}
+    end
+
+    def update(message)
+      update_feature(message.gherkin_document.feature) if message.gherkin_document
+    end
+
+    def get_location(ast_node_id)
```
luke-hill: are these methods "synched up" across all implementations. If they are that's cool. If not traditionally we would drop `get_` and `set_` prefixes in methods.

gherkin/ruby/lib/gherkin/errors.rb
```diff
@@ -1,5 +1,7 @@
 module Gherkin
   class ParserError < StandardError; end
+  class AstNodeNotLocatedException < Exception; end
```
luke-hill: traditionally errors in ruby all end in error (I'm aware that there are existing ones here that don't), so it's not a biggie if you don't fix

aslakhellesoy: Custom exceptions should extend from `StandardError`. [Never extend](https://thoughtbot.com/blog/rescue-standarderror-not-exception) `Exception`.

gherkin/ruby/lib/gherkin/gherkin_query.rb
aslakhellesoy: We aim to have a consistent API between implementations, but I still think we should be idiomatic. Please change to `location`.

gherkin/ruby/lib/gherkin/gherkin_query.rb
```diff
@@ -0,0 +1,54 @@
+module Gherkin
+  class Query
```
aslakhellesoy: Rename the file to `query.rb`

gherkin/ruby/lib/gherkin/gherkin_query.rb
vincent-psarga: done

gherkin/ruby/lib/gherkin/errors.rb
vincent-psarga: inheritance is done.
For the `Error` vs `Exception`, all (except `ParserError` are called 'XXXException`, so I went with the rest

gherkin/ruby/lib/gherkin/gherkin_query.rb
vincent-psarga: done

vincent-psarga:
Note: I'm giving a bit more thoughts on `#argument_location` and it may not be the way to go (it's more a hack that anything else).

I think `PickleDocString` and `PickleTable` messages should both have an `ast_node_id` field, referencing the `DocString` and `DataTable` messages (which should also have their own ids, obviously).
@aslakhellesoy that sound good for you ? I can make a second PR for that.

gherkin/ruby/lib/gherkin/errors.rb
luke-hill: Makes sense for now yeh, top work though, this stuff should get gherkin 9 ready soon :)

vincent-psarga:
> I can make a second PR for that.
which is there (https://github.com/cucumber/cucumber/pull/847) and I progress

PR #847, opened 2020-01-08
Add ids to step arguments

vincent-psarga:
## Summary

Step arguments (`DocString` and `DataTable`) should have an `id` which is referenced by `PickleStepArgument`. This allows to find the location of the original argument from a `PickleStepArgument`.

## Types of changes

## Checklist:


aslakhellesoy:
I wasn’t aware we need to find the location of a DataTable or DocString.

When is this needed?

vincent-psarga:
> When is this needed?

It is used in `cucumber-ruby-core` [`compiler#create_multiline_arg `](https://github.com/cucumber/cucumber-ruby-core/blob/2d97a5c4ee3a5357f9c0ce85c33473df6477434c/lib/cucumber/core/compiler.rb#L47)

Both `Test::DataTable` and `Test::DocString` need `Location` when created. This may be dropped later on I guess, but for now I'll try to be as compliant as possible with the existing when moving `cucumber-ruby` to gherkin 9+

vincent-psarga:
oh crap, I did not mean to merge this to `master` but planned to merge `master` into it. I fix that asap sorry ...

PR #848, opened 2020-01-09
Adds IDs to step arguments (DocString and DataTable)

vincent-psarga:
## Summary

Original PR: https://github.com/cucumber/cucumber/pull/847

So the idea is to add an `id` field to DocString and DataTable and reference them from their pickle equivalent.

## Types of changes

## Checklist:


vincent-psarga:
@aslakhellesoy - pasting answer to your question in the previous PR

> When is this needed?

It is used in `cucumber-ruby-core` [`compiler#create_multiline_arg `](https://github.com/cucumber/cucumber-ruby-core/blob/2d97a5c4ee3a5357f9c0ce85c33473df6477434c/lib/cucumber/core/compiler.rb#L47)

Both `Test::DataTable` and `Test::DocString` need `Location` when created. This may be dropped later on I guess, but for now I'll try to be as compliant as possible with the existing when moving `cucumber-ruby` to gherkin 9+

aslakhellesoy:
> Both Test::DataTable and Test::DocString need Location when created

Ok, I see that. But is that location ever accessed/used anywhere? If not, wouldn't it be simpler to just pass in a dummy location, or better - remove it entirely?

I'd like to avoid adding more code that isn't used if possible.

vincent-psarga:
it's indeed not that useful, I won't set them anymore in `cucumber-ruby`.

PR #880, opened 2020-01-30
gherkin: Fix parsing of commented tags

mpkorstanje:
## Summary

Tag parsing in gherkin suffers from four problems

1. Gherkin does not support comment anywhere but at the start of a line.
2. IDEs and syntax highlighters suggest that it is possible.
3. The Java implementation splits tags on white space rather then `@`.
4. While other languages split on `@`, tag expression do not support white space.

To illustrate:

```gherkin
Feature: Example
  @Example Tag and #Comment
  Scenario: Example
```

In Java this example is tagged with the strings `@Example`, `Tag`, `and`, `#Comment`. In other languages this would be tagged with just `@Example tag and #Comment`. 

When trying to select this scenario using the tag expression `@Example` only the Java scenario would be executed. Trying to use `@Example tag and #Comment` as a tag expression results in a parse error. This means that tags are single words.

So by first removing the comments from the tag line by splitting on `\s#` and then splitting uncommented tags on `@` we can parse tags in line with IDE expectations. Invalidating tags containing white space takes care of non-selectable tags.

## Motivation and Context

Fixes:  https://github.com/cucumber/cucumber/issues/721

## Types of changes

## Checklist:


aslakhellesoy:
To create new cross-platform acceptance tests, this is the easiest way at the moment:

1) TDD the change (like you've done in Java) in `gherkin/go`
2) Create some new `testdata` files (in this case a few `testdata/bad/*.feature` ones)
3) Regenerate the approval test data from `gherkin/go` with `GOLDEN=1 make golden`
4) `cp -R testdata ../testdata && rsync_files`

Since you started with Java, we could do all of this from `gherkin/java` if we add the same `golden` task in `gherkin/java/Makefile`

mpkorstanje:
Cheers! I'll check it out tomorrow.

gherkin/c/testdata/good/tags.feature.pickles.ndjson
```diff
@@ -3,29 +3,29 @@
     "astNodeIds": [
       "4"
     ],
-    "id": "24",
+    "id": "34",
```
mpkorstanje: @aslakhellesoy, @vincent-psarga  I don't understand why these ids changed. These scenarios didn't change position. Is there an implicit ordering going on?

gherkin/dotnet/testdata/bad/whitespace_in_tags.feature.errors.ndjson
```diff
@@ -0,0 +1,12 @@
+{
+  "attachment": {
+    "source": {
+      "location": {
+        "column": 3,
+        "line": 3
+      },
+      "uri": "testdata/bad/whitespace_in_tags.feature"
+    },
+    "text": "(3:3): A tag may not contain whitespace: @a tag containing whitespace"
```
mpkorstanje: Tag line should not be repeated here.

gherkin/c/testdata/good/tags.feature
```diff
@@ -23,3 +23,15 @@ Examples:
...
+Scenario: comments
+  Given a comment
+
+@comment_tag#2 #a comment
+Scenario: hash in tags
+  Given a comment is preceded by a space
+
+@joined_tag3@joined_tag4
+Scenario: joined tags
+  Given the @delimits tags
```
mpkorstanje: Add newline

mpkorstanje:
@luke-hill sorry. Took me like 15 minutes to do the ruby. Getting better at!

gasparnagy:
@mpkorstanje @aslakhellesoy @sebrose I run into this change wen I tried to align the .NET Gherkin parser to the others, but I don't understand it. Regarding my knowledge Gherkin was a language that only supported full line comments (so `#` could only appear at the beginning of the line, except whitespace). We can debate whether this was a good decision or not, but this is how we decided. The motivation was that we wanted the scenario authors to be able to write `#` in the step text without escaping. 

This PR changes this behavior, but only partially. It allows partial line comments, but only in tag lines. (At least based on the test files (`tags.feature`), I haven't checked the code.) 

This seems to be inconsistent for me. Is this really what we wanted?

mpkorstanje:
> Is this really what we wanted?

Yes. It's the defacto behavior of various Gherkin highlighters and parsers. IDEA, Eclipse and the syntax highlighter on Github all have this behavior:

```gherkin
Feature: example
  @Tag #comment
  Scenario: example #not a comment
```

This is confusing users (https://github.com/cucumber/cucumber/issues/721) and I feel that avoiding this confusion is more important then being consistent.

PR #890, opened 2020-02-06
UndefinedParameterType message

aslakhellesoy:
## Summary

Avoid exceptions when step definitions are defined referencing undefined parameter types.

## Details

If `ExpressionFactory` attempts to instantiate a `CucumberExpression` and that fails because the expression string references an undefined parameter type, instantiate an `UndefinedParameterTypeExpression` instead.

This is a [null-object](https://en.wikipedia.org/wiki/Null_object_pattern), which matches no strings. It can be used in step definitions, but those step definitions will never execute, because the expression matches nothing. 

## Motivation and Context

Instantiating a `CucumberExpression` with a string referencing an undefined parameter type throws an exception. This causes Cucumber to abort without running anything.

We don't want a single bad step definition to throw everything off.

We'll also need a new mechanism to notify that their expression is bad (or that they need to define a parameter type), but that is beyond the scope of this PR.

One way to prepare for that is to add a `#getUndefinedParameterTypeNames() => Set<String>` method to `Expression`. We can collect all of these and put them in Cucumber Messages somewhere - immediately or at the end of a run.

## How Has This Been Tested?

Unit tests

## Types of changes

## Checklist:


cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/UndefinedParameterTypeExpression.java
```diff
@@ -0,0 +1,28 @@
+package io.cucumber.cucumberexpressions;
+
+import java.lang.reflect.Type;
+import java.util.List;
+import java.util.regex.Pattern;
+
+public class UndefinedParameterTypeExpression implements Expression {
```
mpkorstanje:
```suggestion
final class UndefinedParameterTypeExpression implements Expression {
```

mpkorstanje: Don't forget `CucumberExpression` and `RegularExpression`.

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/ExpressionFactory.java
```diff
@@ -31,7 +31,15 @@ public Expression createExpression(String expressionString) {
...
             return new RegularExpression(Pattern.compile(m.group(1)), parameterTypeRegistry);
         }
-        return new CucumberExpression(expressionString, parameterTypeRegistry);
+        return createCucumberExpression(expressionString);
+    }
+
+    private Expression createCucumberExpression(String expressionString) {
+        try {
+            return new CucumberExpression(expressionString, parameterTypeRegistry);
+        } catch (UndefinedParameterTypeException e) {
```
mpkorstanje: We're using exceptions for control flow. Might be worth extracting some of the code in the `CucumberExpression` to the factory instead of throwing the exception and then catching it.

aslakhellesoy: Yeah I had the same thought.

aslakhellesoy:
What are yiour thoughts about capturing the undefined parameter type names so we can repoert them back to the user?

I've suggested adding a `Expression#getUndefinedParameterTypeNames() => Set<String>` that we can call from Cucumber. Do you agree with this? We can decide how to report it later.

mpkorstanje:
Not a fan. We're using a null/error object pattern here so I would expect the that information to come up when we try to use the expression. Otherwise the null object pattern isn't benefiting us. 

Did you try to spike it into a project already?

aslakhellesoy:
@mpkorstanje I am planning to work this into fake-cucumber to get a feel for it.

You're suggesting the information should come up when we try to use the expression. Can you explain what you mean by that? Are you suggesting it should come out of `#match` somehow? I don't see how that would work, and it would also cause it to happen for *each* gherkin step.

aslakhellesoy:
Yes maybe I am overthinking this. Maybe we should just try-catch around `createExpression` in Cucumber and handle it there. The exception should expose the name of the undefined parameter.

mpkorstanje:
I don't think we're getting to our goal with this.

The goals:
 1. Inform the user a step definition is using an undefined parameter type
 2. Allow scenarios that do not use this step definition to pass
 3. Disambiguate between scenarios that did not pass because of a missing step definition and undefined parameter type

While #1 and #2 are possible I think #3 is impossible. To do this we would need to know which scenario didn't pass because of the missing step definition caused by the missing parameter type. To know that we'd have to know what the parameter type would be so we could match it against the step. That results in a contradiction.

What we could do is print all the snippets and all the missing parameter type errors together. However this wouldn't really disambiguate between the two. We'd basically throw a bucket of information at the user and tell him to figure out which of the two problems we report is the real problem.

Treating it almost like a compile error like we do now does provide allow us to disambiguate and makes it obvious what the problem is.

luke-hill:
For a scenario with 4 steps, with the 3rd step undefined. What would the resulting execution be?

Would they all be skipped, or would it try execute up to the bad one?

aslakhellesoy:
> For a scenario with 4 steps, with the 3rd step undefined. What would the resulting execution be?
> 
> Would they all be skipped, or would it try execute up to the bad one?

A step definition with an undefined parameter type will not be registered. Instead, a new `UndefinedParameterType` message will be emitted. Cucumber can process this messge and print a snippet for the undefined parameter type, e.g.

```
The {airport} parameter type is not defined. You can define it with the following code:

defineParameterType({
  name: 'airport',
  regexp: /????/, // The regexp that mathes the parameter
  transformer(s: string) {
    throw new Error('TODO: return an object here')
  },
})
```

mpkorstanje:
> Cucumber can process this messge and print a snippet for the undefined parameter type, e.g.

When would this message be printed? And how do we avoid confusing the user with the message about the undefined step that will also be printed?

aslakhellesoy:
> > Cucumber can process this messge and print a snippet for the undefined parameter type, e.g.
> 
> When would this message be printed? And how do we avoid confusing the user with the message about the undefined step that will also be printed?

The message would be printed at the end, I imaging before printing step definition snippets.

The `UndefinedParameterType` message has two fields:

- `name` - the name of the undefined parameter (`airport`)
- `expression` the expression that referenced it (`{airport} is closed because of a strike`)

The gherkin step text (`CDG is closed because of a strike`) would be undefined, and a snippet would be printed. I don't think it would be easy to detect that this step has a step definition, because the step definition's expression doesn't compile until its parameter is defined.

aslakhellesoy:
Maybe we shouldn't print any step definition snippets if we're printing parameter type snippets.

mpkorstanje:
> The message would be printed at the end, I imaging before printing step definition snippets.
> Maybe we shouldn't print any step definition snippets if we're printing parameter type snippets.

In Cucumber JVM we report these snippets per scenario by throwing an exception.  This is useful because test frameworks can only report a non-passing state (e.g. skipped, failed, undefined)  by throwing an exception.

For example: https://github.com/cucumber/cucumber-jvm/blob/master/testng/src/main/java/io/cucumber/testng/TestCaseResultObserver.java#L112

Additionally IDEs will chunk messages printed to stdout and display the relevant output for each scenario separately. 

So we must have some message to mark the test as failed and any additional reporting done at the end is liable to be overlooked.

aslakhellesoy:
A step definition referencing an undefined parameter type cannot be linked back to a step (or scenario). It's a "global" error. A malformed regular expression would be similar.

For JUnit - could we generate a "surrogate" test case on the fly if we detect undefined parameter types (or malformed regular expressions)?

aslakhellesoy:
Any objections if I merge this PR? Can we continue the discussion about how to report in a new issue?

mpkorstanje:
> For JUnit - could we generate a "surrogate" test case on the fly if we detect undefined parameter types (or malformed regular expressions)?

It's not possible to dynamically generate test cases. However JUnit/TestNG test cases are hierarchical so we could be fail the root test case. This is effectively the same as stopping immediately. 

(This doesn't apply to the JUnit formatter ofcourse).

> Any objections if I merge this PR? Can we continue the discussion about how to report in a new issue?

In the current situation when faced with an undefined parameter type we immediately throw an error that instructs the user to fix the problem. No tests are executed. No reports are made. With this change we potentially gain the ability to make a report and finish the test run but we also have to deal with the undefined parameter type at some point.

The current options are:

1. Check for undefined parameter types prior to execution and stop the test. 
2. Ignore the undefined parameter type and report it together with a undefined step exceptions after all tests have been executed. 
3. Ignore the undefined parameter type and report it. Suppressing any undefined step exceptions.

Option 1 is effectively the same as the current situation. However the json and xml formatter do not have a way to report failures on suite level. Tools that consume this output will render and empty, but passing report. The ndjson formatter will be able to include this information but it would effectively be an empty report containing only a `TestRunStarted` and `TestRunFinished` event.

Option 2 is bound to confuse the user. Esp when multiple steps are undefined it will be hard to workout which steps are actually undefined and which are undefined because of the missing parameter type. Additionally compared to the current situation these messages will only happen at the end of the test suite. This makes for really poor feedback.

Option 3 is also bound to confuse the user because the undefined step may not be related to the parameter type at all. This option also doesn't play well with frameworks that report per scenario. And again we report problems after the fact.

So I'm not convinced that there is any clever reporting solution that can fix my objections so far. I would see option 1 as the least problematic but not until after we've phased out the json formatter completely.

messages/dotnet/messages.proto
```diff
@@ -68,6 +68,7 @@ message Envelope {
     StepDefinition step_definition = 25;
     Hook hook = 26;
     ParameterType parameter_type = 27;
+    UndefinedParameterType undefined_parameter_type = 28;
```
mpkorstanje: Implementation wise I would suggest adding a result object to the `TestRunFinished` event with the undefined parameter type exception. That way consumers of the ndjson can easily see why the test run failed.

mpkorstanje:
> something that can potentially be resolved but not until after we've phased out the json formatter completely.

I guess an rather ugly work around would be to make the json formatter not write any output when the `TestRunFinished`  event contains an error. :laughing:

messages/dotnet/messages.proto
mpkorstanje: That's assuming I can convince you to go for option 1.

aslakhellesoy: There may be multiple undefined parameter types, so if we bundle that into `TestRunFinished` it would have to be a `repeated` field. Maybe that's ok, but we could also send it as individual messages like the current code in this PR does. I don't think that would be a problem.

compatibility-kit/javascript/features/parameter-types/parameter-types.ts
```diff
@@ -21,3 +21,8 @@ Given('{flight} has been delayed {int} minutes', function(
   assert.equal(flight.to, 'CDG')
   assert.equal(delay, 45)
 })
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+Given('{airport} is closed because of a strike', function(airport) {
```
luke-hill: I know we have the {x} test in rspec for ruby, but do these tests need copying elsewhere, or does the CCK function like the golden tests?

messages/dotnet/messages.proto
mpkorstanje: It does make it more complex to know if a run failed or passed. In the absence of any test results consumers would now have to know that undefined parameter type messages are in fact failures and not merely a special kind of step definition.

I would also have good use case for a more generic result field in `TestRunFinished`. Undefined parameter types are but one reason why Cucumber can fail to startup. For example the DI container may fail to initialize, invalid regexes, errors while instantiating the glue, ect.

mpkorstanje: I could also use the `result`  field in `TestRunFinished` to store the results of `BeforeAll` and `AfterAll` hooks should they fail.

aslakhellesoy: The front-end should print snippets for all undefined parameter types. In order to do that, we'd have to embed all the `UndefinedParameterType` messages in `TestRunFinished`.

As you point out there can be many reasons a test run fails, and we need to be able to distinguish between them. How about this:

```protobuf
message TestRunFinished {
  bool success = 1;
  Timestamp timestamp = 2;
  // Error message. Can be a stack trace from a failed `BeforeAll` or `AfterAll`.
  // If there are undefined parameter types, the message is simply "There were undefined parameter types", and the `undefinedParameterTypes` field will have the details
  string message = 3;
  repeated UndefinedParameterType undefinedParameterTypes = 4;
}
```

It's tempting to embed `TestResult` in `TestRunFinished`, but it's really a `TestStepResult`, so I think we should rename it.

WDYT?

aslakhellesoy:
> > something that can potentially be resolved but not until after we've phased out the json formatter completely.
> 
> I guess an rather ugly work around would be to make the json and junit formatter not write any output when the `TestRunFinished` event contains an error. It would preserve current behavior. 😆

Even though we're aiming to phase out the JSON formatter in favour of messages, I think we could add a single element to the JSON to indicate a failed run, reusing the existing structure.

```json
[
  {
    "description": "Test run failed",
    "elements": {
      "description": "There were undefined parameter types",
      "steps": [
        "result": {
            "error_message": "Undefined parameter type: airport",
            "status": "failed"
          }
      ]
    }
  }
]
```

That way consumers will be able to detect an error even if nothing was run.

messages/dotnet/messages.proto
mpkorstanje: How about: 

1. Undefined Parameter Types should fail the build. So `TestRunFinished.success = false` and include a message about the missing parameter type.

2. Then we define `TestRunFinished`:

```
message TestRunFinished {
  bool success = 1;
  Timestamp timestamp = 2;
  // Error message. Can be a stack trace from a failed `BeforeAll` or `AfterAll`.
  // If there are undefined parameter types, the message is simply "There were undefined parameter types", and the `undefinedParameterTypes` field will have the details
  string message = 3;
}
```

3. `UndefinedParameterType` as a message for specialized reporters that print snippets.

mpkorstanje:
> That way consumers will be able to detect an error even if nothing was run.

:+1:

messages/dotnet/messages.proto
aslakhellesoy: 1. The error message should include a message about the undefined parameter type*s* (plural)

2. Agree!
3. Agree!

So if I simply add `message` to `TestRunFinished` we can merge this, right?

mpkorstanje: Yes! That is workable! Please make it a new major version too.

aslakhellesoy: Will do - and I'll rename `TestResult` to `TestStepResult` as well, because that's what it is.

compatibility-kit/javascript/features/parameter-types/parameter-types.ts
aslakhellesoy: The CCK defines golden `ndjson` files by running the features against `fake-cucumber`. We don't require other Cucumber implementations to produce identical files, but the files they produce should be consumable by the `json-formatter` and `html-formatter`.

PR #894, opened 2020-02-10
Handle output in json-formatter

vincent-psarga:
## Summary

JSON formatter currently does not handle `output` field in the  generated json files. This field is populated in the steps by calling `puts`(in Ruby, there must be equivalents elsewhere).

The idea here is to use a specific attachment type (`text/output`) which won't end up in the attachments.
Of course we should not ask the users to handle the output type themselves, but we should add this in the various cucumber flavours.

For example, in cucumber-ruby, we could have this the messages formatter:

```ruby
      def puts(message)
        embed(message, 'text/output')
      end
```

I'm not 100% sure of the way to use attachments for output, so any better idea is more than welcome :)

## Types of changes

## Checklist:


json-formatter/javascript/features/step_definitions/steps.ts
```diff
@@ -8,6 +8,10 @@ Given('a passed {word} with text attachment {string}', function(word: string, at
   this.attach(attachmentText, 'text/plain')
 })
 
+Given('a step outputs:', function(txt: string) {
+  this.attach(txt, 'text/output')
```
aslakhellesoy: `text/output` is not a valid media type. It should be `text/plain`. Alternatively, it should be an [extension](https://en.m.wikipedia.org/wiki/Media_type) - I suggest `text/x.cucumber.output+plain`.

vincent-psarga: sounds fine, I was definitely not convinced by `text/output` anyway, sounded too hacky :D

vincent-psarga:
Note, I'm also wondering about a good way to implement this in various languages. From what I've seen, in Ruby, we define `puts` on the JSON formatter.
In Javascript, I haven't found a way to get `output` field set in the JSON formatter.
In Java, I found `createWriteHookAction` which seems to do the same (although limited to Hooks from what I understood when glancing at the code).

What I think should be pretty useful, would be to have, for each languages, two ways to add output:
 - a generic one (so the API between all cucumber implementation would be the same), which we could call `output` (or anything better works for me)
 - a way that uses name more coupled to the language.

For example, this could be:

```ruby
class MessageFormatter
  def output(text)
    # It's called embed in cucumber-ruby but let's forget it for now
    attach(text, 'text/x.cucumber.output+plain')
  end

  def puts(text)
    output(text)
  end
end
```

```typescript
class MessageFormatter implements Formatter {
  public output(text: string): void {
    this.attach(text, 'text/x.cucumber.output+plain')
  }

  // .log as in console.log
  public log(text: string): void {
    this.output(text)
  }
} 
```

```java
class MessageFormatter {
  public void output(String text) {
    this.attach(text, "text/x.cucumber.output+plain")
  }

  // .println as in System.out.println
  public void println(String text) {
    this.output(text);
  }
}
```

aslakhellesoy:
The user isn't invoking the `attach` / `embed` / `puts` methods on the formatter instance - they don't have access to that. In Ruby/Java, the method is on the World object (`self`/`this`), and in Cucumber-JVM it is on the `scenario` object, which is passed to `Before` hooks that declare a `Scenario` parameter.

Let's discuss this on a call.

vincent-psarga:
See #897

PR #895, opened 2020-02-11
React add search bar

aslakhellesoy:
Allow filtering/searching the view


aslakhellesoy:
@bastienmalatier @vincent-psarga I've merged master into this branch and taken it for a spin. This is really neat!!!

vincent-psarga:
> @bastienmalatier @vincent-psarga I've merged master into this branch and taken it for a spin. This is really neat!!!

Cool :) do you think that could be good to go to (once the CI is fixed, I'm having a look at it now)

mpkorstanje:
I'm having a look at the CI right now.

mpkorstanje:
`cucumber-html/java` now prints an error message. It's not very detailed but it will do for now.

```
io.cucumber.messages.internal.com.google.protobuf.InvalidProtocolBufferException: Cannot find field: id in message io.cucumber.messages.GherkinDocument.Feature.Scenario.Examples
```
Looks like a field was added that isn't present in `Example`.

mpkorstanje:
`cucumber-html/java` now prints an error message. It's not very detailed but it will do for now.

```
io.cucumber.messages.internal.com.google.protobuf.InvalidProtocolBufferException: Cannot find field: id in message io.cucumber.messages.GherkinDocument.Feature.Scenario.Examples
```

@aslakhellesoy looks like a field was added that isn't present in `Example`. What is interesting is that adding a field to protobuf should be forwards/backwards compatible. It doesn't appear to be.

@aslakhellesoy  @vincent-psarga it might also be a good idea to separate the implementation of the react search bar from adding additional fields to gherkin. With 500+ changed this MR as is is completely unreadable.

aslakhellesoy:
Not sure what happened here. I'm currently cleaning up messages as per #951. Once that's released and we've updated all dependants this should be fine again.

aslakhellesoy:
Looking at the build error - I think we need to persist `~/.m2` between build jobs

mpkorstanje:
Yes. If we want to use snapshots that is required. No reason not to do it either.

vincent-psarga:
To answer your question about the performance on large docs, that'll be my next goal as soon as we have the tagearch + Textsearch operational :)

vincent-psarga:
I've added a full results from `cucumber-ruby` to see how it works. It's not too slow but it takes a bit of time to load. It's hard to have clear data as this is still loaded in story book for now.

Still in the TODO list:
 - [ ] indicate to user when the search is performed using tag expressions or plain text
 - [ ] have a slightly different highlighting for tags (so people know why it matched)
 - [ ] do not collapse displayed results when filtering
 - [ ] gather data on the memory taken when displaying large result files
 - [ ] try to make the whole thing lighter, we're getting warnings in the make process due to the overall weight of compiled code (it's not supposed to be used as it online but could make the reports hard to share via email for example)

aslakhellesoy:
Hey @vincent-psarga can we merge this to master and do the remaining todos on master or new PRs?

vincent-psarga:
I'd want some inputs on extracting `pretty` to an external lib as I did there. The reason for the extraction is to avoid loading `gherkin` in the browser where it has no use (and again, the formatter is pretty heavy now).

Otherwise I think it's ok to merge, I tried it on pretty big test data already.

PR #896, opened 2020-02-11
Json formatter: use example location for undefined steps

vincent-psarga:
## Summary

When a step is undefined and is executed from a `Scenario` that has `Examples`, the match should show the line of the example, not the one from the scenario. (as can be [seen here in the first build with the failing test](https://circleci.com/gh/cucumber/cucumber/45648?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link))

Also, I updated the `neutralise-json` script so it only update the `location` field if this one does not reference a `.feature` file (that's why we had not seen the issue yet).

Note: I still have another weird difference when running `cck` in `cucumber-ruby`, where location is something like `"location": "features/examples-tables/examples-tables.feature:24:9"`, I'll mark this PR as ready when it's fixed.

## Motivation and Context

Being able to have equivalent json when running `cucumber-ruby --format=json`and `cucumber-ruby --format=message | json-formatter`

## How Has This Been Tested?

Using existing acceptance test but removing some error shadowing in `neutralise-json`script.

## Types of changes

## Checklist:


vincent-psarga:
So I've been able to get an output from Ruby looking like this when running `features/test_examples.feature`:

```json
{
  "keyword": "Then ",
  "line": 6,
  "match": {
    "location": "features/test_examples.feature:13:6"
  },
  "name": "a undefined step",
  "result": {
    "duration": 99,
    "error_message": "some stepdef error",
    "status": "undefined"
  }
}
```
(interesting part is line 5, where match location is defined)

But in order to get this, I had to upgrade to `cucumber-ruby 4.0.0rc3`, so I guess this discrepancy was introduced in the rc releases.

To be sure, @mpkorstanje and @aslakhellesoy, in Cucumber-JVM/JS, when a step is undefined in a scenario using an example table, to you have a location looking like `<feature_path>:<line of example>` or `<feature_path>:<line of the example>:<line of the step>` ?

PR #898, opened 2020-02-12
JSON-formatter: use CCK features as acceptance tests

vincent-psarga:
## Summary

We now have nice edge cases in compatibility kit, so let's use them for validating the `json-formatter`.

There's still a few issues there:
 - [ ] generating the `all.json` (or `.ndjson`) file will be difficult, as the step `'a step throws an exception'` is defined in both `stack-traces_steps.rb`(or the TypeScript equivalent) and `hooks.rb`. That's not a problem when loading individual step definitions but proves to be problematic to run everything at once.
 - [ ] I had to disable the `Rules.feature` example, as the `cucumber-ruby` version we use to generate the golden JSON does not support the `Rule` keyword.

## Types of changes

## Checklist:


compatibility-kit/ruby/.gitignore
```diff
@@ -0,0 +1,6 @@
+*.feature
```
aslakhellesoy: When we create a new language module I think we should use the `.template` (add an `.rsync` file). If you need more specific `.gitignore`, add them in sub folders.

vincent-psarga: indeed. I kept the one when this was under `cucumber-ruby/cck`, I'll use the template one (same for default.mk)

vincent-psarga:
> So `src` could be a string, or it could be some other data-type?

In the case that raised the issue, it's a string of bytes, to be exact, this one:

```ruby
data = (0..size-1).map {|i| [i].pack('C') }.join
# "\x00\x01\x02\x03\x04\x05\x06\a\b\t"
```

But the problem could still arise if someone wants to ember/attach a string that would raise the error when testing `File.file?(src)`

PR #916, opened 2020-02-28
build: Install npm as user

mpkorstanje:
## Summary

The docker file installs NPM as the root user. This results in file permissions problems when using `npm-link-shared `. By changing the global installation directory and reinstalling npm as a user we can avoid this problem.

## Types of changes

- [X] Build fix (non-breaking change which fixes an issue).


aslakhellesoy:
Great stuff!!!

Issue #658, opened 2019-07-19
Cucumber forces Integer type on a \d+ capture group


yktoo:
#  Summary

The compatibility with previous version(s) is broken.

A step definition that uses `(\d+)` capture group forces the method have an `Integer` (or `int`) argument. Previously a `String` would do fine.

Example code:
```java
@When("^I type (\\d+) into input field$")
public void register(String value) {
...
```

## Expected Behavior

The typed value is converted into a `String`, just like it was previously (tested with info.cukes.cucumber-java v1.2.5)

## Current Behavior

An exception is thrown:
```
cucumber.runtime.CucumberException: Failed to invoke nl.ns.cucumber.LoketdienstSteps.register(String) in file:[...path...], caused by java.lang.IllegalArgumentException: argument type mismatch
	at cucumber.runtime.Utils$1.call(Utils.java:29)
	at cucumber.runtime.Timeout.timeout(Timeout.java:16)
	at cucumber.runtime.Utils.invoke(Utils.java:20)
	at cucumber.runtime.java.JavaStepDefinition.execute(JavaStepDefinition.java:57)
	at cucumber.runner.PickleStepDefinitionMatch.runStep(PickleStepDefinitionMatch.java:50)
	at cucumber.runner.TestStep.executeStep(TestStep.java:65)
	at cucumber.runner.TestStep.run(TestStep.java:50)
	at cucumber.runner.PickleStepTestStep.run(PickleStepTestStep.java:43)
	at cucumber.runner.TestCase.run(TestCase.java:46)
	at cucumber.runner.Runner.runPickle(Runner.java:50)
	at cucumber.runtime.Runtime$1.run(Runtime.java:104)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at cucumber.runtime.Runtime$SameThreadExecutorService.execute(Runtime.java:258)
	at java.util.concurrent.AbstractExecutorService.submit(AbstractExecutorService.java:112)
	at cucumber.runtime.Runtime.run(Runtime.java:101)
	at io.cucumber.core.cli.Main.run(Main.java:43)
	at io.cucumber.core.cli.Main.main(Main.java:14)
Caused by: java.lang.IllegalArgumentException: argument type mismatch
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

Issue #153, opened 2017-03-16
Python package 4.1.0 broken


aslakhellesoy:
_From @koterpillar on March 16, 2017 22:19_

`gherkin-official==4.1.0` package was released to PyPI but is missing `gherkin-languages.json`. The example from README crashes immediately:

```
In [1]: from gherkin.parser import Parser
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-1-f036580b63e7> in <module>()
----> 1 from gherkin.parser import Parser

.../lib/python3.5/site-packages/gherkin/parser.py in <module>()
      2 from collections import deque
      3 from .ast_builder import AstBuilder
----> 4 from .token_matcher import TokenMatcher
      5 from .token_scanner import TokenScanner
      6 from .errors import UnexpectedEOFException, UnexpectedTokenException, ParserException, CompositeParserException

.../lib/python3.5/site-packages/gherkin/token_matcher.py in <module>()
      1 import re
----> 2 from .dialect import Dialect
      3 from .errors import NoSuchLanguageException
      4
      5

.../lib/python3.5/site-packages/gherkin/dialect.py in <module>()
      8     'gherkin-languages.json')
      9
---> 10 with io.open(DIALECT_FILE_PATH, 'r', encoding='utf-8') as file:
     11     DIALECTS = json.load(file)
     12

FileNotFoundError: [Errno 2] No such file or directory: '.../lib/python3.5/site-packages/gherkin/gherkin-languages.json'
```

_Copied from original issue: cucumber/gherkin#260_

aslakhellesoy:
Sorry about that! I just release 4.1.1 which should fix this. Please confirm.

davmelikyan:
Thanks! Now it works.

koterpillar:
Confirmed fixed.

PR #918, opened 2020-03-02
cucumber-expressions: Use default transformer in preference to contradicting type hint

scolej:
## Summary

Preliminary PR for discussion.

Don't use a matching paramater type if it's type does not match the type hint from the step-definition. Instead, create an anonymous parameter type which uses the default transformer.

Related to #658, but extended to custom parameter types.

## Details

Using a parameter type which transforms to a type which does not match the step definition's argument type will throw an exception when the glue is invoked.

If we hit this case, prefer to disregard the matching parameter type and just convert via the default transformer.

Note that this change seems to make `useRegexpMatchAsStrongTypeHint` on `ParameterType` redundant. The work to remove it has not been performed here. If the changes look ok I can do this too.

## Some details on tests:

The combination of type hints and parameter types can be expanded as follows. I've tried to match each case to a new or existing test:

1. No method type hint & no custom ParameterType => matches_positive_int
2. Method type hint & no custom ParameterType => matches_positive_int_with_hint
3. No method type hint & ParameterType => uses_parameter_type_registry_when_parameter_type_is_defined
4. Method type hint & ParameterType which agrees => uses_parameter_type_when_type_hint_agrees
5. Method type hint & ParameterType which contradicts => uses_default_transformer_when_type_hint_conflicts_with_parameter_type

The changes here should only affect case 5.


mpkorstanje:
On the mailing list you mentioned that you would like to declare multiple parameter types for qouted strings and have the step definitions find the right transformer by looking at the type hint.

For example:

``java
@ParameterType("\"([^\"]+)\"")
public TypeA typeA(String){
    return // create type A here
}

@ParameterType("\"([^\"]+)\"")
public TypeB typeB(String){
    return // create typeB here
}
```

So when given two step definitions the argument will be transformed by matching via the type rather then the regex.

```
@Given("an object of {typeA}")
public void an_object_of_type_a(TypeA a){

}
@Given("an object of {typeB}")
public void an_object_of_type_b(TypeB b){

}
```

I currently don't see this reflected in any test case and so I don't understand how moving the `parameterType.getType().equals(typeHint)` condition could achieve this. This will result in Cucumber using the default parameter type transformer rather then the registered transformers for `TypeA` or `TypeB` . You created the impression that you were looking to achieve the latter.

Did I understand your correctly?

**Comment by scolej** [comment]
Hi Rien,

I think I've given mixed signals on what I'm trying to achieve as I've slowly wrapped my head around it :)

I'm trying to get to a point where `@ParameterType`s which conflict with the type-hint on regular-expression step-definitions are ignored.

I'm not trying to change the behaviour for Cucumber-expressions at all, and I'm not trying to look up `@ParameterType`s for regular-expression glue based on type-hints.

The move of `!parameterType.getType().equals(typeHint)` is incidental; the important bit is the removal of the `... && !parameterType.useRegexpMatchAsStrongTypeHint()` part, which allows a parameter type to break regular-expression glue by forcing a type incompatible with the definition.

To run with your example...

The following glue:

```java
package cucscratch;

import io.cucumber.java.ParameterType;
import io.cucumber.java.en.Given;

public class Main {

    private static class TypeA {}

    @ParameterType("\\w+")
    public TypeA typeA(String a) {
        return new TypeA();
    }

    @Given("a cucumber expression of {typeA}")
    public void an_object_of_type_a(TypeA a) {
        // works fine!
    }

    @Given("^a regular expression of (\\w+)$")
    public void an_object_of_type_b(String b) {
        // broken: will attempt to use the transformer for TypeA
    }

}
```

plus this feature:

```cucumber
Feature:
  Scenario:
    Given a cucumber expression of type_a
    Given a regular expression of just_a_string
```

fails with:

```
io.cucumber.core.exception.CucumberException: Could not invoke step [^a regular expression of (\w+)$] defined at 'cucscratch.Main.an_object_of_type_b(java.lang.String)'.
It appears there was a problem with the step definition.
The converted arguments types were (cucscratch.Main$TypeA)

The details are in the stacktrace below.
	at io.cucumber.core.runner.PickleStepDefinitionMatch.couldNotInvokeStep(PickleStepDefinitionMatch.java:81)
	at io.cucumber.core.runner.PickleStepDefinitionMatch.runStep(PickleStepDefinitionMatch.java:65)
	at io.cucumber.core.runner.TestStep.executeStep(TestStep.java:64)
	at io.cucumber.core.runner.TestStep.run(TestStep.java:49)
	at io.cucumber.core.runner.PickleStepTestStep.run(PickleStepTestStep.java:46)
	at io.cucumber.core.runner.TestCase.run(TestCase.java:51)
	at io.cucumber.core.runner.Runner.runPickle(Runner.java:66)
	at io.cucumber.core.runtime.Runtime.lambda$run$2(Runtime.java:100)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at io.cucumber.core.runtime.Runtime$SameThreadExecutorService.execute(Runtime.java:243)
	at java.util.concurrent.AbstractExecutorService.submit(AbstractExecutorService.java:112)
	at io.cucumber.core.runtime.Runtime.lambda$run$3(Runtime.java:100)
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
	at java.util.stream.SliceOps$1$1.accept(SliceOps.java:204)
	at java.util.ArrayList$ArrayListSpliterator.tryAdvance(ArrayList.java:1351)
	at java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:126)
	at java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:498)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:485)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
	at io.cucumber.core.runtime.Runtime.run(Runtime.java:101)
	at io.cucumber.core.cli.Main.run(Main.java:73)
	at io.cucumber.core.cli.Main.main(Main.java:31)
Caused by: io.cucumber.core.backend.CucumberBackendException: Failed to invoke public void cucscratch.Main.an_object_of_type_b(java.lang.String)
	at io.cucumber.java.Invoker.invoke(Invoker.java:29)
	at io.cucumber.java.JavaStepDefinition.execute(JavaStepDefinition.java:27)
	at io.cucumber.core.runner.PickleStepDefinitionMatch.runStep(PickleStepDefinitionMatch.java:63)
	... 24 more
Caused by: java.lang.IllegalArgumentException: argument type mismatch
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at io.cucumber.java.Invoker.invoke(Invoker.java:27)
	... 26 more
```

This behaviour makes it difficult to write a mixture of regular-expression glue and cucumber-expression glue with `@ParameterType`s, without the `@ParameterType`s stomping all over the regular-expression glue.

---

I initially thought it was about multiple matching regexes but I had the wrong idea.

The following works just fine:

```java
package cucscratch;

import io.cucumber.java.ParameterType;
import io.cucumber.java.en.Given;

public class Main {

    private static class TypeA {}

    private static class TypeB {}

    @ParameterType("\\w+")
    public TypeA typeA(String a) {
        return new TypeA();
    }

    @ParameterType("\\w+")
    public TypeB typeB(String a) {
        return new TypeB();
    }

    @Given("a cucumber expression with type A: {typeA}")
    public void an_object_of_type_a(TypeA a) {
        // works fine!
    }

    @Given("a cucumber expression with type B: {typeB}")
    public void an_object_of_type_a(TypeB a) {
        // works fine!
    }

}
```

```cucumber
Feature:
  Scenario:
    Given a cucumber expression with type A: type_a
    Given a cucumber expression with type B: type_b
```

---

Hopefully this makes a bit more sense. Thanks for spending the time to review this!

mpkorstanje:
Right. I get it now! The examples really helped.

I totally missed the fact that you're also no longer calling `parameterType.useRegexpMatchAsStrongTypeHint()`. So effectively the regexp is no longer used as type hint. 

[I knew this was a problem when I introduced it](https://github.com/cucumber/cucumber/blob/master/cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/ParameterType.java#L90) :laughing: 

And the problem is that I can't change it now for the same reason I couldn't change it then without breaking existing expectations in #153 

However we could provide a `@ParameterType(useRegexpMatchAsStrongTypeHint=true/false)` option in `cucumber-jvm` that passes this value onto `ParameterType` in `cucumber-expressions`. It would initially default to `true` but at the next major release we could flip the default to `false`.

mpkorstanje:
We need to resolve one more thing:

```
public class Main {

    private static class TypeA {}

    private static class TypeB {}

    @ParameterType("\\w+")
    public TypeA typeA(String a) {
        return new TypeA();
    }

    @ParameterType("\\w+")
    public TypeB typeB(String a) {
        return new TypeB();
    }

    @Given("^a regular expression of (\\w+)$")
    public void an_object_of_type_a(TypeA a) {
        /
    }

    @Given("^an aother egular expression of (\\w+)$")
    public void an_object_of_type_b(TypeB b) {
        
    }

}
```

Rather then using `lookupByRegexp` we should add `lookupByRegexpAndType` when a type hint is available.

scolej:
> we could provide a `@ParameterType(useRegexpMatchAsStrongTypeHint=true/false` ...

This would be very useful, and flipping the default to `false` at next major release sounds like a good plan.

Thanks for your time!

mpkorstanje:
Resolved by https://github.com/cucumber/cucumber-jvm/commit/0136079aa07e5bb14e6d1bdc9afb2d09d5da7570

PR #919, opened 2020-03-02
Remove coveralls for Ruby gems

vincent-psarga:
It is barely used on causes installation issues on Windows due to the non-native JSON gem.


vincent-psarga:
@aslakhellesoy - there was a bit more to do there finally, I had to add `ruby-json` to the Docker image (I guess the Alpine ruby installation is bare as possible).
Do you think that'd be ok ?

PR #931, opened 2020-03-11
Ruby html-formatter

vincent-psarga:
Create `html-formatter` for Ruby, based on what has been done for Java and Javascript.


html-formatter/ruby/assets/.gitignore
```diff
@@ -0,0 +1,3 @@
+*.html
+*.js
+*.css
```
mpkorstanje: Are these removed again when `make clean` is used?

html-formatter/ruby/assets/index.mustache.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+	<title>Cucumber</title>
```
mpkorstanje: This file is duplicated 3 times over now. Perhaps we should copy it from the `html-formatter` directory.

html-formatter/ruby/assets/.gitignore
vincent-psarga: now they should be, thanks for pointing that out :)

html-formatter/ruby/assets/index.mustache.html
vincent-psarga: It should have been removed with commit 14a6e12, but it did not. Normally that's ok now.

html-formatter/ruby/Makefile
```diff
@@ -0,0 +1,42 @@
...
+assets/cucumber-html.js:
+	mkdir -p $(@D)
+	cp ../javascript/dist/main.js $@
+
+assets/index.mustache.html:
+	mkdir -p $(@D)
+	cp ../javascript/dist/src/index.mustache.html $@
+
+
+clean:
```
mpkorstanje: Does this override the `clean` directive in the included make file or does it extend it?

vincent-psarga: it extends it:

```shell
╭─vincent.pretre@Vincents-MacBook-Pro ~/dev/open-source/cucumber/html-formatter/ruby ‹ruby-2.6.3› ‹html-formatter-for-ruby› 
╰─$ make clean
rm -f .deps .linked .tested* Gemfile.lock *.gem
rm -rf acceptance features/all_green.feature features/statuses.feature assets/*.css assets/*.js assets/*.html
```

First line comes from `default.mk`, second one from `Makefile`

mpkorstanje: Cool. Still a bit hazy on how includes actually work.

PR #935, opened 2020-03-13
Move progress formatter to the mono-repo

vincent-psarga:
## Summary

The formatter should all be in the mono-repo to ensure consistency of behaviour between the different implementations of Cucumber.
Also they should consume messages. 

## Types of changes

## Checklist:


vincent-psarga:
I'm thinking that having a library per formatter may not be such a good idea. What would you think to have instead a `cucumber/console-formatters` library that would handle formats the following format ?
 - `progress`
 - `pretty`
 - `summary` 

Maybe the `usage` formatter could all belong there.

mpkorstanje:
> I'm thinking that having a library per formatter may not be such a good idea.

Why not?

vincent-psarga:
> Why not?

The different terminal formatter (progress, pretty and summary) hare lot of common code (handling ANSI colors, printing the summary at the bottom).
It would be simpler to handle releases of a single `console-formatter` than of 3 or 4 different libraries.

mpkorstanje:
Sounds good to me.

mattwynne:
@vincent-psarga do you have any recollection of what's left to do on this?

vincent-psarga:
> @vincent-psarga do you have any recollection of what's left to do on this?

I'd say more or less everything :D
 - port to other languages
 - add tests based on messages to ensure every language behaves the same way

But the main issue here was that the progress formatter relies heavily on common code with other formatters, so as I mentioned earlier, having simply a "progress-formatter" library might not be such a good idea.

mattwynne:
I think it would be fine to start by moving just cucumber-ruby's progress formatter into a separate gem in the common repo, and we could gradually migrate other formatters and languages in other PRs, once we've figured out a good way to make a shared test suite.

No need to try and boil the ocean; that's how things get stuck 😁

vincent-psarga:
> No need to try and boil the ocean; that's how things get stuck 😁

Indeed :)

aurelien-reeves:
Well, we are actually splitting the mono-repo now. We did not felt that happy with it. So, closing this PR

However we could still have a polyglot repo for formatters like the progress formatter

PR #938, opened 2020-03-18
Translate legacy Json to messages

vincent-psarga:
Add a Typescript library to handle legacy JSON and translate it to `message` format.

TODO:
 - [ ] ~~handle matches~~ (not available in JSON reports)
 - [x] handle `duration` of results 
 - [x] handle stack traces correctly
 - [x] ~~fix problem with `html-formatter` on example `acceptance/parameters/parameters.ndjson.html`(note that passing the `ndjson` files to the json-formatter produces good results~~ (see #953)
 - [x] detect JSON format

Add support for:
 - [x] `cucumber-ruby` json reports
 - [x] `cucumber-js` json reports
 - [x] `cucumber-java` json reports
 - [x] `behave` json reports


aslakhellesoy:
🤩🤩🤩🤩🤩🤩🤩🤩🤩 this looks promising!!!! Can't wait to dive in.

aslakhellesoy:
Looks great! Overall changes we should make:

* [ ] Improve streaming
  * [ ] Use `pipe` rather than stream event handlers
  * [ ] Move more code from `cli` to `index` so this can be used as a lib
  * [ ] cli only parses options and passes `stdout` and `stdin` to a lib function
  * [ ] Don't buffer built messages, write them immediately
* [ ] Define a TypeScript schema for ruby json, js json, jvm json etc.
* [ ] Change the `parse` method to `process` or `traverse` and don't return anything
  * [ ] Populates supportCode
  * [ ] Write `GherkinDocument` messages on the fly to the same stream we're passing later to `runCucumber`
* [ ] Refactor towards a more functional design for the parser
  * [ ] Will make it easier to reuse functions that are the same for js/ruby/java/etc
  * [ ] Easier to test as well (no need to pass a full JSON doc)

PR #947, opened 2020-03-30
Store attachment body as text, with content-encoding base64 or identity

aslakhellesoy:
## Summary

Previously an `Attachment` could have a `bytes binary` or `string text` body. Now it just has a `string body` and an additional `content_encoding` field indicating whether or not it is `base64` encoded. 

## Details

This is inspired by the [Content-Encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding) HTTP header.

## Motivation and Context

When `@cucumber/react` renders image attachments, it expected to find the image data in the `binary` field (as bytes). It then ran it through [btoa](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/btoa) to convert it to `base64` and create an image tag like ``.

Node.js doesn't have a `btoa` function (only the browser does), but it can be implemented in a single line of code: `const btoa = s => Buffer.from(s).toString('base64')`. While this generates a valid base64 encoding of the data, it is different from what the browser's `btoa` does.

This would cause React warnings in applications that do both server-side-rendering (in Node) and client-side rendering (in the browser).

Moreover, this approach introduced needless conversions:
* from binary to base64 when a message is JSON-encoded
* from base64 to binary when a message is JSON-decoded
* from binary to base64 again when a message is rendered as an ``

The new approach stores binary data as a base64 encoded string in the message (indicated by a new `content_encoding` field). There is no more conversion during JSON serialisation/deserialisation, and when an attachment is rendered as an `` the raw base64 string can be used directly without the need for a `btoa`.

So this is more performant and avoids React errors. It does incur larger binary-encoded protobuf messages (binary is more efficient), but there is no size in change for JSON-encoded messages, which we're always using since we decided to always use ndjson.

## Types of changes

## Checklist:


aslakhellesoy:
The CCK is failing because of the cucumber-ruby dependency, which uses the previous version of cucumber-messages. I think we may have to disable the ruby part of the CCK to make this pass, and re-enable it again after this is released and cucumber-ruby can upgrade to the new version.

vincent-psarga:
> The CCK is failing because of the cucumber-ruby dependency, which uses the previous version of cucumber-messages. I think we may have to disable the ruby part of the CCK to make this pass, and re-enable it again after this is released and cucumber-ruby can upgrade to the new version.

Works for me. This part of the CI process is merely to ensure the released version of `cucumber-ruby` complies with the CCK (at least when released).

PR #950, opened 2020-04-02
Various work to optimise space taken by node modules

vincent-psarga:
Try to avoid eating whole hard drive with `node_modules`:

 - ignore lock files generated by `pnpm`

to be continued :)


vincent-psarga:
The main issue with the space taken by `node_modules` was due to the usage of `yarn`. Using `npm` seems to work fine again, so let's close this.

PR #953, opened 2020-04-08
React - do not use line number to get Step or Row status

vincent-psarga:
That's a "bug" I discovered when adding `json-to-messages`(#938).
We use the line number of the step to find the corresponding `PickleStep` (and same goes to finds the `Pickle` linked to an example row).

In normal use cases, this should not be an issue, but in `json-to-messages` we end up with multiple steps being on the same line, causing rendering issues.
And as we've introduced IDs to easily find our way between messages, let's use them.


aslakhellesoy:
I think perhaps this PR should also remove:

* `getLocation(astNodeId: string): messages.ILocation`
* `getPickleIds(uri: string, lineNumber?: number): ReadonlyArray<string>`
* `getPickleStepIds(uri: string, lineNumber: number): ReadonlyArray<string>`

WDYT @vincent-psarga?

vincent-psarga:
I did not remove them to ensure backward compatibility, but not sure how much we need it.
That said, `getPickleIds(uri)` is still used to find all pickles from one Gherkin document.

aslakhellesoy:
I think we need to change `getPickleIds(uri: string, lineNumber?: number)` to `getPickleIds(uri: string, astNodeId?: string)`. If we do that we're no longer depending on line numbers.

vincent-psarga:
> I think perhaps this PR should also remove:
> 
> * `getLocation(astNodeId: string): messages.ILocation`

I did some more checks, this one is also used in `fake-cucumber` to create a source frame for the Hook steps, so I'll leave it ass it is now.

I'm updating the other one now and dropping the location of node by line.

vincent-psarga:
@aslakhellesoy does it look good for you now with the new change and removing the old lookup by lines ?

PR #956, opened 2020-04-09
build: Install behave in docker

mpkorstanje:
Installs behave in docker. Avoids installing it on the host system.

TODO:
 - [x] Publish the docker image
 - [x] Update Circle CI with latest docker image
 - [x] Merge and rebase on https://github.com/cucumber/cucumber/pull/955


mpkorstanje:
@aslakhellesoy are the credentials for dockerhub in Keybase still up to date?

```
mpkorstanje@Nyx:~/Projects/cucumber/cucumber$ docker login --username cukebot
Password: 
Error response from daemon: Get https://registry-1.docker.io/v2/: unauthorized: incorrect username or password
```

aslakhellesoy:
@mpkorstanje I don't think I've changed credentials. I'll verify.

aslakhellesoy:
@mpkorstanje worked for me:

```
# load credentials
source ../secrets/secrets.sh
source scripts/functions.sh
docker_push Dockerfile
```

mpkorstanje:
Looks like 1Password and Keybase are not in sync. 1Password has the correct version. Did you push the encrypted repo?

aslakhellesoy:
Yes, everything is pushed.

1Password has an incorrect password. The `scripts/functions.sh` script loads the [git-crypt](https://github.com/AGWA/git-crypt) encrypted password from `./secrets/.bash_profile`:

https://github.com/cucumber/cucumber/blob/cc2a9321af8cdb810c5a39c9396e95541b1431ec/scripts/functions.sh#L66

If you haven't decrypted the repo with `git-crypt` the `docker_push` shell function won't work.

This file contains the same `DOCKER_HUB_PASSWORD` as `/secrets.sh` in the `keybase://team/cucumberbdd/secrets` repo.

To avoid further confusion I think we should have a single source of truth (Keybase):

* Remove all secrets from 1Password
* Remove the `/secrets` directory and stop using `git-crypt`
* Use `keybase://team/cucumberbdd/secrets` to release all our repos. It will have to be checked out to a single place on the machine, and its location must be defined by a `CUCUMBER_SECRETS_DIR` env var.

aslakhellesoy:
My bad, you're right. The 1Password one is correct. I assumed the one on file was correct since I was allowed to push. That must have been since I was already logged in.

I still think we should migrate everything to Keybase. I'll push the updated password on keybase.

mpkorstanje:
> I still think we should migrate everything to Keybase. 

Lets make an item for that. Getting the keybase script to work inside the docker isn't great because it isn't mounted.

aslakhellesoy:
Easy - the `docker_run` script can mount `$CUCUMBER_SECRETS_DIR` as a volume

PR #964, opened 2020-04-10
Support more media types and support both base64 and identity for text

### LINKED ISSUE #945: cck: Improve attachment message
Currently the Java implementation has two issues with the CCK.

1. When provided with the `mediaType` `text/plain` the reference implementation will embed the contents as text. This seems rather unnecessary as there is a dedicated log method to do this. Changing the reference implementation to embed always write attachments as binary would simplify both the reference implementation and Cucumber-JVM.

2. Cucumber-JVM supports file names in attachments. The reference implementation currently does not.

```java
    @Override
    public void embed(byte[] data, String mediaType, String name) {
        bus.send(new EmbedEvent(bus.getInstant(), testCase, data, mediaType, name));

        // TODO: Remove this weird exception from the cck
        if (mediaType.equals("text/plain")) {
            bus.send(Messages.Envelope.newBuilder()
                .setAttachment(
                    Messages.Attachment.newBuilder()
                        .setTestCaseStartedId(testExecutionId.toString())
                        .setTestStepId(currentTestStepId.toString())
                        .setText(new String(data, UTF_8))
                        //TODO: Add file name to message protocol
                        .setMediaType(mediaType)
                        .build()
                )
                .build()
            );
        } else {
            bus.send(Messages.Envelope.newBuilder()
                .setAttachment(
                    Messages.Attachment.newBuilder()
                        .setTestCaseStartedId(testExecutionId.toString())
                        .setTestStepId(currentTestStepId.toString())
                        .setBinary(ByteString.copyFrom(data))
                        //TODO: Add file name to message protocol
                        .setMediaType(mediaType)
                        .build()
                )
                .build()
            );
        }
    }
```

aslakhellesoy:
This PR adds support for more media types as well as supporing both base64 and identity content encoding. Fixes #945.


mpkorstanje:
Nice! This puts us on par with the existing reports.

PR #970, opened 2020-04-14
Re-enable html-formatter/ruby

aslakhellesoy:
It was disabled in #955. That issue says `html-formatter` depends on the `compatibility-kit`, but it isn't obvious to me in what way. @mpkorstanje can you elaborate?


mpkorstanje:
The ruby implementation of the html formatter uses the CCK messages to generate the acceptance report. I'm not sure if that was intentional. The other two use use fake cucumber.

https://github.com/cucumber/cucumber/blob/master/html-formatter/ruby/Makefile#L7

aslakhellesoy:
Thanks for the explanation @mpkorstanje. @vincent-psarga - could you have a look into this? I've released all the monorepo stuff up until `html-formatter` 5.0.0. We need to make another release of `html-formatter` 5.0.1 to make sure the ruby implementation is released (and then release the remaining libs). Otherwise we won't have a cucumber-ruby that works with the latest messages 12.0.0 and gherkin 13.0.0.

aslakhellesoy:
Actually, I have a better idea how to fix this, so I'll impolement it

PR #971, opened 2020-04-15
Extract jsonToMessages function so this can be used as a library

aslakhellesoy:
## Summary

Make it possible to use this as a library.

## Details

- [x] Extract a new `jsonToMessages` function
- [x] Create `src/index.ts` that exposes that function
- [x] Rename "platform" to "implementation"
- [x] Rename "javascript" to "cucumber-js"
- [x] Rename "ruby" to "cucumber-ruby"

## Motivation and Context

When this is available we can write a webapp using the function. That way this can be deployed as a microservice somewhere, making it easier to consume.

### DIFF
--- a/json-to-messages/javascript/src/JSONTraverse.ts
+++ b/json-to-messages/javascript/src/JSONTraverse.ts
@@ -17,7 +17,7 @@ export default function traverseFeature(
   newId: IdGenerator.NewId,
   predictableSupportCode: IPredictableSupportCode
 ): messages.IGherkinDocument {
-  if (platform === 'javascript') {
+  if (platform === 'cucumber-js') {
     return traverseJS(
       feature as IFeature,
       astMaker,
--- a/json-to-messages/javascript/src/JSONTraverse.ts
+++ b/json-to-messages/javascript/src/JSONTraverse.ts
@@ -8,15 +8,16 @@ import { traverseFeature as traverseBehaveFeature } from './behave/JSONTraverse'
 import IPredictableSupportCode from './IPredictableSupportCode'
 import IAstMaker from './IAstMaker'
 import { IdGenerator, messages } from '@cucumber/messages'
+import { Platform } from './types'
 
 export default function traverseFeature(
-  sourceLang: string,
+  platform: Platform,
   feature: IFeature | IBehaveFeature,
   astMaker: IAstMaker,
   newId: IdGenerator.NewId,
   predictableSupportCode: IPredictableSupportCode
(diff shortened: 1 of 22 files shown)


json-to-messages/javascript/src/JSONTraverse.ts
```diff
@@ -8,15 +8,16 @@ import { traverseFeature as traverseBehaveFeature } from './behave/JSONTraverse'
 import IPredictableSupportCode from './IPredictableSupportCode'
 import IAstMaker from './IAstMaker'
 import { IdGenerator, messages } from '@cucumber/messages'
+import { Platform } from './types'
 
 export default function traverseFeature(
-  sourceLang: string,
+  platform: Platform,
```
mpkorstanje, line 14: Platform typically means Windows/Unix/OSx. Would implementation be a better word?

json-to-messages/javascript/src/JSONTraverse.ts
```diff
@@ -8,15 +8,16 @@ import { traverseFeature as traverseBehaveFeature } from './behave/JSONTraverse'
...
   feature: IFeature | IBehaveFeature,
   astMaker: IAstMaker,
   newId: IdGenerator.NewId,
   predictableSupportCode: IPredictableSupportCode
 ): messages.IGherkinDocument {
-  if (sourceLang === 'javascript') {
+  if (platform === 'javascript') {
```
mpkorstanje, line 20: I could use the java impl to generate messages that would be classed as JavaScript. 

Perhaps gherkin and behave would be better.

json-to-messages/javascript/src/JSONTraverse.ts
mpkorstanje: Or dialect?

json-to-messages/javascript/src/JSONTraverse.ts
vincent-psarga: Well, it' not as simple a Gherkin vs Behave (Behave uses Gherkin, not the library but at leat the language).

Here we mainly detect based on specificities of the existing reports. If fact, we don't even have a `cucumber-java` platform/language/dialect as the JSON reports are close enough to `cucumber-ruby`, so we use the same object to traverse.

json-to-messages/javascript/src/JSONTraverse.ts
aslakhellesoy: Yes, I think `implementation` is better. Let's change that too before merging this (adding a TODO above)

PR #975, opened 2020-04-20
Display failed hook in React

bastienmalatier:
#411  Summary

If we have a failed hook in react, we don't see it in the current html report. With this modifications, we want them to appear in it.

## Types of changes

## Checklist:


query/javascript/src/Query.ts
```diff
@@ -157,4 +167,51 @@ export default class Query {
     | undefined {
     return this.stepMatchArgumentsListsByPickleStepId.get(pickleStepId)
   }
+
+  public getBeforeHookSteps(
+    pickleId: string
+  ): ReadonlyArray<messages.TestCase.ITestStep> {
+    const testCase = this.testCaseByPickleId.get(pickleId)
+
+    // const hookSteps = testCase.testSteps.filter((step) => step.hookId)
```
aslakhellesoy, line 176: Remove commented-out code

query/javascript/src/Query.ts
```diff
@@ -157,4 +167,51 @@ export default class Query {
...
+    const hookSteps: messages.TestCase.ITestStep[] = []
+    let pickleStepFound = false
```
aslakhellesoy, line 178: Why is this declared outside the loop?

query/javascript/src/Query.ts
```diff
@@ -157,4 +167,51 @@ export default class Query {
...
+
+    for (const step of testCase.testSteps) {
```
aslakhellesoy, line 180: Can you use `Array#filter` to build `hookSteps` instead?

query/javascript/src/Query.ts
```diff
@@ -157,4 +167,51 @@ export default class Query {
...
+      }
+      if (step.hookId && !pickleStepFound) {
+        hookSteps.push(step)
+      }
+    }
+
+    return hookSteps
+  }
+
+  public getAfterHookSteps(
```
aslakhellesoy, line 192: Can you extract the duplicated code between `getAfterHookSteps` and `getBeforeHookSteps` to a separate function?

query/javascript/src/Query.ts
```diff
@@ -157,4 +167,51 @@ export default class Query {
...
+      if (step.hookId && pickleStepFound) {
+        hookSteps.push(step)
+      }
+    }
+
+    return hookSteps
+  }
+
+  public getTestStepResults(testStepId: string): messages.ITestStepResult[] {
+    const testStepResults = this.testStepResultsbyTestStepId.get(testStepId)
```
aslakhellesoy, line 213: Intermediate variable not needed - just return the result

vincent-psarga:
@aslakhellesoy we were discussing with Bastien and one question came: should we always display the Hook Step if there's an attachment or only when it failed ?
My guess is that we should always display attachments.

PR #976, opened 2020-04-20
Add new Meta message

aslakhellesoy:
## Summary

A new Meta message with meta information

## Details

The first message of a message stream should be a `Meta` message. 

There is also meta information about cucumber implementation, operating system, language runtime etc.

## Motivation and Context

The `protocol_version` field will allow the consumer to determine whether or not the following messages can be processed.

The other meta `Product` fields can be used for diagnostic and statistical purposes, to understand more about the platform on which Cucumber was run.

We need to update `fake-cucumber` to emit this message before merging this.


aslakhellesoy:
Related PRs in Cucumber-Ruby and Cucumber-JVM:

* https://github.com/cucumber/cucumber-ruby/pull/1409
* https://github.com/cucumber/cucumber-jvm/pull/1952

Related issue in Cucumber.js:

* https://github.com/cucumber/cucumber-js/issues/1310

PR #977, opened 2020-04-21
Update messages in json to messages

vincent-psarga:
The `package.json` file was not pointing to the mono-repo version but to 10.0.

Thi fixes it and also do the upgrade to v12


aslakhellesoy:
Good catch! The latest messages version is now 12.1.1

Issue #982, opened 2020-04-29
json-to-messages does not produce the `source` messages


vincent-psarga:
## Summary

When running `json-to-messages`, the `source` messages are not produced. They should be so the output is closer than what we have with any other `cucumber` implementation.

## Expected Behavior

One `source` message should be produced for each `.feature` file used to generate the JSON report.

## Current Behavior

No `source` message at all.

PR #984, opened 2020-04-29
json-to-messages - emit missing messages

vincent-psarga:
Emit the missing messages:
 - [x] `Meta` message to specify the version of the tool/messages
 - [x] `Source` message for each `GherkinDocument` (the feature is not reproduced yet) see #982


vincent-psarga:
Note about the emitted `Meta` message. I've excluded all information about the underlying OS/CPU etc, as the seems not really relevant there (what is more interesting is the environment where tests have been executed which we don't have here).

PR #989, opened 2020-05-02
translate `rule` keyword to Russian

githubeing:
I'm a newbie in Cucumber, not sure if the `rule` keyword is also to be translated to other languages like other keywords are (I've looked at `blame` history of current file and noticed that this keyword was introduced a couple of years ago and for some reason hasn't been translated since then), but if it is - the translation of English noun `rule` to russian is `правило`

## Types of changes

## Checklist:


githubeing:
https://github.com/cucumber/docs.cucumber.io/pull/463

PR #1000, opened 2020-05-13
Exclude some folders from release process

vincent-psarga:
## Summary

When doing release `1.0.2` or `json-to-messages`, we've had some issues with @tooky during the release process.
In this particular lib, we have a few folders (java, ruby, javascript & python) which are there only for producing test data, nothing that needs to be released in the end.

So in this PR, I propose naming those folders `<language>-testdata` so:
 - it's clear when seeing them that their purpose is simply producing test data
 - the Makefile exclude most of the release process for those (except updating libraries)

I did the same for `json-formatter` as the javascript and ruby folders are only there for test purposes. I don't recall other places where we have this kind of folders.

Ah and I also added `rwildcard` in all `default.mk` as this is a pretty useful function, so instead of copy/pasting it every time it's now available in all packages.


aslakhellesoy:
Looks good to me!

vincent-psarga:
great :) I'm gonna update this branch so it ease our next releases.

PR #1001, opened 2020-05-15
build: Add mono and berp to docker

mpkorstanje:
To make https://github.com/cucumber/cucumber/pull/771 possible we need both mono and berp. Currently mono isn't installed and berp is checked in as a binary executable. Both are less then ideal. 

Currently mono isn't supported on the same version of alpine as .NET Core SDK 2.2.2. To avoid messing around too much with APK which doesn't support mono yet I've changed the base image to `ubuntu:20.04`. This makes it easier to install mono.

Additionally we can now also use the node version manager to manage the node and npm versions. They'll be installed without the need for global permissions. This solves another long standing build headace.

The make file will now use `mono  /var/lib/berp/1.1.1/tools/net471/Berp.exe -g gherkin.berp` which is less out of the box then it was before, but the exact installation instructions for berp and mono can be found in the `Dockerfile`.


mpkorstanje:
@vincent-psarga  the `json-to-messages` uses behave but I'm not sure if I see it running. I do see some unit test fail but the build doesn't fail. Not sure what this is all about.

mpkorstanje:
Note to self: Try and use `JAVA_Package -> io.cucumber.cucumberepxressions` in berp file. This may obsolete the use of razor files for each different language implementation.

vincent-psarga:
I think the tests you're talking about are some cucumber tests right ?
There's tests that are meant to faio to produce some examples of JSON with a stack trace, so we expect those failures.

mpkorstanje:
I don' think its those.



```
mkdir -p acceptance/ruby/parameters
../../compatibility-kit/scripts/run-formatter \
                -e .json \
                -o acceptance/ruby/parameters \
                -c "../../compatibility-kit/scripts/cucumber-json-formatter" \
                acceptance/ruby/parameters/parameters.ndjson
✘ acceptance/ruby/parameters/parameters.ndjson
make[2]: [Makefile:88: acceptance/ruby/parameters/parameters.ndjson.json] Error 1 (ignored)
.
.
.
make[2]: Leaving directory '/app/json-to-messages/javascript'
make[1]: Leaving directory '/app/json-to-messages'
```

mpkorstanje:
Running into this:

https://discuss.circleci.com/t/sourcing-bash-file-before-every-run-command/13467

Solution doesn't seem to be working.

mpkorstanje:
Right.`.bashrc` wasn't executable.

mpkorstanje:
Right. Found it.

`.bashrc` doesn't apply when not in an interactive shell. Should move part of nvm scripts to `.profile`

```
# If not running interactively, don't do anything
case $- in
    *i*) ;;
      *) return;;
esac
```

mpkorstanje:
Don't forget to update Berp.

mpkorstanje:
This fails on circle ci with a core dump but passes when building inside the docker container locally.

```
upx dist/cucumber-gherkin-openbsd-386 -o dist_compressed/cucumber-gherkin-openbsd-386
                       Ultimate Packer for eXecutables
                          Copyright (C) 1996 - 2018

upx: dist/cucumber-gherkin-openbsd-386: InternalError: can not apply reloc 'LEXEC000:1' without section 'LEXEC025'
```

mpkorstanje:
Ah. Different executable in the new docker image so the core dump file has a different name.

PR #1002, opened 2020-05-18
Add library to handle creation of the Meta message

vincent-psarga:
## Summary

This library will handle the creation of the `Meta` message which is emitted at the beginning of each stream.

First draft will simply provide basic data (name of the implementation that generated the messages and detail about the execution environment).
We'll add other information later on such a git remote/branch/commit hash etc.

## Types of changes

## Checklist:


vincent-psarga:
Note for the java version: I'm definitely not sure of what I did here, I'm not really familiar with Java library bootstrapping, so it's bits of copy/paste. @mpkorstanje any feedback is more than welcome ;)

create-meta/java/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <version>5.6.1</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.cucumber</groupId>
+            <artifactId>messages</artifactId>
```
mpkorstanje, line 59: This will result in a rather annoying upgrade cycle. For every version of messages well have to release a new version of meta.

I'd suggest creating only the meta object here and copying the fields over in JVM.

aslakhellesoy: What if we moved `createMeta` to the `messages` module? That solves the upgrade cycle problem.

aslakhellesoy: The code looks great @vincent-psarga!

mpkorstanje: It would break the upgrade cycle but would also make the messages lib heavier.

It is actually the same problem with Gherkin. And we don't solve that by pushing Gherkin into messages either.

Seperating the message and code that derives the information would be the best way.

Pushing down, worse then keeping it as is.

aslakhellesoy: > I'd suggest creating only the meta object here and copying the fields over to the message in JVM.

@mpkorstanje are you suggesting keeping the `createMeta` function in a separate module (`create-meta`), but decouple it from the `messages` module?

That works for me.

vincent-psarga: I don't really see the point in decoupling to be honest. Can't we simply declare a minimum supported version of `messages` for the library.

To give an example, we release `messages` version 14.0 and `create-meta` version 1.0 (which relies on `messages` v14.0 minimum).

When we do a new release of `messages` but do not change the `Meta` message, we don't have anything to do in `create-meta` and no release needed.

When we do change the `Meta` message (adding a field or whatever), we'll need anyway to update the code of `create-meta` to handle those changes and populate the added field.

When we do a change in `create-meta` (for example extracting new CI information), we don't have to bump the minimum version of `messages` supported.

I'm pretty sure I'm wrong somehow, but I don't see why :D

messages/dotnet/messages.proto
```diff
@@ -63,14 +63,26 @@ message Meta {
...
+
   // A product has a name and a version
   message Product {
     // The product name
     string name = 1;
     // The product version
     string version = 2;
   }
+
+  message CI {
```
aslakhellesoy, line 78: I'm not sure this is the right name or abstraction. Git-specific information should be in a `Git` message I think. That way we can add support for other source control systems later if needed

I'm not sure what `name` represents, or where it comes from. Same for `projectName`.

How about:

```
message Git {
  string remote = 2;
  string revision = 3;
  string branch = 4;
  string tag = 5;
}
```

The `branch` and `tag` fields should possibly be `repeated` since it's possible to have many.

In practice there will probably only be one, taken from common CI systems' environment variables.

vincent-psarga: That seems a right level of abstraction. I guess notions like the projectName (it would be `cucumber` or `cucumber-ruby` in our case for example) could be extracted from those information.

So something like:

```
message Git {
  string remote = 1;
  string revision = 2;
  string branch = 3;
  string tag = 4;
}

message CI {
  string name = 1;
  string buildId =1;
}
```

I'm not sure about repeating the branch/tags, even if a commit may belong to multiple tags and branch, generally we execute from a given branch or tag.

create-meta/java/pom.xml
mpkorstanje: We'll have to fix the bit of code that removes version ranges when releasing the java modules.

messages/dotnet/messages.proto
mpkorstanje: Maybe we should make it clear that this isn't the git info but CI provided git info.

aslakhellesoy: Yes that's a good point. I think that would be made clear by nesting the `Git` message inside `CI`.

aslakhellesoy: I think I'd prefer a hyperlink instead of a `buildId`. We could name it `url` - it would point to the CI build.

I think we should remove `name` - it can be derived from `remote`

aslakhellesoy:
I'd like to finish this PR - I'm working on some code to populate the `CI` field and nested `Git` field.

It will be based on the environment variables by the [most popular CI tools](https://www.katalon.com/resources-center/blog/ci-cd-tools/). Each CI tool's environment variable name will be in a JSON file that all the implementations can share.

aslakhellesoy:
I've added a TypeScript implementation that populates the `ci` field.

It tries to detect a CI environment based on what environment variables are set. This is the wild west, and CI servers vary a lot.

I have looked at the environment variable documentation for the most common (I think) CI servers out there. This is captured in `ciDict.json` - a file that lists all the supported CI systems, along with expressions to convert environment variables to `ci` properties.

I ended up designing a mini DSL to define the translation from environment variables to these properties. It's simple enough that it can be implemented with a Regexp parser.

I've added approval tests, which will be helpful when this is ported to other languages.

Let's wait to merge this until Java and Ruby is implemented.

I ended up making some minor changes across the project too.

aslakhellesoy:
Any comments about the CI stuff I added before we port it to Java/Ruby @mpkorstanje @vincent-psarga ?

create-meta/javascript/src/createMeta.ts
```diff
@@ -0,0 +1,129 @@
+import os from 'os'
+import { messages, version as protocolVersion } from '@cucumber/messages'
+import defaultCiDict from './ciDict.json'
+
+export type CiDict = { [name: string]: CiSystem }
```
vincent-psarga, line 5: shouldn't it be CIDict, CISystem etc ? (with a capital I ) - although it might make the variable names a bit weird later in the code :/

mpkorstanje: I prefer it as is. My personal favorite: `XMLHTTPRequest`, reads much better when written as `XmlHttpRequest`.

mpkorstanje:
The CI sutff look good to me.

create-meta/java/pom.xml
```diff
@@ -25,39 +25,38 @@
...
+            <version>12.1.2-SNAPSHOT</version>
+        </dependency>
+
+        <dependency>
+            <groupId>com.eclipsesource.minimal-json</groupId>
+            <artifactId>minimal-json</artifactId>
+            <version>0.9.5</version>
```
mpkorstanje, line 38: Can we avoid this dependency or only use it in test scope? When using the `mvn exec:java` plugin you can execute code in test scope. This way we don't need to include the main method in the jar that is distributed.

aslakhellesoy: The library needs to read `ciDict.json` at runtime. I was going to shade it like it's done elsewhere.

Do you have a different suggestion @mpkorstanje?

aslakhellesoy:
@vincent-psarga are you OK to implement the TypeScript for `detectCI`? I've done Ruby and Java, and there are shared approval tests :-P

vincent-psarga:
@aslakhellesoy no problem :) I finish what I do on ruby and start that just after. Do you mind if I do a minor bug fix release of Gherkin before merging/releasing this PR ?

aslakhellesoy:
Sorry I meant TypeScript and Java are done. Ruby is left to do.

vincent-psarga:
The answer is the same, I don't mind handling that :)

create-meta/java/pom.xml
mpkorstanje: Never-mind. I saw it used in the main method and figured that was it. Shading will have to do.

PR #1005, opened 2020-05-20
Tag expressions: Tag exception message

cyocum:
## Summary

This change makes the original tag expression with the error viable to the user via the exception message.

## Details

This change saves the infix string in an object instance variable.  This is then used to create the TagExpressionException object so that the tag expression is visible to the user when an error occurs.

## Motivation and Context

This is required facilitate the fixing of cucumber-jvm bug cucumber/cucumber-jvm#1976.

## How Has This Been Tested?

The tests have been updated to reflect the new exception message.

## Types of changes

## Checklist:

### DIFF
--- a/tag-expressions/java/src/main/java/io/cucumber/tagexpressions/TagExpressionParser.java
+++ b/tag-expressions/java/src/main/java/io/cucumber/tagexpressions/TagExpressionParser.java
@@ -184,7 +184,7 @@ public class TagExpressionParser {
     }
 
     private class Literal implements Expression {
-        private String value;
+        private final String value;
 
         Literal(String value) {
             this.value = value;
@@ -202,8 +202,8 @@ public class TagExpressionParser {
     }
 
     private class Or implements Expression {
-        private Expression left;
-        private Expression right;
+        private final Expression left;
+        private final Expression right;
 
         Or(Expression left, Expression right) {
             this.left = left;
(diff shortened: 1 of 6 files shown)


tag-expressions/java/src/main/java/io/cucumber/tagexpressions/TagExpressionParser.java
```diff
@@ -21,15 +21,19 @@
         put("not", 2);
     }};
     private static final char ESCAPING_CHAR = '\\';
+    private String infix;
```
mpkorstanje, line 24: I'd love to be rid of this bit of mutable state. 

Perhaps we can do this:

```
Expression exp = TagExpressionParser.parse("@food");
```

And then internally
```
public static Expression parse(String infix){
  return new TagExpressionParser(infix).parse();
}

cyocum: That would be nice but I needed the mutable state because the infix is needed in more than one function and I did not initially want to pass that around to multiple functions when something simple would work.  I could pass the `infix` string around but that seems not good as well.

mpkorstanje: That's why I'm suggesting to create a tag parser object per infix that is parsed.

PR #1006, opened 2020-05-27
cucumber-expressions: Support for Optional

gaeljw:
## Summary

This PR aims to support `Optional` types out of the box in Expressions.

## Details

Add support for `Optional<T>` in the (default) `BuiltInParameterTransformer`.

## Motivation and Context

The main goal is to be able to describe a step definition with optional capture groups. For instance: `"^I am logged in(?: as (.+))?$"`.

`Optional` as method parameters are generally considered a bad practice in Java and you would usually expect a `String` parameter with a `null` value in this case.
But on the other hand, other JVM languages (like Scala) consider `null` as forbidden and would benefit from mapping this as an `Optional<String>`.

(Related to https://github.com/cucumber/cucumber-jvm-scala/issues/3)

## How Has This Been Tested?

Tests added in `BuiltInParameterTransformerTest` and `ParameterByTypeTransformerTest` (with Jackson JDK8 Module).

## Types of changes

## Checklist:

## Open questions

I'm just starting to learn how Cucumber "Core" is working, thus I would really appreciate any feedback.

As a start, I have two questions:
- Is this is something we want to document considering that `Optional` is not recommended as parameter in Java?
- Should we add default parameter types in `ParameterTypeRegistry`?

### DIFF
--- a/cucumber-expressions/CHANGELOG.md
+++ b/cucumber-expressions/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 ### Added
+* [Java]  Support for Optional
+  ([#1006](https://github.com/cucumber/cucumber/pull/1006)
+   [gaeljw], [mpkorstanje])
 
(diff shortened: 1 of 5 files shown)


gaeljw:
Great, thanks for the appropriate changes. My Java is a bit rusty :)

PR #1007, opened 2020-05-28
Pretty parse

aslakhellesoy:
## Summary

Move `pretty` and `AstWalker` to the `gherkin` library

## Motivation and Context

We want to use `pretty` without `@cucumber/react`. Initially in the new example mapping tool (for exporting gherkin from an example map), later possibly in other tools, like a linter.

## How Has This Been Tested?

With automated tests

## Types of changes

## Checklist:


vincent-psarga:
Nice :)

That said, why did you remove the description output ? It's pretty useful to have an equivalent output from the source.

Also one thing we should handle too is the tags of the features, scenarios etc.

gherkin/javascript/src/pretty/pretty.ts
```diff
@@ -5,20 +5,13 @@ export default function pretty(
 ): string {
   const feature = gherkinDocument.feature
   let s = feature.keyword + ': ' + feature.name + '\n'
-  if (feature.description) {
```
vincent-psarga, line 8: we should keep the description.

Some search match are done in the description and the tests for filtering should show where the hit was done.

aslakhellesoy:
I see you’ve made aome improvements to `pretty` in the branch that adds tag search (very cool, how does it perform on large docs?)

We’ll have to carry your changes over when we merge it all.

vincent-psarga:
I can make the changes in your branch too, that should not be much of an issue.

By the way there's something weird in this PR, it looks like you're making the PR to be merged into `react-add-search-bar` and not into `master`. I don't know when `react-add-search-bar` will be really ready to be merged into master, I think I'd be better to have this PR directly sent to `master` (or there's something not correctly displaying on GitHub)

vincent-psarga:
Just a thing that popped in my mind: if the `AstWalker` is moved to `@cucumber/gherkin`, that means we'll need to use the library as a dependency instead of a devDependency in react, which I guess wouod make the lib heavier :/
Maybe we coyld extract that to akother lib that only requires `@cucumber/messages` ?

aslakhellesoy:
Sure, we can create a new library, but before we do I'd like to understand the implications of keeping it where it is now (in gherkin).

Can we quantify what problems we might run into by keeping it in gherkin? I'm not worried about size, I'm more worried about frequent releases.

vincent-psarga:
I would worry about size, we already have warnings from `es-slint` about the size of the packaged dist.
It's quite heavy for a simple HTML report, and what's without the messages that end up in the HTML report once everything is there.

Just as an example, when running cucumber ruby suite:
```shell
 ls -lah | grep report                            
-rw-r--r--    1 vincent.pretre  staff   3.8M Jun  5 10:07 report.html
-rw-r--r--    1 vincent.pretre  staff   2.3M Jun  5 15:18 report.messages```

And that's without Gherkin, ElasticLunr etc embedded in the HTML page. The messages are pretty heavy already, let's try not to make this worst.

PR #1012, opened 2020-06-05
Upgrade dependencies

aslakhellesoy:
Upgrading deps


gaeljw:
Not sure this is the right place to discuss this but the PR made me think about it: did you consider using a tool like **RenovateBot** it for keeping depencies up to date?

I use it a lot and find it really great, it supports many languages and build tools (Maven, Go, NPM...).

The idea is to have it as GitHub app and it will open a PR every time a dependency has a new version.
Some rules can also be applied to ignore some dependencies.

aslakhellesoy:
@gaeljw does RenovateBot work with polyglot monorepos?

aslakhellesoy:
They claim it works! Yeah, I'd love to try that out. Didn't have much luck with dependabot and snyk in the past.

gaeljw:
@aslakhellesoy, multiple languages in the same repo it's sure, monorepo not sure to be honest.. 
You can give it a try, it will open a first PR with a clear description of what files it will update afterwards if you accept the initial PR.
If it doesn't suit you, you can then remove it easily from GitHub I guess.

PR #1066, opened 2020-06-17
Cucumber: Fixed #1049 by adding JAVA_HOME to Dockerfile

### LINKED ISSUE #1049: JAVA_HOME is not set in Dockerfile
Had to run

```
 export JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64/
```

to make the java releases work. This should be in the docker file.

cyocum:
## Summary

JAVA_HOME Environment variable added to the release Dockerfile

## Details

The release Dockerfile does not include the JAVA_HOME Environment variable.  This forces someone doing a release to set the variable in their local environment via `export JAVA_HOME=?`.  

## Motivation and Context

This sets the JAVA_HOME for Ubuntu in the Dockerfile so that it is not necessary for the user to set it in their local environment before building and running the container. 

## Types of changes

- [ X ] New feature (non-breaking change which adds functionality).

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,6 +65,9 @@ RUN addgroup --gid "$GID" "$USER" \
         --shell /bin/bash \
         "$USER"
 
+# Configure Java Home
+ENV JAVA_HOME /usr/lib/jvm/java-11-openjdk-amd64
+
 # Configure Ruby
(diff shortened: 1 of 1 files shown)


cyocum:
I rebased as suggested by @luke-hill

Issue #1056, opened 2020-06-15
html-formatter: display a summary of the execution


vincent-psarga:
## Summary

It's a feedback from a `cucumber-ruby` 4 user that can be found here: https://community.smartbear.com/t5/Cucumber-Open/Cucumber-Ruby-4-0-0-has-been-released/m-p/203029/highlight/true#M32

The user misses the summary that was in the old HTMl report and I think it'd be really interesting to put that back in the HTML report.

aslakhellesoy:
I agree. I think we should start with some mockups of how this summary would look:

- What does the summary widget look like?
- Where is it displayed?
- Is it interactive? If so, how?

Pull requests should be made agains `@cucumber/react` - https://github.com/cucumber/cucumber/tree/master/react

PR #1067, opened 2020-06-18
Filter scenarios by status + display Summary

vincent-psarga:
## Summary

Add a filter in the HTML report to only keep scenarios by status. The idea in the long term will be to hide by default scenarios that were not ran.

Screenshot 2020-06-19 at 08 45 16

That's a first draft and looks pretty awful, but provides an idea of what this provides :D

## Types of changes

## Checklist:


vincent-psarga:
Screenshot 2020-06-19 at 14 47 37

I've updated a bit the styling, so it could be a starting point to use the filter as a summary of the execution (see issue #1056) - I'll check if I can do something about displaying more info in a third box (like what's in the `meta` message + execution time for example)

aslakhellesoy:
I like how the header/overview section is emerging. I agree the summary can fit in here too.

vincent-psarga:
I've added the content of the Meta message as a base for the execution summary:

Screenshot 2020-06-22 at 15 20 36

I think we can add more useful information later on, such as the overall execution duration. I'm not 100% sure about the order of the items, maybe the search is a bit more important than the execution summary.

PR #1070, opened 2020-06-19
[gherkin-c] Problem: rule_type.h required to compile library

sappo:
## Summary

Problem: rule_type.h required to compile library

## Details

Solution: Add the generated rule_type.h

Having the generated files part of the source control omits new users
from installing all the dependencies required to generate the file.


aslakhellesoy:
Sounds good to me! @brasmusson @cyocum WDYT?

Issue #952, opened 2020-04-08
cucumber-expressions: optional parameter types do not work in regular expressions


mpkorstanje:
We took some effort to ensure that it is not possible to use optional parameter types. However with some trickery it is possible to use optional parameter types when using regular expressions.

And it turns out this trickery is rather trivial to do: https://stackoverflow.com/questions/61064097/are-cucumber-expressions-interact-somehow-with-regular-expression/61091657#61091657

```gherkin
Scenario: cucumber expression vs regular expression
  And text aaaa text2
  And text text2
```
```java
@Given("^text(?: (.+))? text2$")
public void optionalCaptureGroup(String optionalGroup) {
    System.out.println(optionalGroup);
}

@ParameterType(".+")
public String paramType1(String s) {
    return s;
}
```
I think `ParameterType.transform` should check for this and by pass the transformer (effectively returning null). It seems to assume that there always was something.

```java
T transform(List<String> groupValues) {
        if (transformer instanceof TransformerAdaptor) {
            if (groupValues.size() > 1) {
                if (isAnonymous()) {
                    throw new CucumberExpressionException(String.format("" +
                            "Anonymous ParameterType has multiple capture groups %s. " +
                            "You can only use a single capture group in an anonymous ParameterType.", regexps));
                }
                throw new CucumberExpressionException(String.format("" +
                        "ParameterType {%s} was registered with a Transformer but has multiple capture groups %s. " +
                        "Did you mean to use a CaptureGroupTransformer?", name, regexps));
            }
        }

        try {
            String[] groupValueArray = groupValues.toArray(new String[0]);
            return transformer.transform(groupValueArray);
        } catch (CucumberExpressionException e) {
            throw e;
        } catch (Throwable throwable) {
            throw new CucumberExpressionException(String.format("ParameterType {%s} failed to transform %s to %s", name, groupValues, type), throwable);
        }
    }
```

mpkorstanje:
So this goes a bit deeper. Another way to reproduce the problem:

Yeah. This goes rather deep.

```feature
  Scenario: use a table and generics
    And half
```
```java
ParameterType("half-optional", "(half)?(optional)?", (arg1, arg2) -> Optional.ofNullable(arg1));

Given("^((half)?(optional)?)$", (Optional<String> optional) -> {
    assertEquals(Optional.empty(), optional);
});
```

For some reason I can't quite work out, null values are filtered out in `Group.java`.

```
    public List<String> getValues() {
[...]

PR #1071, opened 2020-06-20
cucumber-expressions/java: Allow parameter types in optional groups

mpkorstanje:
This only happens when a parameter type is used in an optional group. This can
not happen with Cucumber expressions which do not allow parameter types in an
optional group, but it can happen when a parameter type is matched to a regex
in an optional group.

Fixes: #952

## Types of changes

## Checklist:


mpkorstanje:
I'm not happy with this solution yet. What if we have a transformer that transforms to `Optional<String>`? Now it is bypassed and null is inserted.

mpkorstanje:
So this isn't it. This is my favourite type of bug though. 

```
f(g(p(g`(f`(x)))
```
Where f` * g` is the inverse of f * g for some inputs but not all and simply omitting the whole transform makes the problem go away.

PR #1088, opened 2020-06-26
Display CI info in html formatter

vincent-psarga:
As we now have extra information about the CI in the `Meta`message, let's put them to good use so we can have link to the build and the commit.

For now, it displays:
 - a link to the CI build
 - a link to the commit if it comes from Github (more to come after obviously)
 - the box is not there when there's no `CI` on the meta message
  
What it does not:
 - display nice icons for the various tools: the formatter is rather heavy right now, and I did not want to add extra icons with font-awesome and add one or two extra Mb just for that
 - be beautiful :D (I'm not a designer and tend to have poor tastes when it comes to have a usable and beautiful tool, feel free to make it better)

## Screenshot

Screenshot 2020-06-26 at 15 19 24


vincent-psarga:
Screenshot 2020-06-26 at 15 33 57

I've moved the link with other related information about the execution to avoid having an almost empty box.

PR #1093, opened 2020-06-28
Adding scenario location to Pickel

tamil777selvan:
## Summary

Adding scenario location to Pickle object for Javascript.

## Details

Feature to provide location (line and column) of scenario, after Gherkin document is compiled.

## Motivation and Context

For validating the feature file, [gherkin-lint](https://www.npmjs.com/package/gherkin-lint) is being used. 
But unfortunately, with latest Gherkin module, once the gherkin document is compiled, location of the scenario is not given, which make hard to debug the error without line number.

## How Has This Been Tested?

Existing gherkin test results has been modified to validated the functionality by adding location param to the pickles output.

gherkinCompiler

## Types of changes

## Checklist:


aslakhellesoy:
I think it would be better id gherkin-lint used the `GherkinDocument` struct instead of the `Pickle` struct?

Do you know why gherkin-lint uses `Pickle`? It is a data structure designed for execution (cucumber).

Programs that want to obtain the source location for a pickle should use https://github.com/cucumber/cucumber/blob/27726d7939565e6eb3d2265b231f8460dff56682/gherkin/javascript/src/Query.ts#L25

tamil777selvan:
It's not that gherkin-lint uses pickle for validating, but the custom rules which is designed have location as an argument.

If location is passed then it would console the error along with the line number, which is easy to debug if there are more scenario's in a feature file.

The reason why, we use pickle to validate is because if we need to validate rules against tags by parsing gherkinDocument a custom function needs to be created to merge and validate tags present at feature level and scenario level.

But in pickle we have all tags including both feature and scenario level, so it makes easy to validate (PFA).

As suggested in above comments, Have tried to use getLocation method from Query but I'm getting undefined output. 
I'm not sure what is the error I have made here, below is the snippet I used,
`const pickles = new Gherkin.compile(gherkinDocument, '', IdGenerator.incrementing());

pickles.forEach(value => {
   console.log(new Gherkin.Query().getLocation(value.astNodeIds[0]))
});`


tagsExample

aslakhellesoy:
You have to pass all envelope messages to `GherkinQuery#update` before issuing any of the query functions on it. Try that and let us know how it goes.

tamil777selvan:
@aslakhellesoy After passing the all the envelopes to query update, now I can successfully able to `getLocation` using `astNodeIds`. This resolves the issue and also have verified it with gherkin-lint custom rules. Thanks for helping me to fix it :)

aslakhellesoy:
Excellent! I am closing this then.

PR #1094, opened 2020-06-28
Gherkin remove test methods

aslakhellesoy:
Gherkin is currently exporting code from `test`. That folder is not included in the npm package, so the library fails to load.

The exported functions, `parse` and `parseAndCompile` can be replaced by existing alternatives.

This PR refactors the code and removes the test code. It fixes the broken 14.0.0 release.


aslakhellesoy:
This is merged

PR #1099, opened 2020-07-01
Fix computing Timestamps in Ruby

vincent-psarga:
## Summary

This should provide a fix for [cucumber-ruby#1438](https://github.com/cucumber/cucumber-ruby/issues/1438)

Basically, the timestamps generated by `cucumber-messages` are wrong (the way we used the `Time.to_r` does not really correspond to a seconds/nanos division).

This solution is not perfect as there are slight discrepencies  during the conversion, less that 1e6 seconds that said, once again, slight).

I'll try to fix that too before merging.

## Types of changes

## Checklist:


vincent-psarga:
I'd like this to be merged so we can make a release and show correct summaries with `cucumber-ruby` - do you think the 10e-6 precision its enough ? (I personally think so, I'd even think milliseconds would be largely enough as a precision for this kind of tests)

messages/ruby/lib/cucumber/messages/time_conversion.rb
```diff
@@ -23,6 +25,10 @@ def duration_to_seconds(duration)
         nanos_part = duration.nanos.to_f / NANOSECONDS_PER_SECOND
         seconds_part + nanos_part
       end
+
+      def time_to_nanos(time)
+        time.to_i * NANOSECONDS_PER_SECOND + time.nsec
```
mpkorstanje, line 30: This might overflow. And I can't work out who might use this.

mpkorstanje:
Aside from using ISO 8601 rather then seconds since the epoch I don't see a better way to do this. I also think that sub millisecond precision is pointless.

messages/ruby/lib/cucumber/messages/time_conversion.rb
vincent-psarga: indeed, I'll drop that :)

vincent-psarga:
> I also think that sub millisecond precision is pointless.

I agree and don't think anyone will have use of nanoseconds anyway. I think we've inherited that from the default `Timestamp` and `Duration` provided by `protobuf` (before we added our own messages for that), but I'm not sure this will prove useful anytime soon.

Issue #808, opened 2019-11-27
gherkin/javascript: NoSuchLanguageException: if location is undefined, GherkinException thrown


epresi:
## Summary

We would like to parse a feature file with a dialect provided by a user. We do not know if this dialect is valid or not. Our expectations were that `gherkin` will throw the `NoSuchLanguageException`, but we received another exception.

## Expected Behavior

When throwing the `NoSuchLanguageException` exception, `gherkin` should not run into other exceptions.
Outcome should be: 
> Language not supported: en-US

## Current Behavior

Constructor of `TokenMatcher` calls [reset](https://github.com/cucumber/cucumber/blob/69fdf8ec34a8a89dfd4b97b579e9b173a47b4c57/gherkin/javascript/src/TokenMatcher.ts#L18), which then try to [change the dialect](https://github.com/cucumber/cucumber/blob/69fdf8ec34a8a89dfd4b97b579e9b173a47b4c57/gherkin/javascript/src/TokenMatcher.ts#L33).
The [changeDialect ](https://github.com/cucumber/cucumber/blob/69fdf8ec34a8a89dfd4b97b579e9b173a47b4c57/gherkin/javascript/src/TokenMatcher.ts#L21) receives a `location` as parameter, which is optional, and not provided by the `reset`. If `gherkin` is not able to find the dialect, it tries to throw `NoSuchLanguageException`.
In the end, [location is used](https://github.com/cucumber/cucumber/blob/69fdf8ec34a8a89dfd4b97b579e9b173a47b4c57/gherkin/javascript/src/Errors.ts#L23) to display the line and column, however it was not provided to the `changeDialect`. 
Because `location` is `undefined`, we receive `GherkinException` instead of `NoSuchLanguageException`.
~~~
TypeError: Cannot read property 'line' of undefined
    at Function.GherkinException._create (***\node_modules\gherkin\dist\src\Errors.js:34:32)
    at Function.NoSuchLanguageException.create (***\node_modules\gherkin\dist\src\Errors.js:117:21)
    at TokenMatcher.changeDialect (***\node_modules\gherkin\dist\src\TokenMatcher.js:31:52)
    at TokenMatcher.reset (***\node_modules\gherkin\dist\src\TokenMatcher.js:38:18)
    at new TokenMatcher (***\node_modules\gherkin\dist\src\TokenMatcher.js:26:14)
[...]

PR #1102, opened 2020-07-02
gherkin/javascript: Throw NoSuchLanguageException when language is unknown

mpkorstanje:
Fixes: #808

### DIFF
--- a/gherkin/javascript/src/Errors.ts
+++ b/gherkin/javascript/src/Errors.ts
@@ -20,9 +20,10 @@ class GherkinException extends Error {
     }
   }
 
-  protected static _create<T>(location: messages.ILocation, message: string) {
-    const column = location.column || 0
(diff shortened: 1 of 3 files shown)


gherkin/javascript/src/Errors.ts
```diff
@@ -20,9 +20,10 @@ class GherkinException extends Error {
     }
   }
 
-  protected static _create<T>(location: messages.ILocation, message: string) {
-    const column = location.column || 0
-    const m = `(${location.line}:${column}): ${message}`
+  protected static _create<T>(message: string, location?: messages.ILocation) {
```
mpkorstanje, line 23: @vincent-psarga I flipped the argument order of some methods. But I can't really judge of that would be a breaking change or not. I don't think this class is intended to be used by the public?

vincent-psarga:
I don't think it would be a breaking change either.

PR #1111, opened 2020-07-10
Use FilteredResults as the HTML report entry point

vincent-psarga:
## Summary

Replace usage of `GherkinDocumentList` in the `html-formatter` by `FilteredResults`. This allows getting a quick overview/summary of the execution, hide the non-executed scenarios and allows searching document.

## Types of changes

## Checklist:


mpkorstanje:
Seems to work as it should! However when I open the page I'm rather confused. I don't know what to do because there is so much to look at. 

By dividing the header equally between filters, meta and search they all draw equal amounts of attention making it unclear which element I should be looking at. I think the headers primary function should be searching for scenarios so it should be given a more prominent position.

It also seems that the filter element serves a dual purpose. Filtering scenarios and displaying totals. Perhaps these functions could be split. The filters could move to the search element and the totals could go somewhere else.

Also some nit picks that might be good to address after we design a better layout:

* The precision on milliseconds is insane. I've got `671.93701171875 ms`. 
* The unit of time is also duplicated.  `Duration (milliseconds) | 671.93701171875 ms`
* The header bar doesn't use any fonts while the rest of the page does.
* The borders look strange when changing the width of the screen
* The borders in the header are different from the ones in the rest of the report
* For a responsive design it might be good to use a flex-box rather then a table
* The search box doesn't receive the cursor when the page is loaded. 
* The `You can use either plain text for the search or  cucumber tag expressions  to filter the output.` duplicates the purpose of the text in the search bar.

vincent-psarga:
That's good points, I'll try to address them this week.
The only one I don't completelly agree is splitting filtering and totals. I think it would take too much space and might be repetitive (to be honest, the original intent was mainly filrering, but adding totals was pretty straightforward and address two limitations of the first release).
Anyway I'll try some mockups to see how it could be with splitting.

mpkorstanje:
I was thinking of putting the filter in a drop down by the search bar. That way all the search stuff is together.

```
  --------------------------------------------------------------------
  | @Cucumber or (@Gherkin and @Zukini)                   | \ / | O. |
  --------------------------------------------------------------------
```

vincent-psarga:
I guess the filter icon could be found without much problem.

vincent-psarga:
Starting with the nitpicks, it's easier :D

> * The precision on milliseconds is insane. I've got `671.93701171875 ms`.

Ok this has been rounded to 2 digits after the comma.

> * The unit of time is also duplicated.  `Duration (milliseconds) | 671.93701171875 ms`

Done.

> * The header bar doesn't use any fonts while the rest of the page does.

Done.

> * The borders in the header are different from the ones in the rest of the report

Done

> * The search box doesn't receive the cursor when the page is loaded.

I don't know if it should to be honest. I'm not a fan of pages taking cursor input on load, especially on mobile where you suddenly get the keyboard popping up when you did not ask for anything ...

> * The `You can use either plain text for the search or  cucumber tag expressions  to filter the output.` duplicates the purpose of the text in the search bar.

The placeholder disappears as soon as you type something and is kinda painful from an accessibility stand-point, hence the help below.
I see the placeholder as a "quick" help (when it shows correctly etc.) giving a hint of what you can do and the real help below in case someone does not know what `cucumber-query` is for example.

Still TODO:

> * For a responsive design it might be good to use a flex-box rather then a table
> * The borders look strange when changing the width of the screen

vincent-psarga:
I've tried out having the filter moved below the search box and it works effectively fine :)
I've also fixed the responsiveness of the summary, hopefully that's better.

mpkorstanje:
This is good enough to merge.

I do have a few more nit picks:

- The spacing between the filter labels failed, passed, undefined, ect looks a bit tight.

- I think it would make sense to always list all states in the filters and the summary. Right now when all test are passing I'm wondering why I've even got this filter/summary thing.

- The search icon button occasionally detaches from the search box when changing widths:

image

mpkorstanje:
Found  more un-styled text:

image

vincent-psarga:
> * I think it would make sense to always list all states in the filters and the summary. Right now when all test are passing I'm wondering why I've even got this filter/summary thing.

I removed the filter capability when there's only one available option (which indeed does not make any sense).

I think we should keep the number just to get an overall idea of the suite. For showing them all, I started this way but it was not really convincing and taking too much space.

mpkorstanje:
That works too!

vincent-psarga:
Just two last tweaks I had forgotten:
 - styling of the "No results found" message (well, styling, using a font :D )
 - when nothing has been executed and all tests are with the `UNKNOWN` status, we still display the filter so they can be marked as visible again.

luke-hill:
Given the usage of this will be quite prevalent in the 4.x ruby series and other series for other languages, it's likely now due to this modular approach we can constantly refine / refactor if we want. Plus as this is all done in typescript it's a bit more intuitive for others to come in and tweak (I have no knowledge of this though)

Good stuff @vincent-psarga

manish940:
image
How do we exclude un-executed scenarios from HTML report?
Only Passing this in RAKE 
--format html --out #{RakeProfileHelper.output_folder}/#{html_name}.html #{and_tags} #{or_tags}

PR #1120, opened 2020-07-23
messages: Add JavaMethod and JavaStackTraceElement as SourceReference

mpkorstanje:
Cucumber JVM can not reference files by URI. So it should use either java
methods or stack trace elements as a source reference instead.

For example:

```json
{
  "hook": {
    "id": "a1839ec6-f75d-4029-9b75-4b8203d8b2e8",
    "sourceReference": {
      "javaMethod": {
        "className": "io.cucumber.compatibility.attachments.Attachments",
        "methodName": "before",
        "methodParameterTypes": [
          "io.cucumber.java.Scenario"
        ]
      }
    }
  }
}
```

See: https://github.com/cucumber/cucumber-jvm/issues/2058
Fixes: https://github.com/cucumber/cucumber/issues/1119

## Types of changes

## Checklist:

- [X] The change has been ported to .NET.


aslakhellesoy:
Agree - the same problem is likely to exist for other compiled languages, such as C#...

mpkorstanje:
The same problem is likely to exists, but I strongly suspect that different languages will have different semantics when it comes to locating files. By mixing these we make it harder again for consumers to know if they should use the Java semantics or the C# semantics.

vincent-psarga:
yes maybe we could introduce other messages for other languages, it's just that I think the classPath + methodName identifiers should cover most languages (I mean, as long as we don't introduce some semantic in the `classPath` attribute, so it could be `com.example.MyClass` or `Example::MyClass` or any other weird way to name classes, it would work too)

mpkorstanje:
For example given:

```
 "sourceReference": {
    "method": {
      "className": ".....",
      "methodName": "...",
      "methodParameterTypes": [
        "..."
      ]
    }
  }
```

And suppose the name could come from be C#, Java or Ruby. Without looking at the contents of the string. How would a consumer know which parsing method to use? When mixing multiple languages together we lose this information.

PR #1125, opened 2020-07-28
Add human readable duration in react

vincent-psarga:
## Summary

Instead of showing the number of seconds in the html-formatter (or react), we display a human readable version (with days, hours, minutes, seconds and millis when available).

Screenshot 2020-07-28 at 16 11 58

## Types of changes

## Checklist:


marnen:
I can’t speak for anyone else, but I’d rather see the milliseconds as a decimal than as a separate unit (“4.842 seconds”?). Great idea in general, though!

vincent-psarga:
> I can’t speak for anyone else, but I’d rather see the milliseconds as a decimal than as a separate unit (“4.842 seconds”?). Great idea in general, though!

would make sense indeed :)

I've just made a small change based on your feedback:
 - when the execution takes less than a second, we display `ms` with a precision of 2 (for example: "Duration: 245.87 ms")
 - when it's over that, we display `ms` as part of the `seconds field (for example: "Duration: 3m 34.582s")

Which makes me think that we may display only two fields and more than that is not that interesting. For example:
 - 1 day and 6 hours
 - 2 hours and 34 minutes
 - 45 min and 5 seconds
 - 5.789 seconds
 - 267.8 ms

PR #1145, opened 2020-08-12
Remove letter case variations from Gherkin

aslakhellesoy:
Remove letter case variations of Gherkin keywords since their presence causes different annotation classes to be generated during the build of Cucumber-JVM (depending on what OS it was built on).

I discovered this when I was checking semantic versioning of Cucumber-JVM, building on MacOS.

```
CHECK_SEMANTIC_VERSION=true mvn verify -Pcheck-semantic-version -DskipTests=true -DskipITs=true -Darchetype.test.skip=true

...

[ERROR] Failed to execute goal org.revapi:revapi-maven-plugin:0.12.1:check (check) on project cucumber-java: The following API problems caused the build to fail:
[ERROR] java.class.removed: @interface io.cucumber.java.ht.SipozeKe: Class was removed. (breaks semantic versioning)
```

This is related to https://github.com/cucumber/cucumber-jvm/pull/2085

### DIFF
--- a/gherkin/CHANGELOG.md
+++ b/gherkin/CHANGELOG.md
@@ -19,6 +19,13 @@ This document is formatted according to the principles of [Keep A CHANGELOG](htt
 
 ### Fixed
 
+* Remove keyword aliases that only differ by letter case (French, Creole).
+  The reason for this is that Cucumber-JVM generates annotation classes for each
+  step keyword, and some file systems are case insensitive. This led to inconsistencies
+  in the classes that were generated during the build. Removing these keywords fixes
+  this problem. Theoretically this should trigger a new major release, but because the
+  change is so minor and will affect very few users we've made this a patch release.
+
 ## [15.0.0] - 2020-08-07
 
 ### Changed
--- a/gherkin/c/gherkin-languages.json
+++ b/gherkin/c/gherkin-languages.json
@@ -1288,7 +1288,6 @@
       "Scénario"
     ],
     "scenarioOutline": [
-      "Plan du scénario",
       "Plan du Scénario"
     ],
     "then": [
@@ -1616,7 +1615,6 @@
     "given": [
       "* ",
       "Sipoze ",
-      "Sipoze ke ",
       "Sipoze Ke "
     ],
     "name": "Creole",
@@ -1628,11 +1626,8 @@
       "Senaryo"
     ],
     "scenarioOutline": [
-      "Plan senaryo",
       "Plan Senaryo",
-      "Senaryo deskripsyon",
       "Senaryo Deskripsyon",
-      "Dyagram senaryo",
       "Dyagram Senaryo"
     ],
     "then": [
--- a/gherkin/dotnet/Gherkin/gherkin-languages.json
+++ b/gherkin/dotnet/Gherkin/gherkin-languages.json
@@ -1288,7 +1288,6 @@
       "Scénario"
     ],
     "scenarioOutline": [
-      "Plan du scénario",
       "Plan du Scénario"
     ],
     "then": [
@@ -1616,7 +1615,6 @@
     "given": [
       "* ",
       "Sipoze ",
-      "Sipoze ke ",
       "Sipoze Ke "
     ],
     "name": "Creole",
@@ -1628,11 +1626,8 @@
       "Senaryo"
     ],
     "scenarioOutline": [
-      "Plan senaryo",
       "Plan Senaryo",
-      "Senaryo deskripsyon",
       "Senaryo Deskripsyon",
-      "Dyagram senaryo",
       "Dyagram Senaryo"
     ],
     "then": [
--- a/gherkin/gherkin-languages.json
+++ b/gherkin/gherkin-languages.json
@@ -1288,7 +1288,6 @@
       "Scénario"
     ],
     "scenarioOutline": [
-      "Plan du scénario",
       "Plan du Scénario"
     ],
     "then": [
@@ -1616,7 +1615,6 @@
     "given": [
       "* ",
       "Sipoze ",
-      "Sipoze ke ",
       "Sipoze Ke "
     ],
     "name": "Creole",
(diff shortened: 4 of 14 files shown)


mpkorstanje:
You should roll this back. This is not a fix.

This will at the very least break semver for Cucumber JVM and it's not worth a major release. You can use the docker to build and make releases instead.

It will also break parsing. One of the variations won't be recognized any more.

And I also believe the upper/lowercase difference has a semantic meaning for at least some languages.

aslakhellesoy:
> You should roll this back. This is not a fix.

It fixes file name collisions on case-insensitive file systems. Why is it not a fix?

> This will at the very least break semver for Cucumber JVM and it's not worth a major release.

Yes, it does break semver for people using those keywords. I estimate it will only affect a tiny number of users, and the users who are affected can fix it easily by capitalising the keywords in their feature files. That's why I decided to ignore the semver policy in this case.

I added semver ignore rules here: https://github.com/cucumber/cucumber-jvm/blob/a741b27fd3913b3fe724bee3208267a3250065b9/pom.xml#L635-L646

> You can use the docker to build and make releases instead.

The `RELEASING.md` file does not currently suggest or document how to do that. I can document how to do this, and use docker for the next release.

> And I also believe the upper/lowercase difference has a semantic meaning for at least some languages.

I'm pretty sure it does not. I'm fluent in French and there is no difference there. As for Creole, I do not speak it, but I know enough other languages to understand that `Sipoze ke` means `Suppose that`. There is no semantic difference between `Suppose that` and `Suppose That`, and I'd bet you a luxury dinner that the casing bears no semantic difference in Creole either.

I believe it is a mistake to allow mixed casing in keywords. It introduces unnecessary and confusing ambiguity in the language, which is why most formal languages avoid it. SQL is an exception, but I don't believe that's a standard we should strive to mimic. Gherkin should be a case sensitive language.

> At the same time once build properly all variations work because the JVM is case sensitive.

Yep, point taken. However, we cannot expect all contributors to build Cucumber-JVM in docker. I think we should strive to have a build that passes on all operating systems, not just Linux inside docker.

mpkorstanje:
> It fixes file name collisions on case-insensitive file systems. Why is it not a fix?

We use Semver and the Change to communicate towards users. The expectation is that a fix does not change the public API and can generally speaking be automatically applied (for example via Renovate Bot). This breaks this expectation. That this happens only for a small number of users is irrelevant.

>  I estimate it will only affect a tiny number of users, and the users who are affected can fix it easily by capitalising the keywords in their feature files. 


> The RELEASING.md file does not currently suggest or document how to do that. I can document how to do this, and use docker for the next release.

I'm quite proud of that! It doesn't have to be documented because `make release` will [make the release in docker](https://github.com/cucumber/cucumber-jvm/blob/a741b27fd3913b3fe724bee3208267a3250065b9/Makefile#L76).

This is not irrelevant for Semver. 



And in the larger scheme of things, because you've added the Semver Ignore rules, making the build pass on your machine, this MR is also no longer relevant. We can schedule the removal

mpkorstanje:
> It fixes file name collisions on case-insensitive file systems. Why is it not a fix?

The expectation is that a fix/patch does not change the public API and can generally speaking be automatically applied (for example via Renovate Bot). This breaks this expectation and the value of using semver. That this happens only for a small number of users is irrelevant.

> The RELEASING.md file does not currently suggest or document how to do that. I can document how to do this, and use docker for the next release.

I'm quite proud of that! It doesn't have to be documented because `make release` will [make the release in docker](https://github.com/cucumber/cucumber-jvm/blob/a741b27fd3913b3fe724bee3208267a3250065b9/Makefile#L76).

> I believe it is a mistake to allow mixed casing in keywords. It introduces unnecessary and confusing ambiguity in the language, which is why most formal languages avoid it. SQL is an exception, but I don't believe that's a standard we should strive to mimic. Gherkin should be a case sensitive language.

I don't disagree there. However in practice Gherkin has been a case sensitive language for a long time now. And this mistake will have been propagated into various tools. Fixing this should not be taken lightly. We should at the very least follow the usual deprecation strategy and remove the upper case variations on a major release.

However in the larger scheme of things I don't think that fixing in Gherkin is necessary to fix your problem:

1. By adding the semver ignore rules in Cucumber-JVM you'll make the semver check pass on OSX
2. `make release` will build in docker and not omit the duplicate names. Thus not breaking Semver.

This alone resolves your issues in the short term. So there is no need to change Gherkin yet.

The next step would be deprecating the generated upper case versions and removing them in the next major release in Cucumber-JVM. This permanently removes the problem from Cucumber JVM making the Gherkin issue itself irrelevant.

aslakhellesoy:
> It fixes file name collisions on case-insensitive file systems. Why is it not a fix?
>> The expectation is that a fix/patch does not change the public API and can generally speaking be automatically applied (for example via Renovate Bot). This breaks this expectation and the value of using semver. That this happens only for a small number of users is irrelevant.

I was using the word "fix" in the colloquial sense that it fixes a problem. I agree this isn't a "patch/fix" in the semver sense.

> That this happens only for a small number of users is irrelevant.

You're right. This change should have been filed under `# Removed` and not under `# Fixed`. Semver says the major version number [MUST be incremented](https://semver.org/#spec-item-8) for backwards incompatible changes.

What I intended to do was to make a patch release despite the fact that the change is backwards incompatible. In other words, breaking the rule. I believe there are cases when breaking rules can be justified by the outcome. Not particularly for semver, but for all aspects of life. I cross on red when there are no cars around.

Now that I have more information (doing the Cucumber-JVM release in Docker/Linux) I realise breaking the rule in this instance was unjustified, so I'll revert this change and make a new release.

The next version will be 15.0.2 as described [here](https://semver.org/#what-do-i-do-if-i-accidentally-release-a-backwards-incompatible-change-as-a-minor-version).

Thanks for always challenging me @mpkorstanje!

mpkorstanje:
You're welcome! 

I don't think the traffic light metaphor quite works but we can revisit that in the future if the situation ever comes up again. :laughing:

PR #1168, opened 2020-09-02
Remove credentials in meta message

vincent-psarga:
## Summary

Some CI (like GitLab) may use remote location using login/password (eg: `https://mylogin:mypassword@gitlab-ci.com/org/repo.git`).
The meta message will use it and, if used with [reports](https://reports.cucumber.io), will display it.

The idea is to remove the credentials part of the remote URL (in the previous example, https://gitlab-ci.com/org/repo.git`)

## Motivation and Context

Avoid sharing credentials.

## How Has This Been Tested?

Unit + Integration tests.

## Types of changes

## Checklist:

### DIFF
--- a/create-meta/ruby/lib/cucumber/create_meta.rb
+++ b/create-meta/ruby/lib/cucumber/create_meta.rb
@@ -44,7 +44,7 @@ module Cucumber
         url: url,
         name: ci_name,
         git: Cucumber::Messages::Meta::CI::Git.new(
-          remote: evaluate(ci_system['git']['remote'], env),
+          remote: clean_sensitive_information(evaluate(ci_system['git']['remote'], env)),
           revision: evaluate(ci_system['git']['revision'], env),
           branch: evaluate(ci_system['git']['branch'], env),
           tag: evaluate(ci_system['git']['tag'], env),
--- a/create-meta/ruby/lib/cucumber/create_meta.rb
+++ b/create-meta/ruby/lib/cucumber/create_meta.rb
@@ -80,6 +80,13 @@ module Cucumber
       m[1]
     end
 
-    module_function :create_meta, :detect_ci, :create_ci, :group1, :evaluate
+    def clean_sensitive_information(value)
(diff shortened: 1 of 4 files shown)


aslakhellesoy:
> if used with reports, will display it

It won't actually display it, on the web page. The vulnerability is that the ndjson message stream is publicly available, and anyone who guesses a URL can obtain credentials.

create-meta/java/src/main/java/io/cucumber/createmeta/CreateMeta.java
```diff
@@ -57,6 +57,17 @@
                 .build();
     }
 
+    public static String cleanSensitiveInformation(String value) {
```
aslakhellesoy, line 60: The only field that can contain credentials is the git remote, which is a URL. I think it would be safer to parse the value into a URL object, remove the credentials using the URL API and then write it out again.

create-meta/java/src/test/java/io/cucumber/createmeta/CleanSensitiveInformationTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.HashMap;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.matchesPattern;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class CleanSensitiveInformationTest {
```
mpkorstanje, line 13: With JUnit 5 test classes can be package private. This is nice because code completion won't suggest them outside of their package.

create-meta/java/src/main/java/io/cucumber/createmeta/CreateMeta.java
aslakhellesoy: I've done this now

aslakhellesoy:
I think this is good to go now. I've refactored to using url parsing. I've also made the code handle null input to prevent this from blowing up when the environment variables are not defined.

PR #1179, opened 2020-09-09
Enable overriding components used in @cucumber/react

vincent-psarga:
## Summary

In order to ease integration of `@cucucmber/react` into other tools, we need to make sure than it is possible to override/customize the components used for rendering.
This PR proposes a solution allowing to override/change the Attachments` component, if proved useful, it may be applied to other components.

Also, I moved all the `Context` objects into a `/contexts` folder, similar to `/components` as we start having too much of them ;)

## Details

As we render everything based on the Gherkin AST, trying to pass properties to `GherkinDocumentList` will quickly become a mess.
The idea here is to have a new `Context` object, called `MessageToComponentMappingContext` which provides a mapping between a `message` and the component used to render it.

We will not make all components customisable for now, this PR focuses on the Attachments ad if it works fine, it will be generalised based on our need.

## Motivation and Context

Enable customisation of the components in other applications than `cucumber-html-report`

## How Has This Been Tested?

With some mocha tests ensuring we can override the component.

## Types of changes

## Checklist:


aslakhellesoy:
Can you add a StoryBook example?

vincent-psarga:
I've added an extra component to lookup which component to use, but I'm definitely not a fan of the name (`CustomizableAttachment`), if anyone has a better idea, I take it :)

> Can you add a StoryBook example?

Done :)

aslakhellesoy:
I'd like to upgrade StoryBook as part of this PR. We're long overdue upgrading to v6.

I know there is an argument for doing that in a separate PR. However, separate issues for doing upgrades tend to be down-prioritised.

By having a policy that states "Every PR should also make sure all dependencies are the most recent ones", we're adding a bit of boy-scouting to our process....

aslakhellesoy:
Bump

aslakhellesoy:
This is superseded by https://github.com/cucumber/common/pull/1391

PR #1180, opened 2020-09-09
html-formatter: fix when installed locally to node project

davidjgoss:
## Summary

Fixes the `@cucumber/html-formatter` CLI when the package is installed as a project dependency (rather than only working when installed globally).

## Details

Installing to a project like `npm i @cucumber/html-formatter` and then running in an npm script would currently yield an error like:

```
[Error: ENOENT: no such file or directory, open '/Users/davidgoss/Documents/Projects/html-formatter-test/node_modules/@cucumber/html-formatter/dist/src/../../node_modules/@cucumber/react/dist/src/styles/cucumber-react.css'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/davidgoss/Documents/Projects/html-formatter-test/node_modules/@cucumber/html-formatter/dist/src/../../node_modules/@cucumber/react/dist/src/styles/cucumber-react.css'
}
npm ERR! Test failed.  See above for more details.
```

This is due to the relative path to the cucumber react CSS file - this breaks when installed as a dependency, because our own dependencies are usually hoisted up to the `node_modules` of the host project.

This changes to use a reliable function to resolve the path to the `@cucumber/react` package so it works in global or local context.

## Types of changes

## Checklist:

### DIFF
--- a/html-formatter/CHANGELOG.md
+++ b/html-formatter/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
-* [JavaScript] Fix CSS resolution when installed locally (#1180)
+* [JavaScript] Fix CSS resolution when installed locally [#1180](https://github.com/cucumber/cucumber/pull/1180)
 
 ## [9.0.0] - 2020-08-08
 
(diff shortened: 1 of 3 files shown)


davidjgoss:
Actually realised I'll need to do some changes here for https://github.com/cucumber/cucumber-js/issues/1425, so will pick that up on this branch too.

aslakhellesoy:
Awesome - thanks @davidjgoss !

PR #1182, opened 2020-09-13
datatable/java: Support for Optional<T>

rasklaad:
## Summary
Ability to use a tables with Optional types.
## Details
I've added a new TableCellTransformers.

## Motivation and Context

https://github.com/cucumber/cucumber/issues/1091

## Types of changes

## Checklist:


datatable/java/datatable/src/main/java/io/cucumber/datatable/DataTableTypeRegistry.java
```diff
@@ -57,6 +58,44 @@ public DataTableTypeRegistry(Locale locale) {
         TableCellTransformer<Double> doubleTableCellTransformer = applyIfPresent(numberParser::parseDouble);
         defineDataTableType(new DataTableType(Double.class, doubleTableCellTransformer));
         defineDataTableType(new DataTableType(double.class, doubleTableCellTransformer));
+
+        TableCellTransformer<Optional<Object>> optionalTableCellTransformer =
+                (String s) -> s == null || s.isEmpty() ? Optional.empty() : Optional.of(s);
```
mpkorstanje, line 63: The pattern `s == null || s.isEmpty() ? Optional.empty() : Optional.of(do-something-transformer-specific)` is very repetitive. Have a look at the way `applyIfPresent` is used to see if you can remove this repetition this..

datatable/java/datatable/src/main/java/io/cucumber/datatable/DataTableTypeRegistry.java
```diff
@@ -57,6 +58,44 @@ public DataTableTypeRegistry(Locale locale) {
...
+        defineDataTableType(new DataTableType(optionalString, optionalTableCellTransformer));
+
+        Type optionalDouble = new TypeReference<Optional<Double>>() { }.getType();
+        TableCellTransformer<Optional<Double>> optionalDoubleTableCellTransformer =
+                (String s) -> s == null || s.isEmpty() ? Optional.empty() : Optional.of(numberParser.parseDouble(s));
+        defineDataTableType(new DataTableType(optionalDouble, optionalDoubleTableCellTransformer));
+
+        Type optionalFloat = new TypeReference<Optional<Float>>() { }.getType();
+        TableCellTransformer<Optional<Float>> optionalFloatTableCellTransformer =
+                (String s) -> s == null || s.isEmpty() ? Optional.empty() : Optional.of(numberParser.parseFloat(s));
```
mpkorstanje, line 76: The transformer specific function (e.g. `numberParser.parseFloat`) for both the optional and non-optional transformer should be the same. However it is very hard to verify that this is the case. By moving the definition of the optional and non-optional transformer closer together you can resolve this.

Additionally, in combination with my previous comment you can possibly also make it easier to verify that the transformer specific part is indeed the same.

datatable/java/datatable/src/main/java/io/cucumber/datatable/DataTableTypeRegistry.java
```diff
@@ -57,6 +58,44 @@ public DataTableTypeRegistry(Locale locale) {
...
+                (String s) -> s == null || s.isEmpty() ? Optional.empty() : Optional.of(Byte.decode(s));
+        defineDataTableType(new DataTableType(optionalByte, optionalByteTableCellTransformer));
+
+        Type optionalBigDecimal = new TypeReference<Optional<BigDecimal>>() { }.getType();
+        TableCellTransformer<Optional<BigDecimal>> optionalBigDecimalTableCellTransformer =
+                (String s) -> s == null || s.isEmpty() ? Optional.empty() : Optional.of(numberParser.parseBigDecimal(s));
+        defineDataTableType(new DataTableType(optionalBigDecimal, optionalBigDecimalTableCellTransformer));
+
+
+        Type optionalBigInt = new TypeReference<Optional<BigInteger>>() { }.getType();
```
mpkorstanje, line 95: `optionalBigInt` -> `optionalBigInteger`. It's  okay to spell things out.

mpkorstanje (changes requested):
Looking good! Though the code is a bit repetitive. You can probably use lambda functions or method references to extract out the common parts.

datatable/java/datatable/src/main/java/io/cucumber/datatable/DataTableTypeRegistry.java
```diff
@@ -26,69 +26,79 @@
...
     }
 
     private static <R> TableCellTransformer<R> applyIfPresent(Function<String, R> f) {
         return s -> s == null ? null : f.apply(s);
     }
 
-    private static <T> TableCellTransformer<Optional<T>> transformAsOptional(TableCellTransformer<T> tableCellTransformer) {
-        return s -> s == null || s.isEmpty() ? Optional.empty() : Optional.of(tableCellTransformer.transform(s));
+    private static <R> TableCellTransformer<Optional<R>> optionalIfAbsent(TableCellTransformer<R> t) {
+        return s -> s == null || s.isEmpty() ? Optional.empty() : Optional.ofNullable(t.transform(s));
```
mpkorstanje, line 101: While it would be unexpected the function returned by `applyIfPresent` will return null when `f.appy` returns `null`. So `ofNullable` should be used here.

datatable/java/datatable/src/main/java/io/cucumber/datatable/DataTableTypeRegistry.java
```diff
@@ -26,69 +26,79 @@
...
+        }
+        {
+            TableCellTransformer<Short> shortTableCellTransformer = applyIfPresent(Short::decode);
+            defineDataTableType(new DataTableType(Short.class, shortTableCellTransformer));
+            defineDataTableType(new DataTableType(short.class, shortTableCellTransformer));
+            Type optionalShort = new TypeReference<Optional<Short>>() {}.getType();
+            defineDataTableType(new DataTableType(optionalShort, optionalIfAbsent(shortTableCellTransformer)));
+        }
+        {
+            TableCellTransformer<Integer> integerTableCellTransformer = applyIfPresent(Integer::decode);
```
mpkorstanje, line 67: I've moved the optional transformers next to their non-optional counter parts. This allowed me to see that the transformer for integer was missing.

datatable/java/datatable/src/main/java/io/cucumber/datatable/DataTableTypeRegistry.java
```diff
@@ -26,69 +26,79 @@
...
-
-        Type optionalBigInteger = new TypeReference<Optional<BigInteger>>() { }.getType();
-        defineDataTableType(new DataTableType(optionalBigInteger, transformAsOptional(bigIntegerTableCellTransformer)));
+        {
+            TableCellTransformer<Object> objectTableCellTransformer = applyIfPresent(s -> s);
+            defineDataTableType(new DataTableType(Object.class, objectTableCellTransformer, true));
+            defineDataTableType(new DataTableType(Optional.class, optionalIfAbsent(objectTableCellTransformer), true));
+        }
+        {
+            TableCellTransformer<Object> stringTableCellTransformer = applyIfPresent(s -> s);
```
mpkorstanje, line 35: For consistency I've split `Object` and `String`.

datatable/java/datatable/src/main/java/io/cucumber/datatable/DataTableTypeRegistry.java
```diff
@@ -26,69 +26,79 @@
...
+            TableCellTransformer<Double> doubleTableCellTransformer = applyIfPresent(numberParser::parseDouble);
+            defineDataTableType(new DataTableType(Double.class, doubleTableCellTransformer));
+            defineDataTableType(new DataTableType(double.class, doubleTableCellTransformer));
+            Type optionalDouble = new TypeReference<Optional<Double>>() {}.getType();
+            defineDataTableType(new DataTableType(optionalDouble, optionalIfAbsent(doubleTableCellTransformer)));
+        }
     }
 
     private static <R> TableCellTransformer<R> applyIfPresent(Function<String, R> f) {
         return s -> s == null ? null : f.apply(s);
```
mpkorstanje, line 97: I'm wondering why we're not checking for the empty string here.

datatable/java/datatable/src/main/java/io/cucumber/datatable/DataTableTypeRegistry.java
```diff
@@ -26,69 +26,79 @@
...
-
-        Type optionalBigDecimal = new TypeReference<Optional<BigDecimal>>() { }.getType();
-        defineDataTableType(new DataTableType(optionalBigDecimal, transformAsOptional(bigDecimalTableCellTransformer)));
-
-        Type optionalBigInteger = new TypeReference<Optional<BigInteger>>() { }.getType();
-        defineDataTableType(new DataTableType(optionalBigInteger, transformAsOptional(bigIntegerTableCellTransformer)));
+        {
+            TableCellTransformer<Object> objectTableCellTransformer = applyIfPresent(s -> s);
+            defineDataTableType(new DataTableType(Object.class, objectTableCellTransformer, true));
+            defineDataTableType(new DataTableType(Optional.class, optionalIfAbsent(objectTableCellTransformer), true));
```
mpkorstanje, line 32: `true` here marks the `DataTableType` as replaceable. This means that it's possible to register a new transformer that replaces this one. However this means that the optional transformer is not replaced. This seems somewhat inconsistent.

I would expect that when I replace the `objectTableCellTransformer` for `Object`, the `Optional<Objec>t` transformer is used.  And with that in mind would also expect that by registering a transformer for `X` I would have also registered one for `Optional<X>`.

datatable/java/datatable/src/main/java/io/cucumber/datatable/DataTableTypeRegistry.java
rasklaad: I think empty string is completely fine for stringTableCellTransformer, but it is not fine for others.
Now we throw exception for this case:
```
DataTable table = parse("",
                "| header1   | header2   |",
                "| 1              |                 |"
        );
converter.toMaps(table, String.class, byte.class);

>> io.cucumber.datatable.CucumberDataTableException: 'java.util.List<java.util.List<byte>>' could not transform
      | 1 | [empty] |
```
I think we can safely change this to null if the string is empty.

datatable/java/datatable/src/main/java/io/cucumber/datatable/DataTableTypeRegistry.java
rasklaad: So looks like we have to remove Optional transformers from constructor, and just define Optional<X> for any type in io.cucumber.datatable.DataTableTypeRegistry#defineDataTableType
Will that work?

mpkorstanje:
@rasklaad I've been playing around a bit with automatically registering the optional variant of every table cell transformer that is registered. The problem is a bit more complicated because I'm also adding a new `OptionalType` which means that `DataTableTypeRegistryTableConverter` is missing a few cases that would previously be covered by `OtherType`.  However I do think it will be a better solution.

 https://github.com/cucumber/cucumber/tree/datatable-java-optional-type-replacement

mpkorstanje:
Nevermind. Too complicated.

datatable/java/datatable/src/main/java/io/cucumber/datatable/DataTableTypeRegistry.java
mpkorstanje: I don't know yet. This solution has the advantage of being conceptually consistent.

My suggestion isn't.

mpkorstanje:
@rasklaad I've pushed some changes.

Instead of registering transformers for all optionals, we'll look at the type we are trying to convert too. If the type we are trying to convert too is an optional and has an a regular transformer then we'll wrap the regular transformer in an optional transformer and return that. 

Now I had to throw a whole bunch of tests away. They weren't interesting. They all covered the same thing. When a cell transformer is registered for a type it also works for the optional variant of that type. So we don't need to check all types.

What do you think?

mpkorstanje (approved):

mpkorstanje:
I think this will do. Cheers!

PR #1190, opened 2020-09-22
Upgrade to StoryBook 6

aslakhellesoy:
This upgrades to StoryBook 6. Changes:

- Generate a `.ts` file from each `.ndjson` file in the cck
- Load `.ts` files instead of `.ndjson` files in the stories
- Update StoryBook configuration


aslakhellesoy:
I'm merging this. Some storybook stories have been lost. I say we add them back when/if we need them.

PR #1203, opened 2020-10-01
Add a "update dependencies" entry for PR checklist

vincent-psarga:
## Summary

To ease dependencies updates, a new entry in the checklist for pull requests asks for people making PRs to also run the `update-dependencies` make target (see [this comment in another PR](https://github.com/cucumber/cucumber/pull/1179#issuecomment-696037583))

This should help for the releases process, as we will not have to do this manually for each released package.

I'm not 100% sure about the wording, so feel free to make this more user-friendly :)


aslakhellesoy:
This was my suggestion, but after discussing we decided not to do this. Instead we should shorten the checklist!

PR #1209, opened 2020-10-04
Markdown support (alternative to Gherkin)

aslakhellesoy:
## Summary

This PR adds support for Markdown as an alternative to Gherkin

## Details

The Gherkin lexer/tokenizer has been modified to recognise Markdown. This has been done by adding a new dialect named `md`.

The idea is that the rest of the toolchain remains mostly unchanged:
- The parser is the same
- Cucumber is the same (except it needs to modify the glob logic to load both `**/*.feature` and `**/*.md`)
- Formatters are mostly the same

The biggest change will be in the HTML formatter - or more specifically in `@cucumber/react`. It needs a whole new way to render documents:

* Use a Markdown library to render the *source* instead of our own custom React components to render a `GherkinDocument` AST.
* Decorate the rendered Markdown DOM with results, attachments etc from other messages

## Motivation and Context

Adding prose, diagrams and other rich markup to Gherkin documents is cumbersome at best. 

Although the Gherkin grammar doesn't make it explicit, you can put *anything* in the description section of a `Feature`, `Scenario`etc, and some formatters (such as `@cucumber/react` / `html` formatter) will process this as Markdown.

In other words, it's possible to put small snippets of Markdown inside a Gherkin document.

This isn't how people work with Markdown. If you want to use Markdown, it's much more natural if the entire document is Markdown. This give you more flexibility to write a *readable document*.

We use existing Markdown constructs to recognise scenarios:

* `##` is a `Scenario`
* `###` is `Examples`
* `*` (list item) is a step (`Given`, `When`, `Then`)

## Types of changes

## Checklist:


aslakhellesoy:
Here is an example of how this works in Cucumner-JVM - https://github.com/cucumber/cucumber-jvm/pull/2140

image

mpkorstanje:
What happens if I put more information in markdown document then just scenarios? Like for example if I were to have scenarios at a different heading then `h2`? Or if I were to have documentation at a scenario heading. Or if I were to have a folder of `.md` files some with features, some with just text.

```md
Hello world
===========

Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam,
quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
consequat.

## The world is round

Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu
fugiat nulla pariatur.

### Scenario: Something about math
 * step one
 * step two
 * step three

### Scenario: Something about gravity
 * step one
 * step two
 * step three

## The world is wet

Excepteur sint occaecat cupidatat non proident, sunt in
culpa qui officia deserunt mollit anim id est laborum.

### Scenario: Something about chemistry
 * step one
 * step two
 * step three

### Scenario: Something about weather
 * step one
 * step two
 * step three

## Recommended accommodations

Vivamus eget magna eros. Mauris feugiat elit a lectus vulputate eleifend
a sed nulla. 

### If you are adventurous

Nunc sed auctor sem. Quisque vitae ligula commodo quam
vehicula mollis ut at magna. 

### Only in 2020

ed semper feugiat turpis in vulputate. Praesent varius
leo a enim sollicitudin lobortis. Nam a gravida ex. Nulla
interdum orci purus, vitae pellentesque lorem ultricies ac. 

## Tourist attractions

 Pellentesque efficitur turpis lorem, in tempor metus varius vitae. Proin
lectus dolor, luctus eget pulvinar vitae, commodo nec sapien. Donec
mattis quis mi sit amet auctor.

```

With that in mind I would strongly recommend a separate parser for Gherkin in Markdown. It will make evolving and experimenting with the parser easier. It also removes some of the edge cases that are leaking out of the parser abstraction and into other parts right now (e.g. the json file not being usable to generate annotations).

aslakhellesoy:
> What happens if I put more information in markdown document then just scenarios?

These lines would just be ignored by the scanner. Probably marked as `Empty`.

> Like for example if I were to have scenarios at a different heading then h2?

I've added `GHERKIN_MARKDOWN.md` which describes the proposed syntax in more detail. Scenarios would only be recognised with `##` or possibly `###` if the `##` above is interpreted as a rule. 

> Or if I were to have a folder of .md files some with scenarios, some with just documentation.

I'm not sure. If the user doesn't specify what files to include, we could just parse them all. The "documentation" ones might end up having a lot of "undefined" scenarios. If that becomes a hassle we could come up with a more restrictive syntax to reduce the likelihood of this happening.

> With that in mind I would strongly recommend a separate parser for Gherkin in Markdown.

Maybe. I think it's too early to make this decision. It's an investment I would like to defer until we have more data. If we can make this work without writing a new parser I would prefer that.

mpkorstanje:
There were two ideas that underpinned my questions.

1. The proposed markdown syntax is very inflexible. Why should I define my scenarios always at `h2`. I'm writing a document that contains features, not a feature file that happens to look like markdown. There seems to be no point in using markdown otherwise.

2. The implementation is a feature file with different keywords that happen to intersect with markdown but isn't actually markdown. The Gherkin markdown parser should accept all valid markdown documents but only provide pickles if there are indeed scenarios contained within. Currently the parser will reject anything that isn't structured like a feature file.

aslakhellesoy:
> Why should I define my scenarios always at `h2`

I think it's important that users understand what's regarded as a scenario. The parsing rules must be easy to retain and understand for humans. If we make this flexible, as you seem to suggest, I think it will be harder for people to retain and understand the parsing rules.

Gauge is a tool inspired by Cucumber that has supported Markdown from the start. It's not as popular as Cucumber, but it seems to have a healthy user base. I take this as a sign that the Markdown syntax they have settled for works for end users, and that it would be safe for us to adopt a similar (or perhaps identical) syntax. [Gauge defines a scenario as a h2](https://docs.gauge.org/writing-specifications.html#scenario).

In order to support the alternative syntax for h2 (text underlined with `-----`) we'd need a more advanced parser. I'd like to try with a 3rd party commonmark parser when/if we decide to add support for this. But as I said above, I think we can defer this until we have more feedback.

> Currently the parser will reject anything that isn't structured like a feature file.

Yes, I think we need to improve the Markdown parser to be more tolerant of documents that don't use the Gherkin structure.

gherkin/GHERKIN_MARKDOWN.md
```diff
@@ -0,0 +1,38 @@
...
+The rules are as follows:
+
+- `#` - Feature
+- `##` - Scenario, Background or Rule, depending on context
+- `###` - Examples, Scenario or Background, depending on context
+- `####` - Examples
+- `* {Keyword}` - Given, When, Then, And and But
+- `|` - Tables (DataTable and ExamplesTable)
+- `\`\`\`` - DocString
+- `>` - prefix for @tags
```
slavcodev, line 15: The quotes may be not very appropriate for tags, they maybe useful for descriptions of feature or scenarios.

Please consider to allow tags in html comments, especially for features, e.g.

```
<!-- @tag @wip -->
# Feature

<!-- @foo -->
## Scenario
```

aslakhellesoy: Using `>` for descriptions of scenarios and features wouldn't be necessary. You'd just use normal paragraphs for that:

```markdown
## This is a scenatio
This is the description
over a few lines...

* a step
* another step
```

Are there other reasons we want to consider something else than `>` to prefix tags?

If we put tags inside HTML comments, they won't be rendered once the Markdown is converted to HTML, and I think most users would expect them to be rendered.

mpkorstanje:
> If we make this flexible, as you seem to suggest, I think it will be harder for people to retain and understand the parsing rules.
> I take this as a sign that the Markdown syntax they have settled for works for end users, and that it would be safe for us to adopt a similar (or perhaps identical) syntax.

I'm not making it flexible. I'm expecting it to be flexible because is called Markdown. The example is meant to illustrate that. Gauge on the other hand is using the `.spec` extension. This avoids the problems of mixing markdown and specifications and manages the expectations.

aslakhellesoy:
One idea is to define scenarios like this:

```markdown
Bla bla

# Feature: Addition

Bla bla

## Scenario: 2+3

Bla bla

* Given I have entered 2

Bla bla

```

People could use any number of `#` we match based on `Scenario`. This also allows internationalising the markdown parsers.

I think this kind of “overlaying” Gherkin on top of Markdown might work better.

It addresses the concern you had about pure documentation documents.

It also makes it easy for authors and consumers to spot more easily what parts of the documentation is executable.

I also think this would be easy to implement.

gherkin/GHERKIN_MARKDOWN.md
slavcodev: Sorry, I maybe not very clear with my bad English. I did not mean `>` as marker for feature description, I mean `>` is a quote in markdown and I imagine I would want to add some quotes in the document which describes the a feature.

> If we put tags inside HTML comments, they won't be rendered once the Markdown is converted to HTML, and I think most users would expect them to be rendered.

Yeah, this is what I also came to after thinking more about my proposal. 

In general I think the `>` is fine (if parser will look tags which starts with `@` in the quotes).

The only thing that unclear to me, is the parser going to look for tags before `# Feature` or after? Because I think the quotes before level 1 header is not how we usually write the documents.

aslakhellesoy: Yeah, I think it would look better if the tags came after the heading. However, that changes the grammar a bit and might complicate things.

This could work:

```markdown
# @foo @bar
# Feature: Hello
According to research:
> Water boils at 200C

## @zap
## Scenario: Hello
```

It admittedly looks a bit weird in Markdown and the default rendering, but we could style it so that `# @tag` headers are rendered more like small tags above the real header.

aslakhellesoy:
Here is a summary of what's in this PR so far:

A TypeScript implementation with the following features:

## `@cucumber/gherkin`

* [x] A `MarkdownTokenMatcher` which is used when parsing `.md` documents.
  * Modification of the original `TokenMatcher`, but handles Markdown syntax.
* [x] The `GherkinStreams.fromPaths` function will use `TokenMatcher` or `GherkinTokenMatcher` based on each file's extension.
  * This allows users to adopt Markdown gradually, one file at a time
  * Markdown handling is encapsulated, so Cucumber should be able to use this without modification except for changing the glob used to find feature files: `*.{feature,md}`
  * Formatters work as before - even `@cucumber/react` - it will simply render the Gherkin parts of the Markdown document as a Gherkin document (stripping away everything else).

I haven't added support for tags yet as I am still undecided about syntax.
What about simply using an [indented code block](https://github.github.com/gfm/#indented-code-blocks)?

```markdown
    @foo @bar
# Feature: hello
```

## `@cucumber/react`

* [x] A new `<MarkdownDocument>` React component
  * Renders the document with `react-markdown` and overlays results from the message stream.
    * The implementation differs from `<GherkinDocument>` in that it uses the
      source to render the document, and doesn't use the AST at all.

## `@cucumber/fake-cucumber`

* [x] Can execute Markdown documents: `fake-cucumber file.md`

## `@cucumber/compatibility-kit`

* [x] Example with Markdown
* The generated ndjson messages are used in the Storybook example for `<MarkdownDocument>`

Here is a screenshot:

Screenshot 2021-02-09 at 23 54 18


After a bit more polish I wonder if I should merge this to master and release it. It could be used in cucumber-js for a while to test out the waters.

Is this something you see yourself using? What benefits and what challenges do you see?

---

Markdown documents can have "executable" sections that Cucumber runs.
A Scenario is `## Scenario: hello`, a step is a `* Given I have 42 cukes`
etc. 

The idea with this is to make it possible to build a **collaborative living documentation wiki**. 
Documents can be a combination of Markdown or Gherkin. Markdown documents
use the GFM flavour, which is sufficient for fairly rich documents.

Documents can be browsed, searched and navigated easily. Various revisions
can be examined. Collaborative editing is just a click away.

This allows for an interactive and collaborative experience when working
with living documentation.

The freedom of Markdown enables users to write better and more engaging living documentation.
It's even possible to use a WYSIWYG editor like [ProseMirror](https://prosemirror.net/) 
or Quill to do this.

Results would not be instantaneous - but the editor would allow saving directly to
Git, and update itself when results come in after CI has run them.

mpkorstanje:
# Proposal

Use the extension `.feature.md` for all gherkin-markdown files.

## Motivation
 * Cucumber JVM by default scans the entire classpath for `.feature` files by default. This is reasonably safe. This is not true for `.md`. 
 * Existing syntax highlighters\will be able to highlight a reasonable representation `.feature.md` and can be extended to more specific support later on.
 
 ## Alternatives

Keep using `.feature` and use a pragma, to determine the syntax used:

```
 #flavour  markdown
```

This would cover #1, but not #2.

aslakhellesoy:
I think it's essential that Markdown documents can be rendered by existing Markdown tooling (such as GitHub). It's kind of the whole point. We can provide a *better* Markdown renderer (which also displays results), but standard ones also need to work.

This means the files have to end with `.md` - existing Markdown tools won't know how to render a `.feature` file as Markdown. I think that's a show stopper. WDYT @sebrose ?

My intention is that the parser will process arbitrary `.md` documents (even ones without any Gherkin in them) and produce empty ASTs and pickles.

If we manage to do that, what problems do you see just keeping the extension `.md` (as opposed to `.feature.md`) @mpkorstanje ?

davidjgoss:
This looks really good!

I think the `.feature.md` convention would be a nice balance. As well as the better specificity for classpath scanning and default globs, editors could use it as signal to try to provide type assistance for steps and "run" buttons inline, like they do with feature files.

mpkorstanje:
> My intention is that the parser will process arbitrary .md documents (even ones without any Gherkin in them) and produce empty ASTs and pickles.
> If we manage to do that, what problems do you see just keeping the extension .md

To be fair, just because I don't see problems doesn't mean they don't exist  So you'd have to turn the argument around and come up with a solid reasoning why it won't run into problems

However by not being explicit we can run into all sorts assumptions. Most of which we can't know about. So in principle it's better to be explicit.

But some ideas:
 - An arbitrary feature could be a feature that was intended to be valid but isn't actually valid. The parser can't throw a parse error because you intended to ignore regular markdown files.
 
 Now we can't tell the user and they'll be left wondering why their in their mind valid feature isn't being executed.

- I think all files read are published as events (test source read event, don't know the proto equivalent).

These events are scooped up by various plugins. Scooping up all markdown files and publishing them could result in an unintentional leak of information.

gherkin/GHERKIN_MARKDOWN.md
```diff
@@ -0,0 +1,42 @@
+# Gherkin Markdown
+
+Gherkin Markdown lets you write Cucumber scenarios in Markdown. Plain and simple.
+
+Gherkin Markdown is not a Markdown dialect per se, but rather a set of rules for
+parsing scenarios from any Markdown document.
+
+The parsing rules are as follows:
+
+- `#` - Feature
```
tooky, line 10: Does this only support the `#` syntax for specifiying H1? or also the underline syntax?

e.g.

```
An H1
====
```

Would that also be matched as a feature?

aslakhellesoy: Not at this time - implementing support for the underline syntax would be too hard for the current line-oriented parser.

cbliard:
> [...] it's much more natural if the entire document is Markdown. This give you more flexibility to write a readable document.

I like the initial idea of letting people write a markdown document, and add examples where they see fit. I like how it is rendered. 

Meanwhile having a mapping h1 <-> feature and h2 <-> scenario puts a big constraint on the markdown document expected to be produced. If there is a 1-1 mapping between a feature markdown file and its feature file counterpart, then that trumps the initial goal of being flexible.

As I'd expect an example to appear at any point in the markdown document, an alternative approach could be to locate those examples, and then backtrack the last seen header as a scenario name. If there are multiple examples in the same section, then the scenario name can be suffixed with a number to differentiate them.

And for the `Feature:` level, I understand it was needed with the feature file to give it a title and some structure, but would it still necessary for a markdown feature file? And what happens if there are multiple h1? That means the markdown file maps to multiple feature?

Not thought about tags and datatable so far.. I'd love to have feedback.

aslakhellesoy:
> Meanwhile having a mapping h1 <-> feature and h2 <-> scenario puts a big constraint on the markdown document expected to be produced.

@cbliard the number of `#` in a header is insignificant. The following documents will create the same Gherkin ASTs:

```md
# Some heading ignored by the parser
## Feature: a feature
### Scenario: A scenario
Bla bla

* This is
* a bullet list 
* ignored by Gherkin

Bla bla
* Given this is picked up by Gherkin
* Then some more Gherkin
```

And this:

```md
## Some heading ignored by the parser

#### Feature: a feature

## Scenario: A scenario
Bla bla

* This is
* a bullet list 
* ignored by Gherkin

Bla bla
* Given this is picked up by Gherkin
* Then some more Gherkin
```

> And for the Feature: level, I understand it was needed with the feature file to give it a title and some structure, but would it still necessary for a markdown feature file?

Yes, a single header with `# Feature:`, `## Feature:` etc. is required. We can look at making this optional (or allow multiple) in a later version.

cbliard:
> @cbliard the number of `#` in a header is insignificant. The following documents will create the same Gherkin ASTs:

Great! I was fooled by the initial description:

> We use existing Markdown constructs to recognise scenarios:
>
> * `##` is a `Scenario`
> * `###` is `Examples`
> * `*` (list item) is a step (`Given`, `When`, `Then`)

> Yes, a single header with # Feature:, ## Feature: etc. is required. We can look at making this optional (or allow multiple) in a later version.

Fair enough

aslakhellesoy:
Jon Acker suggested looking at https://concordion.org/extensions/java/markdown/

Also suggested for tags:

```
`@foo` `@bar`
# Feature: hello
```

aslakhellesoy:
This is now ready for review!

The `@cucumber/react` code still needs a bit of work before this can be merged - I'm interested in a review of the other parts though!

aslakhellesoy:
In order to finish this we have to wait for https://github.com/remarkjs/react-markdown/pull/588 to be merged and released.

aslakhellesoy (commented):
Reviewed with @aurelien-reeves and @mattwynne

* Call it "Markdown with Gherkin" (MdG)
  * Rename token matcher
* Document why we're not matching dialect (has to be set globally by cucumber, see cucumber-js has a command line option for this).
* Document that JSON formatters won't have descriptions for Markdown documents.
  * Tell people who want it that JSON formatter is in maintenance mode - use message format instead.
  * Maybe add descriptions to AST, it might be easy...

PR #1227, opened 2020-10-23
[react] Fix storybook issue

vincent-psarga:
## Summary

After the upgrade to Storybook 6, it was not possible to render the `<FilteredResults>` component anymore in StoryBook (and potentially other stuff were broken, like the `html-formatter`). This should be fixed by now.

To be honest, I still don't get why I had to change:

```
constructor(private readonly gherkinQuery: GherkinQuery) {}
```

to 

```
private readonly gherkinQuery: GherkinQuery

constructor(gherkinQuery: GherkinQuery) {
   this.gherkinQuery = gherkinQuery
 }
```

as both syntax are supposed to be equivalent. But with the original (and shorter) syntax, GherkinQuery was getting undefined in the `TagSearch` object ...

I also updated the acceptance test we have to be more realistic, so:
 -  it uses `<FilteredResults>` as the rendered component (like `html-formatter`does)
 - it does not "share"  gherkin query with `fake-cucumber`. We now render all the messages in a separate function and feed that to the different queries (once again, like what we do in `html-formatter`)

## Types of changes

## Checklist:


aslakhellesoy:
@vincent-psarga you might be able to find a workaround in https://github.com/storybookjs/storybook/issues/11255

aslakhellesoy:
I think this one is outdated - storybook seems to be working fine now.

Issue #1232, opened 2020-10-30
New html formatter produces very large report files


ben-nc2:
The new html formatter produces very large report files. Please consider reducing the size or providing the old html formatter as an option to switch back to for 6.x.x

We have been using cucumber 4 for a number of years and recently made the changes to first upgrade to 5.7.0 and then on to 6.8.1. Our tests are running fine and although the new html formatter is much easier to read and use for debugging of test failures for our build (which currently runs 423 scenarios that invoke selenium for ui testing).

Our 423 scenarios produce an html report file of 314MB which when we are storing this test artifact uses up a lot of space fast.

aslakhellesoy:
Hi @ben-nc2 - we're aware that the large size can be a problem.

How big were your HTML reports prior to upgrading to v6?

Screenshot sizes should be identical between v5 and v6, so the increase in size might come from elsewhere. I think it might have to do with the large number of duplicated `stepDefinition` messages on Cucumber-JVM, as well as the inclusion of more JavaScript.

Some things we can try to minimize the size:

## More space-efficient message format

Some options:

* Use binary representation and use a more compact protobuf library
* Keep using NDJSON, but don't include `messages.js` + the protobuf library in the bundle
* Use msgpack

## More space efficient JavaScript

* Tree shaking
* Minification

## Fewer messages

* Cucumber-JVM outputs all the step definition messages once for every single scenario. It only needs to be outputted once.

ben-nc2:
@aslakhellesoy running with 5.7.0 html reports were 7MB in size where all tests have passed. The same tests running with 6.8.1 create html reports of 310MB in size. If it would be helpful to share both I can check with my employers if the information is sensitive or not.

aslakhellesoy:
Hi @ben-nc2 having some example reports would be very helpful! If you're allowed to share, you can email them to aslak hellesoy smartbear com (add dots and @)

ben-nc2:
I hope the example reports I emailed help you find ways to reduce the overall size of the report files.

I'm very willing to test any alpha/beta releases against the same scenarios to provide a direct comparison if that is at all helpful.

mpkorstanje:
I can think of a few options:

a. Fix Cucumber JVM to only emit step definition once.
b. Externalize attachments when creating the html report
c. Use a different encoding then Base64 to reduce the overhead and make zipping more efficient
d. Filter out messages that are not used by the html report.

None of these are ideal.

a. Is a significant amount of work.
[...]

PR #1236, opened 2020-11-05
cucumber-html: Reduce size of generated file

mpkorstanje:
By ignoring used messages we can reduce the size of the html file significantly.

Fixes: #1232

## Types of changes

## Checklist:

### DIFF
--- a/html-formatter/java/src/main/java/io/cucumber/htmlformatter/MessagesToHtmlWriter.java
+++ b/html-formatter/java/src/main/java/io/cucumber/htmlformatter/MessagesToHtmlWriter.java
@@ -70,6 +70,15 @@ public class MessagesToHtmlWriter implements AutoCloseable {
             writer.write(",");
         }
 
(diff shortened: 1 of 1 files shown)


aslakhellesoy:
I think a fix for #1232 should be addressed at a more fundamental level: *Reducing the size of the generated messages*.

For example, this PR does not address the size problem for Cucumber Reports.

mpkorstanje:
It doesn't have to be an either/or solution. Reducing the size of the messages and reducing the size of the html report are different but related goals.

Fixing this in Cucumber JVM is a problem because the glue is currently not immutable. So there must be a distinct set for each runner/thread.

Additionally because parameter types influence the the regexs in the glue, and because Cucumber JVM supports scenario scoped glue we are forced to use a parameter type registry per scenario. This in turn means we generate and new set of regexs for each scenario. And thus emit a new set of step definitions.

Fixing this is possible but it requires a redesign of how glue is discovered and shared between threads.

PR #1252, opened 2020-11-19
include offending line in error message

aslakhellesoy:
When the NDJSON stream contains lines that are not valid JSON, we want the error to include the line to make it easier to diagnose the problem.

### DIFF
--- a/messages/java/src/main/java/io/cucumber/messages/NdjsonToMessageIterable.java
+++ b/messages/java/src/main/java/io/cucumber/messages/NdjsonToMessageIterable.java
@@ -41,7 +41,7 @@ public class NdjsonToMessageIterable implements Iterable<Messages.Envelope> {
                     try {
                         JSON_PARSER.merge(line, builder);
                     } catch(InvalidProtocolBufferException e) {
-                        throw new RuntimeException(String.format("Not JSON: %s", line));
+                        throw new RuntimeException(String.format("Not JSON: %s", line), e);
                     }
                     next = builder.build();
                     return true;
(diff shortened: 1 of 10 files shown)


messages/javascript/src/NdjsonToMessageStream.ts
```diff
@@ -43,7 +43,7 @@ export default class NdjsonToMessageStream<T> extends Transform {
         const object = JSON.parse(this.buffer)
         this.push(this.fromObject(object))
       } catch (err) {
-        return callback(err)
+        return callback(new Error(`Not JSONs: ${this.buffer}`))
```
mpkorstanje, line 46: Perhaps also include the original error? That way we know what the parser was complaining about.

messages/java/src/main/java/io/cucumber/messages/NdjsonToMessageIterable.java
```diff
@@ -37,7 +38,11 @@ public boolean hasNext() {
                         return hasNext();
                     }
                     Messages.Envelope.Builder builder = Messages.Envelope.newBuilder();
-                    JSON_PARSER.merge(line, builder);
+                    try {
+                        JSON_PARSER.merge(line, builder);
+                    } catch(InvalidProtocolBufferException e) {
+                        throw new RuntimeException(String.format("Not JSON: %s", line));
```
mpkorstanje, line 44:
```suggestion
                        throw new RuntimeException(String.format("Not JSON: %s", line), e);
```

messages/javascript/src/NdjsonToMessageStream.ts
aslakhellesoy: That's not easily done in JavaScript - exceptions can't have references to other exceptsions like in Java. I don't think it's needed either.

Issue #1275, opened 2020-12-14
Use react-markdown instead of marked+sanitizeHtml


aslakhellesoy:
The current solution is vulnerable to XSS, and it also causes cucumber-html to fail the build since sanitizeHtml requires polyfills for the browser.

More info here:

* https://github.com/remarkjs/react-markdown#readme

PR #1276, opened 2020-12-14
Use react-markdown as it is more XSS resilient

aslakhellesoy:
## Summary

Fix for #1275

The current `marked` + post-sanitization approach is [not safe](https://medium.com/javascript-security/avoiding-xss-via-markdown-in-react-91665479900). Furthermore, it relies on nodejs libs, which requires large polyfills for the browser.

This PR fixes that by using [react-markdown](https://github.com/remarkjs/react-markdown).


react/javascript/test/HighlightTest.tsx
```diff
@@ -74,74 +77,18 @@ describe('HighLight', () => {
...
-        '<strong>Given</strong> a strong step',
-        'strong',
-        true
-      )
-      const highlighted = Array.from(
-        document.querySelectorAll('#content .highlight')
-      )
-        .map((span) => span.innerHTML)
-        .join('')
+    console.log(document.body.innerHTML)
```
davidjgoss, line 90: Assume you didn't mean to leave this console call in?

react/javascript/test/HighlightTest.tsx
```diff
@@ -74,74 +77,18 @@ describe('HighLight', () => {
     ])
   })
 
-  context('when htmlContent is not set', () => {
-    it('escapes HTML characters', () => {
```
davidjgoss, line 78: We now won't have a test that proves the XSS protection, is that okay?

davidjgoss (commented):
lgtm, couple of queries

react/javascript/test/HighlightTest.tsx
aslakhellesoy: Thanks, gone now.

react/javascript/test/HighlightTest.tsx
aslakhellesoy: Good call - I'll add one

aslakhellesoy: I've added a couple more tests for XSS. It strips away `<script>` tags and allows all other HTML tags.

I'm not entirely sure if we should be that permissive though - it sort of defeats the purpose of Markdown to allow any HTML. Should we be more restrictive?

davidjgoss: Good question. Most markdown implementations seem to allow HTML by default, but I think in a Gherkin document I can't see a use case for doing that. The only thing where Markdown semantics fail a bit

davidjgoss: Good question. Most markdown implementations seem to allow HTML by default, but in context of a Gherkin document I can't see a use case for doing that. I would say start out restrictive and if there's demand for broad HTML support we can look at why.

(I write a lot of Markdown, the only HTML tag I can recall writing recently is <small>, other than that I think the syntax is more than rich enough.)

aslakhellesoy: The standalone HTML report has a search function which will highlight searches with `<mark>`, so we could allow `<small>` and `<mark> for now.

But that seems a bit arbitrary. Are there any downsides to allowing anything but `<script>`?

vincent-psarga: In my opinion, the main issue is not the HTML tags themselves but the possibility to have Javascript executed with `onclick` events etc.
Is there a way with `react-markdown` to ensure there's no Javascript added via HTML tag attributes ? If so, we could allow any HTML tag* without having to question too much the safety :)

* well, any tag except `<script>` :D

aslakhellesoy: Good point @vincent-psarga! More tests to add then....

react/javascript/test/HighlightTest.tsx
```diff
@@ -74,74 +77,47 @@ describe('HighLight', () => {
...
+    const html = document.querySelector('#content').innerHTML
+    assert.deepStrictEqual(html, '<div class="highlight"><p>We <em>like</em> other HTML tags:</p><section>hello</section></div>')
+  })
 
-      assert.equal(
-        highlighted,
-        '<strong>Given</strong> a <mark>strong</mark> step'
-      )
-    })
+  it('does not render JavaScript event handlers on tags in markdown', () => {
```
aslakhellesoy, line 114: @vincent-psarga does this address your concern about javascript injection in element event handlers?

I assume it's more thoroughly tested in `react-markdown`...

davidjgoss (approved):

Issue #1279, opened 2020-12-17
html-formatter: report fails to execute in browser with circular dependency issue


davidjgoss:
## Summary

What looks like a circular dependency in the JavaScript bundle for the HTML formatter is causing it to fail to render in browsers, resulting in a blank page.

## Current Behavior

Chrome:

~~~
Uncaught TypeError: Object prototype may only be an Object or null: undefined
    at setPrototypeOf (<anonymous>)
    at n (html-formatter.html?_ijt=5or4r1oa2n5t1va26tsrh6pvtr:307)
    at a (html-formatter.html?_ijt=5or4r1oa2n5t1va26tsrh6pvtr:307)
    at html-formatter.html?_ijt=5or4r1oa2n5t1va26tsrh6pvtr:307
    at Object.2411 (html-formatter.html?_ijt=5or4r1oa2n5t1va26tsrh6pvtr:307)
    at __webpack_require__ (html-formatter.html?_ijt=5or4r1oa2n5t1va26tsrh6pvtr:307)
    at Object.1459 (html-formatter.html?_ijt=5or4r1oa2n5t1va26tsrh6pvtr:307)
    at __webpack_require__ (html-formatter.html?_ijt=5or4r1oa2n5t1va26tsrh6pvtr:307)
    at Object.9299 (html-formatter.html?_ijt=5or4r1oa2n5t1va26tsrh6pvtr:307)
    at __webpack_require__ (html-formatter.html?_ijt=5or4r1oa2n5t1va26tsrh6pvtr:307)
~~~

Firefox:

~~~
Uncaught TypeError: Object.setPrototypeOf: expected an object or null, got undefined
    n file:///Users/Aslak.Hellesoy/git/cucumber/cucumber-js/html-formatter.html:307
    a file:///Users/Aslak.Hellesoy/git/cucumber/cucumber-js/html-formatter.html:307
    o file:///Users/Aslak.Hellesoy/git/cucumber/cucumber-js/html-formatter.html:307
    2411 file:///Users/Aslak.Hellesoy/git/cucumber/cucumber-js/html-formatter.html:307
    __webpack_require__ file:///Users/Aslak.Hellesoy/git/cucumber/cucumber-js/html-formatter.html:307
    1459 file:///Users/Aslak.Hellesoy/git/cucumber/cucumber-js/html-formatter.html:307
    __webpack_require__ file:///Users/Aslak.Hellesoy/git/cucumber/cucumber-js/html-formatter.html:307
[...]

PR #1280, opened 2020-12-18
Verify HTML reports with puppeteer

aslakhellesoy:
This automates the verification of generated HTML reports by rendering them with puppeteer. If the rendering fails, the build fails.

In order to make it pass I also had to downgrade to webpack 4 - not sure what's up with webpack 5.


html-formatter/javascript/check.js
```diff
@@ -0,0 +1,22 @@
+const fs = require('fs')
+const path = process.argv[2]
+const html = fs.readFileSync(path, 'utf-8')
+const puppeteer = require('puppeteer');
+
+(async () => {
+  const browser = await puppeteer.launch();
+  const page = await browser.newPage();
+  await page.setContent(html)
+  await page.pdf({path: 'hn.pdf', format: 'A4'});
```
davidjgoss, line 10: What purpose does the PDF generation serve?

html-formatter/javascript/check.js
```diff
@@ -0,0 +1,22 @@
...
+  const dynamicHTML = await page.evaluate(() => {
+    const content = document.querySelector('.cucumber-react')
```
davidjgoss, line 12: Not sure if there could be errors in lower components (e.g. markdown) but the root one still render enough for this to pass?

To solve a similar problem before I've queried the console for errors - I could always do a follow up if you think it's a good idea.

davidjgoss (commented):
Nice approach.

> In order to make it pass I also had to downgrade to webpack 4 - not sure what's up with webpack 5.

Yeah more detail on that in #1279, seems like Webpack's module resolution must have changed and not be able to handle these circular imports any more. I can't see a reference to it in [their announcement](https://webpack.js.org/blog/2020-10-10-webpack-5-release/) unless I'm not recognising the language.

(I've got a change in flight to address the circular imports in our code and add an eslint rule, so I'll leave that issue open.)

html-formatter/javascript/check.js
aslakhellesoy: re: pdf. Copy paste error - removed

html-formatter/javascript/check.js
aslakhellesoy: Yes, inspecting console errors could be useful too. This check does not guarantee absence of errors, but it does detect presence of errors. I think that's a good start.

But we have a problem with CircleCI launching puppeteer.....

davidjgoss: I've seen this before. They started requiring some libraries on the host system from 3.0.0 inc. `libgbm`.

davidjgoss (approved):

PR #1281, opened 2020-12-21
gherkin-elixir: Add missing library runtime dependency

alvivi:
## Summary

This PR adds the `resource` folder to the set of files distributed in Hex. Also, loads the contents of `gherkin_languages.json` in compile time.

## Motivation and Context

If you try to use the current published library in Hex, you get the following error:

```
** (File.Error) could not read file ".../deps/cucumber_gherkin/resources/gherkin_languages.json": no such file or directory
    (elixir 1.11.1) lib/file.ex:354: File.read!/1
    (cucumber_gherkin 16.0.0) lib/gherkin/lexicon.ex:11: CucumberGherkin.Lexicon.load!/0
    (cucumber_gherkin 16.0.0) lib/gherkin/lexicon.ex:12: CucumberGherkin.Lexicon.load_lang/1
    ...
```
This is because the `resources` folder is not distributed in Hex. Fixing this issue is just half of the issue. Later, when distributing an app built using `cucumber_gherking`, for example, using `mix release`, this folder must be also reachable. There are two ways to fix this, add this file to the Erlang `priv` folder, or just read it at compile time and avoid any external file dependency. This PR implements the later.

## How Has This Been Tested?

I uploaded a patched version of the library to Hex. It's available from [here](https://hex.pm/packages/cucumber_gherkin_hotfix). 

## Types of changes

## Checklist:

### DIFF
--- a/gherkin/elixir/lib/gherkin/lexicon.ex
+++ b/gherkin/elixir/lib/gherkin/lexicon.ex
@@ -1,6 +1,6 @@
 defmodule CucumberGherkin.Lexicon do
   @moduledoc false
-  @default_lexicon_path [File.cwd!(), "resources", "gherkin_languages.json"] |> Path.join()
+  @default_lexicon_path Path.join(__DIR__, "../../resources/gherkin_languages.json")
   @feature_keywords ["feature"]
   @scens_both ["scenario", "scenarioOutline"]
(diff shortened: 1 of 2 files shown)


aslakhellesoy:
LGTM - any comments @WannesFransen1994

PR #1285, opened 2020-12-28
gherkin-perl: Use Carton (Perl's Bundler) to store project-local dependencies

ehuelsmann:
## Summary

`cpanm` defaults to writing its packages to global locations. To work around that, the environment needs to be set correctly, which is exactly what Carton is helping us do.

## Motivation and Context

Together with the issued change for the Docker development image (installing Carton), this change sets up the foundation needed to build the next PR I'm about to issue: an update for the Perl implementation of Gherkin.

## How Has This Been Tested?

This PR has been tested on the cucumber-build Docker image and succeeds testing when run on top of the updated Perl implementation.

### DIFF
--- a/gherkin/perl/.gitignore
+++ b/gherkin/perl/.gitignore
@@ -1,6 +1,8 @@
 acceptance/
+local/
 .build/
 .built
 .cpanfile_dependencies
+cpanfile.snapshot
 .compared
 Gherkin-*
--- a/gherkin/perl/Makefile
+++ b/gherkin/perl/Makefile
@@ -19,7 +19,7 @@ default: .compared
 	touch $@
 
 .cpanfile_dependencies:
-	cpanm --installdeps --notest .
+	carton install
 	touch $@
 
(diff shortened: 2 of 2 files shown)


ehuelsmann:
The change to the development image, installing Carton, has landed and the changed image is now available on Docker Hub. This change should now land without problems.

tooky:
Hey @ehuelsmann, I don't have a great idea of the Perl ecosystem. From what I can pickup from google it seems that essentially it is bundler or npm etc. for Perl. I have a couple of questions:

1. Is carton an unusual thing to see in perl projects?
2. Does this change anything for _users_ for gherkin-perl? Or is it simply for managing the dependencies within the project?

If the answer to both of those is no, I don't see any reason not to merge.

ehuelsmann:
Hi @tooky,

The answer to both is indeed "no", although in the Perl world, there's more than one way to do it (Perl's philosophy) and the Perl crowd have - much more than other ecosystems - numerous ways to reinvent the wheel (making bundler and npm ubiquitous, while Carton is just one of the solutions).

To confirm your question: this is only to manage dependencies in the development environment and - more importantly even - to set the correct library paths upon (test) command execution from it.

In short: I think this can be merged indeed. Thanks for your review.

Regards,

Erik.

tooky:
👍

PR #1292, opened 2021-01-02
Remove circular imports in typescript code, add linting

davidjgoss:
## Details

Remove circular imports in messages TypeScript code and add `eslint-plugin-import` to lint against future regressions.

## Motivation and Context

See #1279 which this helps with but doesn't close, because we have a similar issue in the protobufjs library.

These issues block us upgrading to webpack 5 in the HTML formatter. Node's module resolution handles them okay and apparently so does webpack 4, but the runtime module resolution seems to have changed somehow with 5.

## Types of changes

## Checklist:


davidjgoss:
@aslakhellesoy thanks for moving this forward, I hadn't caught on that it would fail unless all eslint configs were the same (this should get easier if we can make npm workspaces happen). I see there are some failures on this rule in gherkin and cucumber-expressions, shall I pick this back up and work on those? (It's not super urgent to fix this, just prevents upgrading to webpack 5.)

aslakhellesoy:
Hi @davidjgoss yes, that would be great if you could continue the untangling of the circular deps in gherkin and cucumber-expressions.

Thanks!!

aslakhellesoy:
Allright, that's cucumber-expressions done. I'll have a quick go at gherkin too....

aslakhellesoy:
Gherkin circular dependency:

Screenshot 2021-02-03 at 21 57 35

aslakhellesoy:
I think Gherkin circulars are gone now - still some small things to fix in cucumber-expressions it seems...

aslakhellesoy:
You opened a can of worms here @davidjgoss :-)

But I'm glad you did. Keeping code nicely decoupled is something I care deeply about, and this has been an eye opener for me. It's the first time (I think) that I have worked with a system that disallows circular dependencies at the compilation level.

I've not seen it in Java or C# before, or any other language (although I'm sure it exists).

In order to untangle these circular dependencies I was forced to do what I have considered good practice for 15 years - [interface-implementation separation](https://wiki.c2.com/?SeparateInterfacesFromImplementation). But I don't always practice what I preach, so having a tool that enforces this is great!

I'll be using this on all my TypeScript code from now on I think.

@mpkorstanje is there anything similar in Java that you know of?

davidjgoss:
@aslakhellesoy awesome, thanks for getting it over the line (i.e. the last, er, 99%)

PR #1293, opened 2021-01-04
[Elixir] Released package didn't include the gherkin-languages.json file

WannesFransen1994:
Missing file, moved to priv folder.


WannesFransen1994:
@aslakhellesoy , when you have time could you review this and publish a new version please? Thank you!

alvivi:
Should I close https://github.com/cucumber/cucumber/pull/1281 then?

WannesFransen1994:
Oh wow sorry I completely missed your PR there! I'd personally suggest the default folder structure its priv folder. If everything is fine for @aslakhellesoy , this could be merged, published and it should be fixed then.

alvivi:
Btw, @WannesFransen1994, Have you tried this patch when building/running a release? For our case, gherkin is being parsed in prod, no just dev/test environments. I'm not sure about the implications of using directly `"priv"` instead of [`:code.priv_lib/1`](http://erlang.org/doc/man/code.html#priv_dir-1).

WannesFransen1994:
Thank you for pointing that out @alvivi . I've played around with it on [https://github.com/WannesFransen1994/ex_gherkin](https://github.com/WannesFransen1994/ex_gherkin) , it worked with just the `"priv"` (in prod, dev and test env) but as you stated, using `:code.priv_dir/1` is definitely a better practice!

I've just looked at your PR a bit more, do you think that the `@external_resource` usage with decoding at compile time will result in a lot of performance increases? I personally didn't pay a lot attention to performance, so I'm really thankful that you mentioned it!

alvivi:
> Thank you for pointing that out @alvivi . I've played around with it on https://github.com/WannesFransen1994/ex_gherkin , it worked with just the `"priv"` (in prod, dev and test env) but as you stated, using `:code.priv_dir/1` is definitely a better practice!

Yeah, for `prod`, `test`,etc. should be no problems. But I was talking about running it inside a [release](https://hexdocs.pm/mix/Mix.Tasks.Release.html). Sorry for the misunderstanding. As far I know, the `priv` directory is copied in the release, at least of the current project. But not sure about dependencies :crossed_fingers:   

> I've just looked at your PR a bit more, do you think that the `@external_resource` usage with decoding at compile time will result in a lot of performance increases? I personally didn't pay a lot attention to performance, so I'm really thankful that you mentioned it!

Is going to be faster and increase the memory footprint a little bit for sure (well, that memory is going to be used anyway). Not sure for this use case is going to make a difference or not (Is just bootstrap time). But no, the main reason for me was to ease distribution. I think this is a common approach for immutable data, for example, https://github.com/elixir-plug/mime (which includes a `priv` folder, but it is been read at compile time).

WannesFransen1994:
Just confirmed that running it in a release isn't an issue. In the release map, the dependencies their `priv` folder is copied as well to that dependency its subfolder.

Did stumble upon an error though, regarding the `elixir_uuid` library. When building a release, `:crypto` isn't included and will raise an error that the function doesn't exist. Simply adding this to its `mix.exs` `extra_applications` section resolved it for me. I'll make an issue for that.

I see your point. As you stated, the memory is going to be used anyway so I'll adjust it as well. Thanks for taking the time to explain this!

PR #1299, opened 2021-01-08
react: Prevent SearchBar from triggering HTML form action

### LINKED ISSUE #1298: react: Search button does not work and causes page to reload
## Summary
In the SearchBar, clicking the search button does not update the search query, and instead causes the page to reload. As a workaround, hitting `Enter` while focus is on the search text box correctly updates the query.

This was tested in the page generated using the `html-formatter` subproject. I haven't tested the component in a wider context.

## Expected Behavior

When clicking the button, the search query should update. If using inside the `FilteredResults` element, the feature list should be filtered and highlighted accordingly.

## Current Behavior

The page reloads.

## Possible Solution

I believe the problem stems from the fact that the text input and button are defined inside an HTML form, and the button is bound to the `submit` action of the form. Thus, in addition to the "onClick" action, it triggers an HTML form submission causing the page to reload, which clears the "query" state from the parent react element.

To fix, either the elements could be unbound from the default form submission (i.e. `Event.preventDefault()`), or the search bar could make explicit use of HTML form submission as means of communicating the search query to the wider app (see #1302 for a suggestion regarding this second option).

1. Run `html-formatter.js` against a messages file containing at least one document.
2. Open the resulting html file.
3. Enter a search term in the search box
4. Click the search button

## Context & Motivation

I came across this bug by accident, while trying to modify the page generated by `html-formatter` to accept queries via the document URL search parameter. I attempted to add a script element to the page that filled the search box and simulated the button click. An ideal solution for me would also provide some other hook into the search form without having to simulate user input. However, I accept that such a request would significantly widen the scope of this issue.

Regardless, to me this seems like a straight-up UI bug.

## Your Environment

* Version used: 11.0.2 (html-formatter 11.0.4)
* Operating System and version: Linux (Manjaro; Gnome); Firefox 84.0.1 & Chromium 87.0.4280.88

hWorblehat:
## Summary

This commit stops the `SearchBar` from triggering an HTML form action
when the search button is clicked. In addition, it correctly changes the
`SearchQueryContext` such that other components (e.g. `FilteredResults`
and `Highlight`) can react to the change, rather than using a
state-setter passed down via props.

## Motivation and Context

Fixes #1298

## How Has This Been Tested?

* All existing tests against the sub-project have been run.
* Manually tested the changes using the `html-formatter`.
* Added a few component unit tests in `test/SearchBarTest.tsx`

## Types of changes

## Checklist:


hWorblehat:
This is ready to be merged, pending review.

One note for review: with this change, highlighting of search terms is updated when the search is performed (either by pressing 'enter' or clicking the button) - it does not update as the user types. This is not a change in behaviour, but the way the code was written before made it look like it was possibly the intention to update highlighting as the user typed (if so, this didn't work). This PR makes the binding between the string used for search and for highlighting more explicit.

react/javascript/src/components/app/SearchBar.tsx
```diff
@@ -53,20 +51,17 @@ const SearchBar: React.FunctionComponent<IProps> = ({
...
         />
-        <button
-          type="submit"
-          onClick={() => queryUpdated(searchQueryContext.query)}
-          value="search"
-        >
+        <button type="submit" value="search">
           <FontAwesomeIcon icon={faSearch} />
         </button>
+        <input type="submit" style={{ display: 'none' }} />
```
davidjgoss, line 64: Why is the hidden input needed? AFAIK the browser will handle the enter key basically as a click on the submit button. Could add a test to prove this either way.

davidjgoss (commented):
Looking good, one point inline.

react/javascript/src/components/app/SearchBar.tsx
hWorblehat: Sorry for the slow response...

You're right: this is unnecessary - my bad. I've removed it. However, I couldn't find a reasonable way of testing this as it looks like a simulated key(down|up|press) in JS won't trigger the DOM form submit action (this seems to be corroborated in the discussion on this SO question [here](https://stackoverflow.com/questions/29268613/simulate-enter-keypress-in-javascript-to-trigger-a-form-submit#comment46738693_29268613)). I've tested it manually and it works fine.

davidjgoss: Yep I know what you mean. I think the testing you've added covers it well enough.

davidjgoss (approved):
Thanks for the contribution

aslakhellesoy:
Hi @hWorblehat,

Aslak Hellesøy
Creator of Cucumber

hWorblehat:
Hi @aslakhellesoy - thanks for the kind words. I'll be launching a new software project and team over the next year, and plan to make some heavy use of Cucumber, so I'm sure I will indeed have some further contributions to make. I already had one other PR for a feature enhancement: #1302.

Anyway, thanks for bringing Cucumber to the world - keep up the good work!

PR #1302, opened 2021-01-09
Allow searching in HTML report via URL search param

hWorblehat:
## Summary

The `react` library and `html-formatter` template have been updated to allow searching of the results via a URL search param.

E.g.:
~~~
http://my.test.results/cucumber-report.html?search=%40slow
~~~
will cause the page to open with `@slow` in the search bar, and the results filtered accordingly.

Conversely, there's a link at the bottom of the search bar that stays up-to-date with a URL of the current page with the current search and status filters.

Here's a short demo:

https://user-images.githubusercontent.com/17484917/106456071-f394ad80-6484-11eb-9201-c4001ae2bbd0.mp4

## Details

In the `react` library, the `SearchQueryContext` has been expanded to include both the search text and the status filters. On the back of #1299, the `SearchBar` interacts with the rest of the app entirely via this context. The `SearchBar` also has an additional optional property: a "render URL" function that is expected take a `SearchQuery` context and convert it to a URL string. If this property exists, the `SearchBar` will use it to render a link at the bottom.

The `html-formatter` app has been changed to pass such a render function in, along with initial query search text and status filter props derived from the current URL. To do this, it makes use of a new function exported from `SearchQueryContext` that defines a default URL parser and renderer.

All additional functionality in the `react` library is entirely optional - it must be activated by passing search text, filters, and render function in as props. Any existing app using these components should continue to behave as they did before without modifications.

## Motivation and Context

This makes it possible to create hyperlinks to a web page of cucumber
results with a particular search applied. This could be used in DevOps
workflows, for example by attaching a link to an issue ticket that
points to a set of features tagged with the issue number.

This is what I am trying to achieve.

## How Has This Been Tested?

Added mocha tests for the expanded functionality in `SearchQueryContext`, `SearchBar`, and `QueriesWrapper`.

Also tested manually with `html-formatter`.

## Types of changes

## Checklist:


hWorblehat:
Just to add a bit of history to this, I tried a number of alternatives to displaying the link at the bottom of the search bar:

- Actually reloading the page with a different URL query string each time a search is performed
- Pushing a history element to the window history state each time

I rejected both of these in the end because it means that a lot of searching quickly clutters up the browser history. Also, the first option (reloading), causes the page to forget which accordion items are expanded & collapsed.

hWorblehat:
Hi @davidjgoss, @aslakhellesoy,

Is there anything I can do to help move this PR along, or find an appropriate reviewer? I appreciate I'm a single user of many effectively asking for a feature to meet my needs, and I'm sure everyone is busy. Still I thought I'd ask in case I'm just not approaching an enhancement proposal in the right way, and need to engage via a different medium.

Cheers,

davidjgoss:
@hWorblehat sorry for late reply. Aiming to do a proper review this week.

mattwynne:
What's left to do to get this merged? @aslakhellesoy @davidjgoss do you still want to review it?

Looks like it will need a rebase now.

davidjgoss:
I'll get this up to date and then do a review - sorry it's gone unreviewed so long.

davidjgoss (commented):
This is looking good @hWorblehat, thanks for doing it and sorry it's been dormant for a while. I've made a couple of edits to bring it up to date with the changes on `main` since you raised.

Couple of observations:

- It's very customisable, down to what the query parameter names are and the function to redraw the URL. We _could_ consider dialling that down to make for a smaller albeit more opinionated change initially
- It would be neat if the URL bar of the browser also updated in real time - as a user my muscle memory wants to copy and paste from there. We could maybe look at doing that with [`history.replaceState`](https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState) but no rush to do that in this PR - would need to take care we don't cause issues with client-side routing in React apps etc.

Interested in @mattwynne and @aslakhellesoy's takes as well.

mattwynne:
> It would be neat if the URL bar of the browser also updated in real time - as a user my muscle memory wants to copy and paste from there. We could maybe look at doing that with history.replaceState but no rush to do that in this PR - would need to take care we don't cause issues with client-side routing in React apps etc.

👍 to this. I really like it if you can copy/paste URLs for search results.

mattwynne:
> It's very customisable, down to what the query parameter names are and the function to redraw the URL. We could consider dialling that down to make for a smaller albeit more opinionated change initially

I've not read the code so I don't know how much complexity this adds, but I'm curious if there was a use-case in mind, like embedding this stuff in another app. In general making our stuff extensible / embed-able seems like a good ambition.

hWorblehat:
Hi @davidjgoss, @mattwynne,

Thanks for taking a look, and no worries about it being dormant - I've been distracted with other things anyway.

---

> It would be neat if the URL bar of the browser also updated in real time - as a user my muscle memory wants to copy and paste from there. We could maybe look at doing that with `history.replaceState` but no rush to do that in this PR - would need to take care we don't cause issues with client-side routing in React apps etc.

I actually started out with this approach, but using `history.pushState`. After playing with it I felt that the browser history got polluted too quickly, so switched to the current approach.

I'm not that familiar with browser APIs, and it looks like I missed `replaceState` as an option. I can add a change to do this next week - it should remove quite a bit of code.

---

> It's very customisable, down to what the query parameter names are and the function to redraw the URL. We could consider dialling that down to make for a smaller albeit more opinionated change initially

I didn't have a particular use case in mind, other than trying to keep cucumber-react as a component library that could easily be reused and embedded in other apps. Given that we're talking about a feature that messes with a property of the entire web page, it felt like being able to customize things like the names of the query parameters would be useful. If we decide to go with the `replaceState` option as above, that would remove a lot of the complexity in the code, and the need for customization, as we wouldn't need to separately render a URL outside of the window location.

davidjgoss:
@hWorblehat fantastic, that sounds like a plan! Let me know if you need anything.

mattwynne:
Regarding `replaceState` / `pushState` I haven't played with this but I assume the pollution of the browser history is because we're searching as the user types, so each keypress results in a new URL in the browser?

davidjgoss:
We don't update on every keystroke, but I think what we're looking to avoid with `replaceState` is the experience where you tweak a few filters or searches and then hit the back button expecting to land back on GitHub/Jenkins whatever but instead are just going back over each iteration of your filters.

(Correct me if I'm wrong @hWorblehat!)

hWorblehat:
:+1:  Exactly

hWorblehat:
OK @davidjgoss, @mattwynne,

This is ready for you to take a look at again now (assuming CI passes). Searching or updating the status filters replaces the URL in the bowser bar, but doesn't add an additional history element each time.

There's still a fair amount of customizability, but it's all encapsulated in the `SearchQueryContext` module and everything has sensible (IMO) defaults:
  - Continuing to use `QueriesWrapper` as before, with either a string `query` prop or no such prop, will result in the old behaviour (i.e. no interaction with document URL). Hence, any third party uses of cucumber/react in the wild should continue to work without modification.
  - The `main.tsx` in `html-formatter` passes the result of `searchFromURLParams()` into `QueriesWrapper` props, which binds the search and filters to browser URL query parameters, using default parameter names.
  - Advanced uses (e.g. custom URL parameter names, binding the search to some other object in an embedding app) are catered for by passing options to `searchFromURLParams(...)` or instead manually setting the `{query, hiddenStatuses, onSearchQueryUpdated}` props on `QueriesWrapper`.

If you still think there's too much flexibility here, I'm happy to defer to your judgement and take it out. However, I'd always make that case that a component library shouldn't be too opinionated about how it's being used, particularly WRT interaction with global state such as the browser URL.

davidjgoss (commented):
Thanks for updating @hWorblehat.

Something that didn't really occur to me before was the disparity between what the user does with filters and how that's expressed in the URL.

For example, when the run includes statuses of PASSED, FAILED and UNDEFINED and we select:

image

In the URL we see:

```
?hide=unknown&hide=undefined
```

What would you think about something more like one of:

```
?only=passed&only=failed

?only=passed,failed
```

I think that would be more intuitive as it lines up with the filters UI, and make the special case for UNKNOWN an internal detail. Also, the filters wouldn't be reliant on which statuses were present in the test run - meaning for example you could have a bookmark to your latest build with `?only=failed` which would reliably show you only failures regardless of which other statuses were in play.

hWorblehat:
Hi @davidjgoss,

The problem generally with the status filters is that the UI only unambiguously reflects the filtering behaviour for the current set of results. For instance, if we have a set of results with only `PASSED` and `FAILED` statuses, then we only have check boxes for those 2 statuses. What should the filtering behaviour be for the `SKIPPED` status? For these results, it doesn't matter, but we need to make a decision about how it translates into URL query parameters, as that same URL may refer to a different set of results in the future.

Originally, I went with the `hide=` scheme, as I felt this had the most explicit relationship with textboxes: if a checkbox is shown and unchecked, then you get a `hide=` parameter, and if a `hide` URL parameter exists for a status that you don't get a checkbox for, that's because that status doesn't appear in the results, so it doesn't matter. Also, it has the advantage that you won't accidentally hide a status that just happens not to be present in the current set of results by toggling check-boxes and copying the URL (you have to explicitly untick a checkbox for each status you want to hide). It feels safer to construct a URL that by default _shows_ statuses that the user hasn't explicitly set a check-box for. However, I do agree that it's jarring, particularly WRT the unknown status (more on that below).

Currently, I have added a change that uses your `only=` suggestion:
  - If all shown checkboxes are ticked, there is no query parameter and no filtering, even if the URL later points to a result set that includes more statuses.
  - If at least one shown checkbox is unticked, then there will be a query parameters and filtering to only show those statuses for which there _are_ ticked checkboxes.
  - A URL with a single `only=` parameter with nothing after the `=` corresponds to all checkboxes being unticked, and no results being shown.
  - If the URL includes an `only=` parameter for a status that isn't present in the current set of results, then the filtering will still be applied accordingly, but as usual there will be no checkbox for that status. As soon as the user changes any of the other checkboxes in the UI, that URL parameter will be removed, changing the URL to match the current state of the checkboxes only.

That last point in particular can be quite jarring, particularly if the set of results only includes a single status, but not the one you have an `only=` filter for. In that case, there are no checkboxes (old behaviour, which I've kept, only includes checkboxes when there is more than one status present), so the only way to alter the filtering is to manually change the URL. To avoid this, we could change the UI to always include checkboxes for status filters, even if there's only one – we couldn't do it conditionally on whether there was a superfluous filter in the URL, because as soon as that checkbox was interacted with, it would disappear as the URL is brought in line with it!

I implemented the `only=` scheme because initially I thought I agreed it would be better. Having thought about it more, particularly writing this response, I've now changed my mind again. I think the `hide=` scheme is cleaner, more explicit, and safer in terms of not accidentally excluding results. Still, since I've implemented `only=` now, I'll wait for your response before reverting.

Finally, regarding `unknown`. My original reading of the old code was that `unknown` statuses are always silently filtered out. Hence their annoying presence in the `hide=` filters as soon as you did something non-default. However, looking again I think I was wrong. It looks like `unknown` wasn't treated specially for filtering, except that if it was the only status present, then a checkbox would still appear for it in the search bar, giving the user the ability to filter it out (thus showing no results). Unless my reading of the old code is wrong, I can't understand why we would want this behaviour at all, so my current proposal would be to remove all special handling for `unknown`, whichever URL scheme we choose. Am I missing something here?

aurelien-reeves:
I must say that the `?hide=` approach does not bother me.
If feels like 'I am opting-in to hide some specific statuses as they are all visible per default'.

davidjgoss:
Good points @hWorblehat @aurelien-reeves 

> I implemented the only= scheme because initially I thought I agreed it would be better. Having thought about it more, particularly writing this response, I've now changed my mind again. I think the hide= scheme is cleaner, more explicit, and safer in terms of not accidentally excluding results. Still, since I've implemented only= now, I'll wait for your response before reverting.

Sure, let's revert if it looks like it's not an improvement. Really appreciate you giving it a run though!

> Finally, regarding unknown. My original reading of the old code was that unknown statuses are always silently filtered out. Hence their annoying presence in the hide= filters as soon as you did something non-default. However, looking again I think I was wrong. It looks like unknown wasn't treated specially for filtering, except that if it was the only status present, then a checkbox would still appear for it in the search bar, giving the user the ability to filter it out (thus showing no results). Unless my reading of the old code is wrong, I can't understand why we would want this behaviour at all, so my current proposal would be to remove all special handling for unknown, whichever URL scheme we choose. Am I missing something here?

No, I agree, let's remove the treatment of unknown. Thanks!

hWorblehat:
Reverted back to using `hide=`. Awaiting final approval...

html-formatter/CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+* Allow searching in HTML report via URL search param
```
aurelien-reeves, line 12:
```suggestion
* Add filters and highlighted search terms in URL query parameter of HTML reports to make sharing those easier
```

I would be more user centric here. What do you think?

react/javascript/package.json
```diff
@@ -12,7 +12,7 @@
   "license": "MIT",
   "scripts": {
     "prepare": "make .codegen && npx -y cpy-cli '**/*.scss' '../dist/src' --cwd='src' --parents",
-    "test": "mocha",
+    "test": "mocha --exit",
```
aurelien-reeves, line 15: The need for the `--exit` option indicates we may have some async work still waiting in the background.
Did you need this even before adding your own tests?

aurelien-reeves (commented):
It looks pretty good to me.
But I would prefer @davidjgoss @charlierudolph or @aslakhellesoy to review it too. They are more experienced with that package than I am.

react/javascript/package.json
davidjgoss: Removing `--exit` didn't cause an issue for me locally. Just pushed a commit to remove it, let's see what happens in CI.

hWorblehat: Without it mocha was hanging for me (when running via `make`). Must be the version of Node I'm using or something else. TBH I didn't investigate whether it was code/tests that I'd added that was causing the hang.

I had a lot of problems generally with building. For instance, after changing code in the react library, running `PACKAGES="messages react html-formatter" LANGUAGES=javascript make` didn't seem to update html-formatter accordingly. I then tried running `make clean` first, but that caused the subsequent build to fail. In the end, my rebuild cycle was:
```
git clean -dfX
PACKAGES="messages react html-formatter" LANGUAGES=javascript make
```
which is somewhat annoying as it meant I had to reinstall NPM packages every time. Dunno what I was doing wrong, if anything.

hWorblehat: Also, I couldn't find it documented anywhere, but trial and error suggests that the build system specifically requires NPM v7.

davidjgoss: > Also, I couldn't find it documented anywhere, but trial and error suggests that the build system specifically requires NPM v7.

You're right and we do need to document that. Those changes happened after you first raised this PR but the devex is quite a bit simpler for JS as a result. From the root of the project you can run `npm i && npm run build` and then you should just be able to run tests etc from any module without any extra building/linking/installing.

davidjgoss (approved):

hWorblehat:
:+1:  Thanks guys

davidjgoss:
Thank you @hWorblehat!

mattwynne:
Thanks for working with us on this @hWorblehat - it's a great feature for the HTML report to have!

Thanks to you too @davidjgoss for keeping the feedback rolling!

PR #1308, opened 2021-01-14
html-formatter: reclassify webpack-bundled dependencies to be dev dependencies

davidjgoss:
## Summary

Change `dependencies` that are wrapped into the webpack bundle to be `devDependencies`, to reduce unnecessary dependency footprint of cucumber-js and projects which consume it.

## Motivation and Context

Because `cucumber-js` depends on `@cucumber/html-formatter`, it brings any `dependencies` with it. Currently that includes several cucumber packages plus `react` and `react-dom` which isn't necessary since these are inlined into the JS bundle by webpack so we don't need them when we're using the already-built package. It's also confusing for users and the extend dependency tree can cause other issues. See:

- https://github.com/cucumber/cucumber-js/issues/1516
- https://github.com/cucumber/cucumber-js/issues/1536

## Types of changes

## Checklist:


html-formatter/javascript/package.json
```diff
@@ -21,18 +21,15 @@
     "build": "tsc && webpack-cli && cp src/index.mustache.html dist/src"
   },
   "dependencies": {
```
davidjgoss, line 23: The ones left in `dependencies` are those used by the CLI and/or the CucumberHtmlStream.

davidjgoss:
Almost missed the runtime import of CSS from `@cucumber/react` in https://github.com/cucumber/cucumber/blob/master/html-formatter/javascript/src/cli-main.ts#L33

To address we'll either have webpack handle the CSS, or maybe just create a copy of that file as part of this package's build.

PR #1323, opened 2021-01-28
cucumber-expressions/java: Expose parser api

laeubi:
## Summary

as discussed with @mpkorstanje I'd like to reuse the parsing of expressions in the cucumber-eclipse project to supply the user with advanced editing capabilities.

## Included changes
- Ast parser class is public now
- Expression's carry the type of the expression + Expressionfactory has methods to guess if a string is a Cucumber or regular expression
- GroupBuilder carries information about the start/end index, this simplifies the parser in cucumber-expression and can be reused by downstream projects to get the bounds of capture groups

### DIFF
--- a/cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/Ast.java
+++ b/cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/Ast.java
@@ -1,14 +1,17 @@
 package io.cucumber.cucumberexpressions;
 
+import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.joining;
+
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.StringJoiner;
 
-import static java.util.Arrays.asList;
-import static java.util.Objects.requireNonNull;
-import static java.util.stream.Collectors.joining;
+import org.apiguardian.api.API;
 
-final class Ast {
+@API(status = API.Status.INTERNAL)
+public final class Ast {
 
     private static final char escapeCharacter = '\\';
     private static final char alternationCharacter = '/';
@@ -17,14 +20,14 @@ final class Ast {
     private static final char beginOptionalCharacter = '(';
     private static final char endOptionalCharacter = ')';
 
-    interface Located {
+	public interface Located {
         int start();
 
         int end();
 
     }
 
-    static final class Node implements Located {
+	public static final class Node implements Located {
 
         private final Type type;
         private final List<Node> nodes;
@@ -33,7 +36,7 @@ final class Ast {
         private final int end;
 
         Node(Type type, int start, int end, String token) {
-            this(type, start, end, null, token);
+			this(type, start, end, Collections.emptyList(), token);
         }
 
         Node(Type type, int start, int end, List<Node> nodes) {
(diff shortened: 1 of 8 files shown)


cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/Ast.java
```diff
@@ -17,14 +20,14 @@
     private static final char beginOptionalCharacter = '(';
     private static final char endOptionalCharacter = ')';
 
-    interface Located {
+	public interface Located {
```
mpkorstanje, line 23: Please use spaces for indent rather then tabs.

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/Ast.java
```diff
@@ -33,7 +36,7 @@
         private final int end;
 
         Node(Type type, int start, int end, String token) {
-            this(type, start, end, null, token);
+			this(type, start, end, Collections.emptyList(), token);
```
mpkorstanje, line 39: ??

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/CucumberExpression.java
```diff
@@ -1,27 +1,26 @@
...
 import static io.cucumber.cucumberexpressions.CucumberExpressionException.createParameterIsNotAllowedInOptional;
-import static io.cucumber.cucumberexpressions.ParameterType.isValidParameterTypeName;
 import static io.cucumber.cucumberexpressions.UndefinedParameterTypeException.createUndefinedParameterType;
 import static java.util.stream.Collectors.joining;
 
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Function;
+import java.util.regex.Pattern;
```
mpkorstanje, line 18: Don't reorder imports when not needed.

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/Expression.java
```diff
@@ -1,17 +1,23 @@
...
-import java.util.Set;
 import java.util.regex.Pattern;
 
+import org.apiguardian.api.API;
+
 @API(status = API.Status.STABLE)
 public interface Expression {
+
+	enum ExpressionType {
+		CUCUMBER, REGULAR;
```
mpkorstanje, line 13: This can be derived from the class of the expression.

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/Ast.java
```diff
@@ -1,14 +1,17 @@
...
 import java.util.Objects;
 import java.util.StringJoiner;
 
-import static java.util.Arrays.asList;
-import static java.util.Objects.requireNonNull;
-import static java.util.stream.Collectors.joining;
+import org.apiguardian.api.API;
 
-final class Ast {
+@API(status = API.Status.INTERNAL)
```
mpkorstanje, line 13: This should be experimental. The eclipse plugin is not internal to the cucumber expressions library.

mpkorstanje (commented):
Please keep the changes minimal. There are many unrelated changes in this PR.

> as discussed with @mpkorstanje I

While I'm an important contributor to `cucumber-expressions` this doesn't provide sufficient explanation as to why the changes are made. There are other contributors too. and you should explain the rationale to them aswel.

laeubi:
Thanks for the pointers, i'll update the PR, maybe it is a good idea to even split it into two parts, one for the CcumberExpressionparsing and one for the regex parsing...

laeubi:
I'll open smaller PRs for the different changes see https://github.com/cucumber/cucumber/pull/1324

PR #1324, opened 2021-01-30
cucumber-expressions: mark GroupBuilder and TreeRegexp as experimental API

laeubi:
## Summary

This makes the GroupBuilder/TreeRegexp available as experimental API to allow reuse in other sub-projects, also start/end index was added to retain the positional information for descending processing steps.

## Motivation and Context

I'm currently overhaul the [cucumber-eclipse plugin](https://github.com/cucumber/cucumber-eclipse) to provide better editing capabilities and user experience.

There are these main goals:

1. reuse as much code as possible from cucumber-jvm to prevent diverging results between IDE and the actual cucumber run and reduce code-duplication/implementation of features already available (especially parsers)
2. provide a better user-experience, e.g user should be able to tab-trhough parameters and simply type instead of placing cursor manually

With this change it is possible to use the same regex group parsing that cucumber-jvm uses.

## How Has This Been Tested?

I use a modified version of cucumber in development.

grafik


aslakhellesoy:
Seems like a good idea to me.

Could you please reformat the code according to existing indentation conventions?

laeubi:
@aslakhellesoy is there some formating-profile for eclipse available somewhere? Or what specific rules have to be applied (I can the try to create one on my own)

aslakhellesoy:
I'm afraid we don't have a formatting profile. Feel free to copy over the one from cucumber-jvm that @mpkorstanje put in place a few months ago: https://github.com/cucumber/cucumber-jvm/commit/f1e0b3c4cb902ad5222e8f17ffe2c757844dde28

mpkorstanje:
That one might be a bit aggressive. This repo is using the defaults from IDEA. When it comes to spacing and such the editor config should be sufficient. Try not to change any unrelated lines or revert them before committing.

https://github.com/cucumber/cucumber-jvm/blob/main/.editorconfig

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/GroupBuilder.java
```diff
@@ -1,14 +1,25 @@
...
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.regex.Matcher;
 
-final class GroupBuilder {
+import org.apiguardian.api.API;
+
+@API(status = API.Status.EXPERIMENTAL)
+public final class GroupBuilder {
```
mpkorstanje, line 12: Semantically when exposed this wouldn't be a builder but just a `Group`.

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/GroupBuilder.java
```diff
@@ -23,22 +34,18 @@ Group build(Matcher matcher, Iterator<Integer> groupIndices) {
...
-    void setNonCapturing() {
-        this.capturing = false;
-    }
-
-    boolean isCapturing() {
+	public boolean isCapturing() {
         return capturing;
     }
 
-    public void moveChildrenTo(GroupBuilder groupBuilder) {
```
mpkorstanje, line 34: Nice catch.

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/TreeRegexp.java
```diff
@@ -41,18 +39,13 @@ private static GroupBuilder createGroupBuilder(Pattern pattern) {
             } else if (c == ']' && !escaping) {
                 charClass = false;
             } else if (c == '(' && !escaping && !charClass) {
-                groupStartStack.push(i);
-                boolean nonCapturing = isNonCapturingGroup(source, i);
-                GroupBuilder groupBuilder = new GroupBuilder();
-                if (nonCapturing) {
-                    groupBuilder.setNonCapturing();
-                }
```
mpkorstanje, line 49: Why was this section deleted?

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/TreeRegexp.java
```diff
@@ -24,13 +24,11 @@
...
         this.pattern = pattern;
-        this.groupBuilder = createGroupBuilder(pattern);
+		this.groupBuilder = createGroupBuilder(pattern.pattern());
     }
 
-    private static GroupBuilder createGroupBuilder(Pattern pattern) {
-        String source = pattern.pattern();
-        Deque<GroupBuilder> stack = new ArrayDeque<>(singleton(new GroupBuilder()));
-        Deque<Integer> groupStartStack = new ArrayDeque<>();
+	public static GroupBuilder createGroupBuilder(String source) {
```
mpkorstanje, line 30: We should consider moving this method to group so the api becomes `Group.from(Pattern)`.

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/TreeRegexp.java
```diff
@@ -24,13 +24,11 @@
 
     TreeRegexp(Pattern pattern) {
         this.pattern = pattern;
-        this.groupBuilder = createGroupBuilder(pattern);
+		this.groupBuilder = createGroupBuilder(pattern.pattern());
     }
 
-    private static GroupBuilder createGroupBuilder(Pattern pattern) {
```
mpkorstanje, line 30: This method assumes that a valid regex is provided. Rather then dealing with parsing a regex internally we can ensure the regex is valid by making the caller deal with that.

mpkorstanje (commented):
I'm having a hard time following these changes. Can you make sure you don't change whitespace and unrelated lines? If you are changing multiple things at once try to do them in seperate commits so it is clear which changes go together (e.x the changing group builder constructor vs opening up the api).

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/TreeRegexp.java
laeubi: Its just a bit strange to require parsing it into a regexp and then unwraping it to get the text :-)
I would argue that passing invalid data on the parser does not necessarily yields valid results.

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/TreeRegexp.java
laeubi: With the adjusted GroupBuilder this is no longer necessary (all required data is passed in the constructor)

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/TreeRegexp.java
mpkorstanje: A method should have well defined behavior. In this case its behavior is not well defined when the precondition of a valid regex is violated. We can enforce this by using the type system. That this results in parsing a regex, and then parsing the regex again doesn't seem relevant.

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/GroupBuilder.java
laeubi: The problem is that there is already a "Group" class, the GroupBuilder builds Group(s) out of the Matched groups of a RegExp. I can rename thsi to TreeGroup if you think thats suits better?

laeubi: Or do you think the (internal) GroupBuilder should be converted to public Groups instances as they have very similar API?

laeubi:
Sorry for any inconvenience. I have created a new [PR](https://github.com/cucumber/cucumber/pull/1337) where I even was able to not introduce new API, split into individual commits and hopefully now did not include any unrelated codechanges. The formatting was applied according to the supplied IDEA profile (white spaces versus tabs).

PR #1337, opened 2021-02-04
cucumber-expressions/java: Expose regex group parser api

laeubi:
## Summary

Make the RegExp Group parsing of Cucumber accessible in a way  that it could be used for descending processing steps.

## Motivation and Context

I'm currently overhaul the [cucumber-eclipse plugin](https://github.com/cucumber/cucumber-eclipse) to provide better editing capabilities and user experience.

There are these main goals:

1. reuse as much code as possible from cucumber-jvm to prevent diverging results between IDE and the actual cucumber run and reduce code-duplication/implementation of features already available (especially parsers)
2. provide a better user-experience, e.g user should be able to tab-trhough parameters and simply type instead of placing cursor manually

With this change it is possible to use the same regex group parsing that cucumber-jvm uses.

## How Has This Been Tested?

I use a modified version of cucumber in development.

grafik

### DIFF
--- a/cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/Group.java
+++ b/cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/Group.java
@@ -3,9 +3,13 @@ package io.cucumber.cucumberexpressions;
 import org.apiguardian.api.API;
 
 import java.util.List;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import static java.util.Collections.singletonList;
+
+import java.util.ArrayList;
+import java.util.Collection;
 
 @API(status = API.Status.STABLE)
 public class Group {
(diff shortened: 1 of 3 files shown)


cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/Group.java
```diff
@@ -43,4 +47,27 @@ public int getEnd() {
...
+     * @param expression the expression to decompose
+     * @return A collection of {@link Group}s, possibly empty but never
+     *         <code>null</code>
+     */
+    public static Collection<Group> parse(Pattern expression) {
+        GroupBuilder builder = TreeRegexp.createGroupBuilder(expression);
+        return toGroups(builder.getChildren());
+    }
+
+    private static List<Group> toGroups(List<GroupBuilder> children) {
```
mpkorstanje, line 63: Neat!

mpkorstanje (approved):
LGTM!

laeubi:
Any idea whats wrong with the CI server? Can the result be displayed without sign in to circle-ci (I don't like to give access to all my repositories just to see an error message...)

mpkorstanje:
Looks like the network failed during the build. I've restarted the job.

> Can the result be displayed without sign in to circle-ci 

@aslakhellesoy  it's time to consider github actions. Do we track this somewhere?

mpkorstanje:
I think we have a few more changes you wanted to make, after we've merged those I  can make a new release.

laeubi:
Thanks, I will open another PR soon...

PR #1359, opened 2021-02-12
[gherkin-perl] Create Gherkin API & better releases

ehuelsmann:
## Summary

Perl's Gherkin implementation lacked the central API that e.g. Java and Ruby have, to spawn a message stream off from a list of files. This PR adds that. Using that API, it consolidates `bin/gherkin-generate-{ast,pickles}` into `bin/gherkin`, just like how Java and Ruby are organized.

Based on my experience changing the source code and dependencies, improve the installation and checking of dependencies.

## Details

There was documentation saying this implementation is a line-by-line port of the Python implementation. Because I'm using Ruby and Java as references now (as discussed), it's no longer fair to say that this implementation takes to the Python one. Additionally, since this version is developed and subject to the release policy of the Cucumber project, there's no need to mention anything about specific version numbering.

## Motivation and Context

Having an API is really the point of having the Gherkin library.

## How Has This Been Tested?

By merging the `bin/gherkin-generate-*` tools into the new `bin/gherkin` executable and driving that from the test suite in the Makefile, I'm certain this change delivers at least the same experience as the prior implementation.

## Types of changes

## Checklist:

- [X] I have updated the documentation accordingly.


ehuelsmann:
@aslakhellesoy , this branch also holds the fixes I had to apply today to fix the Perl release process.

PR #1360, opened 2021-02-12
Spanish translation for Feature and Rule synonym keywords

sergioforerogomez:
## Summary

Translated keywords

## Motivation and Context

Translations missing

## How Has This Been Tested?

It hasn't

## Types of changes

## Checklist:

- [X] The change has been ported to .NET.
- [X] I have updated the CHANGELOG accordingly.


aslakhellesoy:
Muchas gracias!

I'm not sure it is a good idea to have two synonyms where one is a substring of the other. Do we really need both `Necesidad del negocio` and `Necesidad`? And both `Regla de negocio` and `Regla`?

sergioforerogomez:
> Muchas gracias!
> 
> I'm not sure it is a good idea to have two synonyms where one is a substring of the other. Do we really need both `Necesidad del negocio` and `Necesidad`? And both `Regla de negocio` and `Regla`?

Hi! 

In Spanish it isn't as descriptive as we would like to have just `Regla`, so we can start working with Cucumber 6 rules.
And, we can't just replace it, that's why.

What would you think if I take just one for `Necesidad del negocio` and `Necesidad`, and we keep both for Rules?

sergioforerogomez:
@aslakhellesoy I've made the change and all checks have passed! :D

PR #1372, opened 2021-02-19
Add npm workspaces

aurelien-reeves:
## Summary

Binging NPM 7 Workspaces into the monorepo

## Details

Refs. https://github.com/cucumber/cucumber/issues/1259

## TODO

- [x] Set up npm workspaces and `tsconfig.build.json` so we can `npm i && npm run build` all npm modules from root.
- [x] Make `npm run eslint-fix` work from root
- [x] Make `npm run upgrade` work from root
- [x] Remove common dev dependencies and configs from `*/javascript/package.json` and only declare them once in the root `package.json`
- [x] Remove "build", "lint" and "lint-fix" from all `*/javascript/package.json` files (should only be in the root one).
- [x] Ensure that the npm `prepare` lifecycle script won't run when consumers install the package
- [x] Update `.templates/javascript/default.mk`
- [x] Remove `.templates/javascript/.eslintrc.json` and all copies
- [x] Remove `.templates/javascript/.prettierrc` and all copies
- [x] Figure out what to do with the generated `messages/javascript/src/messages.js` and `messages/javascript/src/messages.d.ts` files
  - Currently, `npm i && npm run build` will fail from a fresh clone. We don't want to have to do `npm i && pushd messages/javascript && make && popd && npm run build`
- [x] Allow contributors to run TypeScript code without needing to build the code (no need to run `npm run build` from the root)
  - [x] Update all `tsconfig.json` files to just `{ "extends": "../../tsconfig.json" }`
  - [x] Set up `compilerOptions.paths` in root `tsconfig.json` file
  - [x] Set up [tsconfig-paths](https://github.com/dividab/tsconfig-paths)
- [x] Update `.circleci/config.yml`
  - Can we still build npm modules independently?
  - If yes, do they need to depend on a job that does the `npm install && npm run build`?
  - What's left to do in npm module CI jobs? Just `npm test`?
- [x] Update the monorepo build
  - Need to tun `npm i` in root. Running `npm run build` should be the last step. Run `npm test` to test all modules.
- [ ] Update contributor docs
- [x] Update `.templates/javascript/default.mk`
  - [x] Remove `remove-local-dependencies` 
- [x] Decide what to put in `./tsconfig.base.json` and `./tsconfig.build.json` - at the moment some duplication remains in those files
- [x] Add missing files (`.rsync`) to the new packages (`message-streams`, `gherkin-streams`)
- [x] Decide whether we should move stream classes for other implementations as well (to remain consistent)
- [x] Make sure `gherkin` acceptance tests are running


davidjgoss:
> Figure out what to do with the generated `messages/javascript/src/messages.js` and `messages/javascript/src/messages.d.ts` files...Currently, npm i && npm run build will fail from a fresh clone. We don't want to have to do `npm i && pushd messages/javascript && make && popd && npm run build`

In the messages/javascript package.json:

```json
{
  "scripts: {
    "prepare": "npm run pbjs && npm run pbts",
    ...
  }
}
```

That should run automatically after `install` in a local/workspace context. See https://docs.npmjs.com/cli/v7/using-npm/scripts#life-cycle-scripts.

aslakhellesoy:
> > Figure out what to do with the generated `messages/javascript/src/messages.js` and `messages/javascript/src/messages.d.ts` files...Currently, npm i && npm run build will fail from a fresh clone. We don't want to have to do `npm i && pushd messages/javascript && make && popd && npm run build`
> 
> In the messages/javascript package.json:
> 
> ```json
> {
>   "scripts": {
>     "prepare": "npm run pbjs && npm run pbts",
>     ...
>   }
> }
> ```
> 
> That should run automatically after `install` in a local/workspace context. See https://docs.npmjs.com/cli/v7/using-npm/scripts#life-cycle-scripts.

My understanding of npm workspaces is that you're only supposed to run `npm install` from the root directory. I wonder if this skips running of individual modules' npm lifecycle scripts.

Any idea?

davidjgoss:
Looks like the answer is "yes since we fixed it recently" https://github.com/npm/cli/issues/1965.

I'll push a commit and we find out?

davidjgoss:
Worked for me locally with a fresh clone! Also npm 7 is very fast.

react/javascript/Makefile
```diff
@@ -7,7 +7,7 @@ TS_MESSAGE_FILES = $(patsubst ../../compatibility-kit/javascript/features/%.ndjs
 
 dist/src/styles/cucumber-react.css: src/styles/styles.scss src/styles/react-accessible-accordion.css
 	mkdir -p $(@D)
-	./node_modules/.bin/sass $< > $@
+	../../node_modules/.bin/sass $< > $@
```
aurelien-reeves, line 10: What about using `npx sass` instead of a relative path to the binary?

davidjgoss: and/or could we move to the `package.json`, maybe another candidate for using `prepare`?

.gitignore
```diff
@@ -1,4 +1,6 @@
 .DS_Store
 .rsynced
 .idea/
-executables
\ No newline at end of file
+executables
+node_modules
+package-lock.json
```
davidjgoss, line 6: Ideally we'd want the lockfile version controlled for stability. I know we have historically not done, was this perhaps a practical issue that may go away with this change?

tsconfig.build.json
```diff
@@ -0,0 +1,44 @@
+{
+  "extends": "./tsconfig.base.json",
+  "compilerOptions": {
+    "composite": true,
+    "declaration": true,
+    "declarationMap": true,
+    "sourceMap": true,
+    "target": "es6",
```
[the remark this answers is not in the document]
davidjgoss: Although I guess we need to ensure some packages work in browser. Do we leave it up to consuming projects to do transpilation?

.gitignore
aslakhellesoy: I agree - let's commit `/package-lock.json`

aslakhellesoy: Done!

tsconfig.build.json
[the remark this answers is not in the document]
aslakhellesoy: > Given Node 10 is our lowest common denominator

Is it though? Some modules also need to work in the browser, such as `cucumber/@react` and all its dependencies. I also know that some teams use `cucumber/@gherkin` in the browser - I recently decoupled node-specific libraries to make this possible without polyfills.

So with that in mind, what do you think we should target?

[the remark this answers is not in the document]
davidjgoss: Yeah I remembered about browsers right after I hit Comment :)

I'd like to think we could just ship ES[current] code and have consuming projects do the transpiling they need, but I don't think the ecosystem is there yet. So:

- `es6` assuming we aren't worried about IE
- `es5` if we _do_ need to worry about IE

bbros-dev:
The details provided by @aslakhellesoy [here](https://github.com/cucumber/cucumber/issues/1400#issuecomment-792117710) and @davidjgoss, [here](https://github.com/cucumber/cucumber/issues/1400#issuecomment-791253706),  would be useful inclusions in the existing item:

- [ ] Update contributor docs

PR #1391, opened 2021-02-27
react: styling improvements for gherkin components

davidjgoss:
## Summary

Improve styling and CSS architecture of gherkin components in `@cucumber/react`.

Introduce an MDG component for rendering Markdown features.

## Details

Styles are colocated and encapsulated with presentational components via (S)CSS modules. This means we are now producing processed CSS output via webpack in the `html-formatter` build, not just copying some stylesheets from `react` - this is reflected in some makefile and code changes.

An optional `CustomRendering` component enables customisation of styles via providing own CSS classes, or even overriding of React rendering. This is intentionally for a small subset of presentational
components at this point.

Theming is enabled via CSS custom properties - we could conceivably let people set these via `--formatOptions`.

More detail in the updated `README.md`. There are also new stories in Storybook illustrating what's possible with the customisation.

## Motivation and Context

* Improve general look and feel
* Improve component encapsulation
* Allow incremental customisation/theming

118369710-5590a300-b59c-11eb-9f72-a626c97ea4ea

You can preview here in Storybook form:
https://serene-wilson-f8fc77.netlify.app/

## Types of changes

## Checklist:


aurelien-reeves:
What would you think about the possibility to propose several styles through optional themes?

davidjgoss:
@aurelien-reeves yes definitely, that should be nice and easy with custom properties.

I’ll run an example up today and work on the CSS modularisation too.

@aslakhellesoy raised this in lieu of an issue, hope that’s okay?

davidjgoss:
I've added CSS modules now - see `DocString` for initial example. Once rolled out this will mean we no longer need a global stylesheet as each component will have its own styles colocated, and encapsulated at build time (I'll need to update html-formatter to handle this with webpack as well).

Customisation comes from CSS custom properties, which could be set from a custom stylesheet, plus built-in themes can be activated via the `<CucumberTheme>` component (for example in html-formatter we could get this from a CLI option).

Inevitable dark theme:

image

aslakhellesoy:
OMG this looks awesome! 😍 And very clean scss + minimal logic.

I'd love to make it obvious to consumers of `@cucumber/react` to take complete control over the styling.

Do you think it makes sense to use TailwindCSS to build the default stylesheet? Would we just use a lot of `@apply` in the definitions of the classes we have to define? Is there anything to gain from TailwindCSS?

I've had another thought about styling. It's a inspired by [React-Markdown Custom Renderers](https://github.com/remarkjs/react-markdown#use-custom-renderers-syntax-highlight). A custom renderer would typically use a component from the library, and pass custom `class` attributes. That way, a TailwindCSS renderer would add a bunch of tailwind classes. Another one could just add a simple class name and do the styling in a custom stylesheet. Each to their own.

I think it's important to get this styling story straight. That way it will become much easier to adopt and tailor to users' needs.

LMK what you think

davidjgoss:
@aslakhellesoy yep I know what you mean. If we could make that `renderer` a prop of `<CucumberTheme>` then potentially we could handle it internally with a context. I'll have a play over the weekend and see what I can do.

davidjgoss:
@aslakhellesoy @aurelien-reeves 

I've pushed a first pass at a custom styling/rendering and added docs to the readme that describe it. Basically it allows for overriding the classnames or even overriding the component implementation. Not sure if the latter is a bridge too far - but it would allow for things like people overriding the `Tags` component to render links based on their tagging conventions (JIRA identifiers etc).

A bit of refinement is probably needed to make this more usable in the real world e.g.:

- When overriding classes, merge the custom class names object onto the built-in one, so users can still take advantage of the built-in styles except the exact ones they want to override
- When overriding components, make the built-in styles available to use if they want to

Anyway, let me know what you think!

aurelien-reeves:
@davidjgoss 
Personally, I like it! 👍 

You pushed the concept of custom styling to its last retrenchments keeping the PR decent.
And your documentation in the README is pretty clear, well done!

aslakhellesoy:
Hi @davidjgoss - I added `custom.d.ts` to get past the `tsc` compilation error.

Now it fails when mocha is run. The error is the same as the one I got from `tsc` before adding `custom.d.ts`. @aurelien-reeves found https://stackoverflow.com/a/63789958/99475 and suggested we can get past that error by adding the following to `react/javascript/tsconfig.json`:

```
"ts-node": {
    "files": true
 }
```

That got us to yet another error:

```
$ mocha

/Users/Aslak.Hellesoy/git/cucumber/cucumber/react/javascript/src/components/gherkin/DocString.module.scss:1
@import '../../styles/theming';
^

SyntaxError: Invalid or unexpected token
    at Object.compileFunction (node:vm:355:18)
    at wrapSafe (node:internal/modules/cjs/loader:1022:15)
    at Module._compile (node:internal/modules/cjs/loader:1056:27)
```

We're not sure how to get past that one. We've tried https://github.com/mrmckeb/typescript-plugin-css-modules without much luck.

Any ideas?

davidjgoss:
Tests working now via `ignore-styles` which basically causes any stylesheet imports to just resolve to an empty object, similar to webpack's `null-loader`.

(Installed this at the root, will revert that and install at the package level once I get past some local issues.)

davidjgoss:
(Planning to get back to this properly from Thursday in terms of finishing the restyling and modularisation. Examples tables are going to be the most fun!)

react/javascript/.rsync
```diff
@@ -1,3 +1,3 @@
 ../../LICENSE LICENSE
 ../../.templates/github/ .github/
-../../.templates/javascript/ .
+../../.templates/javascript/ . --exclude .mocharc.json --exclude tsconfig.json
```
davidjgoss, line 3: Not ideal to opt out of the sync here, but I do think `react/javascript` is a bit of a special case?

react/javascript/src/components/customise/CustomRendering.tsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react'
+import { messages } from '@cucumber/messages'
+
+export function mixinStyles(
+  builtIn: Record<string, string>,
+  custom?: Record<string, string>
+): Record<string, string> {
+  const mixed: Record<string, string> = {}
+  Object.keys(builtIn).forEach((key) => {
```
davidjgoss, line 9: Could simplify to use `Object.assign` if we aren't concerned with supporting IE.

Are we?

aurelien-reeves: I think that it could be managed by the transpiler(s) with some options?

In the code, I suggest we don't care, but we make sure we know how it looks like after transpilation.

Regarding the question regarding IE support, do we know how it is supported now? Does react - and other dependencies we rely on - actually still support it?

davidjgoss: Good point about transpilation. Because this is the `Object` API rather than a language/syntax feature, I wouldn't think transpilation would cover it, it would be more like a polyfill. I'll check it out though.

davidjgoss: In terms of React, implication is that IE11 is fine (for now) https://reactjs.org/docs/react-dom.html#browser-support.

davidjgoss:
Discussed at open source day:

- Include tailwind in the cucumber-react and html-formatter stack so we can port some of the improved styling of the "chrome" around the report (meta, search etc) over from Cucumber Reports

react/javascript/.rsync
aslakhellesoy: Neat trick!

react/javascript/README.md
```diff
@@ -34,6 +34,98 @@ This is fine for simple use cases where results are not important.
...
+</CustomRendering>
+```
+
+Some components have multiple styling hooks - e.g. the `<Tags>` component has the `tags` class name for the list, and the `tag` class name for each item. In these cases, you can provide custom class names for just the ones you want to change, and any you omit will pick up the built-in styling like normal.
+
+### Custom rendering
+
+To change the rendering of some components entirely, you can selectively provide your own component implementations to be used instead of the built-in ones.
+
+Staying with the doc string example, you can use the same `CustomRendering` wrapper, but this time instead of an object with class names, you provide a React functional component, giving you full control over the rendering:
```
aslakhellesoy, line 114: Love this! I assume this makes https://github.com/cucumber/cucumber/pull/1179 obsolete?

react/javascript/package.json
```diff
@@ -11,7 +11,7 @@
   "author": "Aslak Hellesøy",
   "license": "MIT",
   "scripts": {
-    "prepare": "make .codegen",
+    "prepare": "make .codegen && npx -y cpy-cli '**/*.scss' '../dist/src' --cwd='src' --parents",
```
aslakhellesoy, line 14: I think we should add `cpy-cli` to `devDependencies` to avoid the prompt if we want to install it?

react/javascript/src/components/customise/CustomRendering.tsx
```diff
@@ -0,0 +1,69 @@
...
+}
+
+export declare type CustomRenderer<R, C> = React.FunctionComponent<R> | Partial<C>
+
+export interface DocStringProps {
+  docString: messages.DocString
+}
+
+export interface DocStringClasses {
+  docstring: string
```
aslakhellesoy, line 27: should this be `docString`

react/javascript/README.md
davidjgoss: Yep, I think so.

aslakhellesoy:
Hi @davidjgoss - I've brought this uptodate with a few big changes that have gone into master. What's left to do before we can merge this? How can I help?

davidjgoss:
@aslakhellesoy 

Thanks for updating, and agreed on deferring work on the "chrome" to a future change.

I'm just working through the rest of the Gherkin component styling in terms of getting it looking good, encapsulating the styles and making it customisable - should be ready to go by Monday/Tuesday.

One question about `GherkinDocumentList` - do we think this needs to support custom rendering? It contains some business logic so I think if we wanted to allow users to replace it we'd probably want to break that stuff away from the presentational (accordion) stuff. Maybe we only allow custom CSS classes in this case?

BTW I deployed the Storybook output here so people can preview easily (I'll update it as I do more refinement):
<https://serene-wilson-f8fc77.netlify.app/>

aslakhellesoy:
Personally I think I'd like to get rid of the accordion - I don't think it's an efficient way to navigate, especially with reports that have many files. You have to scroll too much to find what you want.

I think a scrollable tree navigation (Feature->Scenario->Step) on the left and a document on the right might work better. The index page (no document selected) could display a summary.

Not suggesting we do that in this PR, but if you agree we don't need to worry too much about GherkinDocumentList.

WDYT?

davidjgoss:
@aslakhellesoy agreed re accordion, I've changed `GherkinDocumentList` to not be customisable.

davidjgoss:
Almost done here. Left to do:

- [ ] Cleanup of a few last styling and encapsulation issues
- [ ] Add the ability for custom components to bail out and fallback to the default rendering - e.g. provide a custom implementation for a particular kind of attachment, but carry on with defaults for images etc

react/javascript/src/components/gherkin/MDG.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react'
+import ReactMarkdown from 'react-markdown'
+import rehypePlugins from '../app/rehypePlugins'
+import GherkinQueryContext from '../../GherkinQueryContext'
+import Step from './Step'
+import styles from '../app/GherkinDocumentList.module.scss'
```
aslakhellesoy, line 6: I think we should rename these styles to `./Gherkin.module.css`

react/javascript/src/components/gherkin/MDG.tsx
```diff
@@ -0,0 +1,42 @@
...
+const MDG: React.FunctionComponent<IProps> = ({ uri, children }) => {
+  const gherkinQuery = React.useContext(GherkinQueryContext)
+
+  return (
+    <div className={styles.list}>
+      <ReactMarkdown
+        rehypePlugins={rehypePlugins}
+        components={{
+          ul({ children }) {
+            return <ul className={'cucumber-steps'}>{children}</ul>
```
aslakhellesoy, line 19: I had to add the `cucumber-steps` class here in order to get the `list-style: none` CSS rule, but the drawback is that *any* `<li>` now gets that style, including list items that are *not* Gherkin steps.

I wonder if it would be better to move this style up to the dom element that renders the enclosing `Scenario`. That way the style would only apply to Gherkin Step list items.

react/javascript/src/components/gherkin/MDG.tsx
aslakhellesoy, line 6: I think we should rename these styles to `./Gherkin.module.css`

react/javascript/src/components/gherkin/MDG.tsx
```diff
@@ -0,0 +1,42 @@
...
+          li({ node, children }) {
+            const line = node.position.start.line
+            const step = gherkinQuery.getStep(uri, line)
+            if (!step) {
+              // Non-Gherkin list item
+              return <li>{children}</li>
+            }
+            return (
+              <li>
+                <Step step={step} renderStepMatchArguments={true} renderMessage={true} />
```
aslakhellesoy, line 30: This will render the step *as well as its DataTable*. This results in the Markdown table being rendered twice (once by the `<Step>` component, and once by the Markdown renderer (where it doesn't get any styles).

Screenshot 2021-05-19 at 22 39 07


In order to avoid this double rendering I think it would be better to lift the `<DataTable>` rendering out of the `<Step>` component and have `<StepList>` render `<DataTable>` underneath the `<Step>`.

In Markdown mode, we'd let Markdown handle the rendering, but we could register custom `td`, `th` and `tr` compoents in `react-markdown`. We should consider extracting `Td`, `Th` and `Tr` components from `DataTable` so they can use the same components (and get the same styles).

aslakhellesoy (changes requested):
I took the liberty to add a new `<MDG>` component that renders [Markdown with Gherkin](https://github.com/cucumber/common/blob/master/gherkin/MARKDOWN_WITH_GHERKIN.md).

That's highlighted some issues that I think it would be good to deal with. We don't necessarily need to deal with them in this PR - it could be done in a different PR.

react/javascript/src/components/gherkin/MDG.tsx
davidjgoss: Addressed by changing to

```scss
.cucumber-steps {
  padding: 0;
  margin: 1em;

  > li {
    list-style: none;
  }
}
```

davidjgoss: Do we support using ordered lists as well with MDG?

react/javascript/src/components/gherkin/MDG.tsx
davidjgoss: Feels a little broad. Is just `DocumentList` maybe good enough?

react/javascript/src/components/gherkin/MDG.tsx
davidjgoss: Oh interesting. I guess we'd have a similar situation for a doc string - a duplicate `<pre>`?

Pulling them out of the `<Step>` would help the Markdown problem, but would also change the order of things so that the data table or doc string would appear after any attachments or errors for the step - I think this might be a bit confusing to read.

Coming at it from the opposite angle, what if we were to, for `table` and `pre`, check if the next non-whitespace line above is a list item, and if so just render nothing?

davidjgoss: FWIW pushed a commit that applies our data table style to react-markdown's table rendering.

Although looking at it now, it being presented like a Gherkin table but in a Markdown document feels a bit off 😬

react/javascript/src/components/gherkin/MDG.tsx
aslakhellesoy: You mean this?

```markdown
1. one
1. two
1. three
```

Which renders as:

1. one
1. two
1. three

MDG doesn't currently support it, but it definitely should! I'll add support for that

react/javascript/src/components/gherkin/MDG.tsx
aslakhellesoy: > ... but would also change the order of things so that the data table or doc string would appear after any attachments or errors for the step

Couldn't we address that by also lifting `<Attachment>` out of `<Step>`?

davidjgoss: Okay, so `<Step>` just becomes about the title line, the the other bits become other items in the list. Confused about how we'd still be able to maintain the order for Markdown though?

aslakhellesoy: Do you mean how we'd be able to display the error messages and attachments (runtime content) in Markdown?

I wonder if we could add some state to the `table` and `li` components and render runtime content directly after the `<Step>` (if there is no DataTable/DocString), otherwise after the DataTable/DocString.

davidjgoss: Here's the relevant excerpt of [the Markdown doc we're testing with](https://github.com/cucumber/common/blob/main/compatibility-kit/javascript/features/markdown/markdown.md):

```markdown
### Example: DataTable

Here is an example of what it looks like:

* Given this is a typical table:

| name | age |
| ---- | --: |
| Bill |   3 |
| Jane |   6 |
```

If we follow the same pattern as the steps with doc strings, and put the table immediately after the step bullet, like this...

```markdown
### Example: DataTable

Here is an example of what it looks like:

* Given this is a typical table:
  | name | age |
  | ---- | --: |
  | Bill |   3 |
  | Jane |   6 |
```

...the problem disappears, because the Markdown AST has the `table` as a child of the `li`, and we override the entire rendering of the `li` anyway because it's a step.

Would it be reasonable to mandate that in terms of what's valid MDG?

aslakhellesoy: Yes I think that’s reasonable. So we’ll only recognise datatables that have 2 or more spaces indentation? Need to check the spec....

react/javascript/src/components/gherkin/MDG.tsx
davidjgoss: Pushed something that deals with lists better - when we come to render a `ul`, we check if there's a step on that line (since the `ul` and its first `li` will have the same starting line in the source) - if so we render a steps list, otherwise just a regular one (with bullets).

aslakhellesoy:
I accidentally edited the summary in this PR thinking I was editing #1564 - I've done my best to restore it (reverting the edits wasn't possible).

aslakhellesoy:
I've made some more improvements to Markdown. The Markdown rendering and AST rendering is now very similar:

Markdown:
Screenshot 2021-05-24 at 13 08 46

Gherkin:
Screenshot 2021-05-24 at 13 09 29

aslakhellesoy:
@davidjgoss do you think we can merge this to main now?

davidjgoss:
@aslakhellesoy yep I'm happy if you are.

aslakhellesoy:
Let's do it! I'll create a new issue with various remaining styling niggles I'd like to address.

I'll merge when the CI is green.

PR #1406, opened 2021-03-05
cucumber-expressions/java: Support character in BuiltInParameterTransformer

manish7-thakur:
Add character transformer in `@cucmber/cucumber-expressions-java`

Currently we don't have character transformer so it make sense to have a transformer for this basic type.

## Summary

Just add Character/char transformer after String transformer in `BuiltInParameterTransformer`
## Motivation and Context
Solves the issue : https://github.com/cucumber/cucumber/issues/1405

## How Has This Been Tested?
Unit tests included

## Types of changes

- [X ] Bug fix (non-breaking change which fixes an issue).

## Checklist:

- [ X] The change has been ported to Java.
- [X ] I've added tests for my code.


cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/BuiltInParameterTransformer.java
```diff
@@ -42,6 +42,10 @@ private Object doTransform(String fromValue, Type toValueType, Type originalToVa
             return fromValue;
         }
 
+        if (Character.class.equals(toValueClass) || char.class.equals(toValueClass)) {
+            return fromValue.toCharArray()[0];
+        }
```
mpkorstanje, line 47: I think this should throw an exception when the captured string is not a single character.

Otherwise this may lead to surprises when the captured string is not a single character.

This may happen because the regex was capturing more then expected or because the step used a unicode codepoint that requires two bytes to represent.

mpkorstanje (changes requested):

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/BuiltInParameterTransformer.java
manish7-thakur: Ideally this should not happen, if it is an empty string it's deciphered type should be string and should not land into character type check construct. But anyways to avoid surprises I've added checks  to transform only when string length is 1 & type is `char`.

cucumber-expressions/java/src/test/java/io/cucumber/cucumberexpressions/BuiltInParameterTransformerTest.java
```diff
@@ -95,6 +95,31 @@ public void should_transform_boolean() {
...
     }
 
+    @Test
+    public void should_transform_char() {
+        assertThat(objectMapper.transform("A", char.class), is(equalTo('A')));
+        assertThat(objectMapper.transform("b", Character.class), is(equalTo('b')));
+    }
+
+    @Test
+    public void should_throw_exception_for_empty_and_nonsingelchar_string() {
```
mpkorstanje, line 105: These are two different tests.

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/BuiltInParameterTransformer.java
```diff
@@ -43,7 +43,7 @@ private Object doTransform(String fromValue, Type toValueType, Type originalToVa
         }
 
         if ((Character.class.equals(toValueClass) || char.class.equals(toValueClass)) && fromValue.length() == 1) {
-                return fromValue.toCharArray()[0];
+                return fromValue.charAt(0);
```
mpkorstanje, line 46: Little bit of a nit-pick but I initially misread this condition. Might be good to put the `fromValue.length() == 1` in a seperate if statement.

cucumber-expressions/java/src/test/java/io/cucumber/cucumberexpressions/BuiltInParameterTransformerTest.java
manish7-thakur: Sure...

cucumber-expressions/java/src/main/java/io/cucumber/cucumberexpressions/BuiltInParameterTransformer.java
manish7-thakur: Sure...

mpkorstanje (approved):
Cheers, looks good to me. Would you like to solve the same problem in Go too? (Though I can't really work out right now what would be a sensible equivalent in Go).

manish7-thakur:
> Cheers, looks good to me. Would you like to solve the same problem in Go too? (Though I can't really work out right now what would be a sensible equivalent in Go).

I think the closest equivalent in go would be a rune. Go uses unicode code points. I'll write some test first to verify if it is indeed a problem in Go as well.

manish7-thakur:
@mpkorstanje I can confirm that this is not an issue with Go library, we have a test case for exactly one rune: https://github.com/cucumber/cucumber/blob/master/cucumber-expressions/go/cucumber_expression_generator_test.go#L283. Single rune is deciphered as string and returned as string and throw any exception. Although the assertion were a bit confusing initially as `expected` & `actual` were flipped. I rectified `cucumber_expression_generator_test.go`. I also added one more example in the `example.txt` to test this rune case but CI emits an error, not sure how to fix. 
```
touch .rsynced
[[ -z $(git status -s) ]] || ( \
	echo "Working copy is dirty. Please run 'source scripts/functions.sh && rsync_files' and commit modified files." && \
	echo "Found: " && \
	git status -s && \
	exit 1 \
)
Working copy is dirty. Please run 'source scripts/functions.sh && rsync_files' and commit modified files.
Found: 
 M cucumber-expressions/go/examples.txt
```

I can revert it if not needed. Let me know your thoughts.

mpkorstanje:
The the file `cucumber-expressions/go/testdata/` is a duplicate of `/cucumber-expressions/testdata`. The script is complaining about the fact that you didn't duplicate it to the other language implementations.

mpkorstanje:
Please add entry under the fixed header in the change log.

https://github.com/cucumber/cucumber/blob/master/cucumber-expressions/CHANGELOG.md

manish7-thakur:
> The the file `cucumber-expressions/go/testdata/` is a duplicate of `/cucumber-expressions/testdata`. The script is complaining about the fact that you didn't duplicate it to the other language implementations.

Done : https://github.com/cucumber/cucumber/pull/1411/files. Pease have a look, raised as a separate PR just to separate out concerns.

aslakhellesoy:
Hi @manish7-thakur,

Aslak Hellesøy
Creator of Cucumber

Issue #1386, opened 2021-02-25
Remove runtime dependency on Protobuf


aslakhellesoy:
When we adopted protobuf as the basis for messages, we had 3 goals:

* Be able to define messages with an [IDL](https://en.wikipedia.org/wiki/Interface_description_language)
* Wide programming language support
* JSON serialisation format (for debuggability)

Protobuf satisfies all of those, but has some drawbacks.

The primary drawback is runtime dependencies. The `protoc` code generator (or similar code generators in unofficial ports) generates code that depends on protobuf runtime libraries. This poses two challenges:

* Size. Loading all this code into a browser take a long time because the resulting code is megabytes.
* Compatibility. A project depending on a different version of protobuf than the one used by Cucumber might lead to conflicts

Another challenge with protobuf is maintenance cost. Because we want to inspect messages visually, we've decided to use the JSON serialisation format instead of the more commonly used binary representation. While JSON is "supported" by protobuf, the implementations have several bugs that we've had to work around or fix. These bugs exist because few people use protobuf with JSON.

During the weekly community meeting on [25 Feb 2021](https://docs.google.com/document/d/1ewOKjaK5F17pwv7wtiaBFWCdzHQBXqQMCxqkwWKVmX8/edit#) two alternative solutions were put forward:

# Keep `.proto` but use custom code generator

We could implement a custom code generator that parses the `messages.proto` file and reads a programming language-specific template file for code generation.

The code generator would generate code with an interface similar to the code generated by `protoc`, but without any runtime dependencies. It would only support JSON serialisation and not the binary representation.

For TypeScript for example, this could be just `messages.d.ts` (just the interface) and no implementation at all. We'd get type safety at compile time, and wouldn't need any runtime dependencies for serialisation (that would be handled by the stdlibrary JSON functions).

# JSON Schema

(OpenAPI was suggested. OpenAPI has a lot of HTTP stuff in it that isn't relevant to us, and my understanding is that it [relies](http://spec.openapis.org/oas/v3.1.0#data-types) on [JSON Schema](https://json-schema.org/) to define the schema of request/response payloads).

We could [translate](https://github.com/chrusty/protoc-gen-jsonschema) the `.proto` schema to JSON Schema and use existing tools to generate lightweight code. Here are some existing ones:

* TypeScript: [json-schema-to-typescript](https://github.com/bcherny/json-schema-to-typescript)
* .NET: [NJsonSchema](https://github.com/RicoSuter/NJsonSchema)
[...]

PR #1414, opened 2021-03-10
Json schema for message protocol

aslakhellesoy:
This is a Draft PR to replace protobuf with pure JSON and JSON Schema, addressing #1386.

TODO:

* [x] TypeScript: Modify ts code generator to define a `Status` alternate type from the enum in the schema. Same for `ContentEncoding`
* [ ] Working PR for [cucumber-js](https://github.com/cucumber/cucumber-js/pull/1645)
* [ ] Working PR for cucumber-ruby
* [ ] Working PR for cucumber-jvm

Later:

* [ ] Validate all generated messages in compatibility kit with the JSON schema.
* [ ] Elixir implementation/code generator
* [ ] .NET implementation/code generator
* [x] Go implementation/code generator

We can live with a combination of protobuf/json schema implementations for some time, as long as we keep them in sync. The gherkin acceptance tests will catch any inconsistencies.


WannesFransen1994:
Not sure if it's relevant, but following libraries might be helpful for the Elixir code generation:

* [https://hex.pm/packages/ex_json_schema](https://hex.pm/packages/ex_json_schema)
* [https://hex.pm/packages/json_schema](https://hex.pm/packages/json_schema)
* [https://hex.pm/packages/typed_struct](https://hex.pm/packages/typed_struct)

aslakhellesoy:
Thanks @WannesFransen1994 we'll have a look at those.

mpkorstanje:
For java we should make sure the compiler preserves the parameter names. This will help various object mappers figure things out.

```xml

    <build>
        <pluginManagement>
            <plugins>
                <plugin>
                    <groupId>org.apache.maven.plugins</groupId>
                    <artifactId>maven-compiler-plugin</artifactId>
                    <version>3.8.1</version>
                    <configuration>
                        <parameters>true</parameters>
                    </configuration>
                </plugin>
            </plugins>
        </pluginManagement>
    </build>
```

mpkorstanje:
For java we should also get rid of Jackson. Even shaded it will trigger a frequent release cycle due to security findings.

aslakhellesoy:
What do you recommend we use instead of Jackson @mpkorstanje?

We need to be able to serialise POJOs to JSON. I don't think we need deserialisation.

aslakhellesoy:
ping @mpkorstanje - what do you think we should use instead of jackson?

We definitely need a library that can serialise objects to JSON.

I don't think we need deserialisation, so we could use a simple library that only does serialisation. I've looked at https://github.com/fabienrenaud/java-json-benchmark, and not a single one of them has serialisation only (which increases the attack-vector surface).

Shall we just go with a shaded `com.eclipsesource.minimal-json` as we use in `create-meta`?

mpkorstanje:
You want to turn abitrary POJOs into json right? `minimal-json` only supports turning json objects into json.

aslakhellesoy:
> You want to turn abitrary POJOs into json right? minimal-json only supports turning json objects into json.

Ah yes, forgot about that. Yes, we need to turn POJOs into JSON.

mattwynne:
@aslakhellesoy after the discussions today with @gasparnagy at the committers' Zoom, it struck me that it might be a good idea to write an [ADR](https://github.com/npryce/adr-tools) as part of this PR. I'd be happy to write a draft if you like, or we could pair/ensemble on it?

aslakhellesoy:
> @aslakhellesoy after the discussions today with @gasparnagy at the committers' Zoom, it struck me that it might be a good idea to write an [ADR](https://github.com/npryce/adr-tools) as part of this PR. I'd be happy to write a draft if you like, or we could pair/ensemble on it?

Yes, an ADR would be great @mattwynne!

aurelien-reeves:
The removal of "sudo" before installing npm globally was the thing to fix the CI? 😨

This is a big PR to review. Here's a big list of question 😅

- What should we focus on while reviewing?
- What are the next steps?  
- When could it be merged? 
- Should we wait for the PR in cucumber-js, cucumber-ruby and cucumber-jvm to be ready too? 
- Should we wait for .Net and Elixir to be implemented?

aslakhellesoy:
> The removal of "sudo" before installing npm globally was the thing to fix the CI? 😨
> 
> This is a big PR to review. Here's a big list of question 😅
> 
> * What should we focus on while reviewing?

Primarily finding breaking changes in APIs and data structures

> * What are the next steps?

Merge to master and continue working on Cucumber PRs

> * When could it be merged?

I think we can merge now, unless there are objections.

> * Should we wait for the PR in cucumber-js, cucumber-ruby and cucumber-jvm to be ready too?

I think we can work on those after this is merged.

> * Should we wait for .Net and Elixir to be implemented?

And Perl. I think we should disable them all (they already are). Waiting for them will cause too much delay. They can catch up at their own pace. We should create an issue for all 3 of them.

davidjgoss (approved):

PR #1419, opened 2021-03-12
datatable/java: all asX should behave consistently

mpkorstanje:
## Summary

The `DataTable` api had several inconsistencies that were surprising:

1. `asX` methods without a type parameter (e.g. `.asList()`) did not use the build in table transformer while those that did take a type parameter did (e.g. `.asList(String.class)`. Leading people to wonder why their data tables were not transformed https://github.com/cucumber/cucumber-jvm/issues/2262.

   By changing all `asX` methods use the converter we avoid these surprises. To retain the old behaviour:
    - Replace `DataTable.asList()` with -> `DataTable.values()`
    - Replace `DataTable.asLists()` with -> `DataTable.cells()`
    - Replace `DataTable.asMaps()` with -> `DataTable.entries()`

2. `asX` methods only accept `Type` parameters. As a result the generic types can not be inferred by the compiler leading to the confusion in [stackoverflow/66779421](https://stackoverflow.com/questions/66779421/). 
  
   By providing a `<T> X<T> asX(Class<T>)` we ensure the compiler can infer the type when a class is used.

3. Not all `asX` methods has a zero arg, class arg and type arg variant.

   Resolved by adding an `X<String> asX()`, `<T> X<T> asX(Class<T>)` and `<T> X<T> asX(Type)` parameters.

## Types of changes

## Checklist:

- [X] I have updated the documentation accordingly.
- [X] I have updated the CHANGELOG accordingly.


datatable/CHANGELOG.md
```diff
@@ -10,7 +10,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 ### Changed
-
+ * [Java] Use transformer for all `DataTable.asX` methods.
+    ([cucumber-jvm/2262](https://github.com/cucumber/cucumber-jvm/issues/2262)
+    [1419](https://github.com/cucumber/cucumber/pull/1419)
+      [mpkorstanje])
+  - To retain the old behaviour:
```
aslakhellesoy, line 17: I think there is a little indentation problem here.

aslakhellesoy (approved):
Looks good to me!

PR #1452, opened 2021-04-03
build: fix files spec in package.json for query, cucumber-expressions

davidjgoss:
Before:

image

After:

image

Seems likely that npm7 has different handling of `files` (specifically of `"*"` in relation to what's gitignored), though I can't find documentation to prove that.

The lone `index.js` making its way in was perhaps because of the reference to it in the `main` field.

### DIFF
--- a/cucumber-expressions/javascript/package.json
+++ b/cucumber-expressions/javascript/package.json
@@ -33,10 +33,6 @@
     "ts-node": "^9.1.1",
     "typescript": "^4.2.3"
   },
-  "files": [
-    "src/",
-    "dist/src/"
-  ],
   "dependencies": {
     "becke-ch--regex--s0-0-v1--base--pl--lib": "^1.4.0"
   },
--- a/cucumber-expressions/javascript/package.json
+++ b/cucumber-expressions/javascript/package.json
@@ -34,7 +34,8 @@
     "typescript": "^4.2.3"
   },
   "files": [
-    "*"
+    "src/",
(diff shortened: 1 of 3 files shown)


aslakhellesoy:
That's odd - good find!

I wonder if we might have to re-release some other modules released earlier this week too then. I'll have a look.

aslakhellesoy (commented):
Let's fix any other packages too in the same PR

davidjgoss:
@aslakhellesoy looking now as well. There are no issues at least with any of the other packages that `cucumber-js` imports (if I manually patch `query`, all our tests pass).

aslakhellesoy:
It looks like the only modules that declare `files` are `@cucumber/query` and `@cucumber/cucumber-expressions`. The others don't declare it at all (and seem to include the right files).

So perhaps a better fix is to remove the `files` declaration?

davidjgoss:
I think it's best to only include files in the package that consuming projects use/need - by not doing anything we include lots of stuff like tests, test data, github config, makefiles etc that bloats the package.

We could maybe address that more broadly in another PR though? Happy to change to just omit `files` for consistency today - let me know.

aslakhellesoy:
> I think it's best to only include files in the package that consuming projects use/need - by not doing anything we include lots of stuff like tests, test data, github config, makefiles etc that bloats the package and gets unnecessarily indexed by IDEs.

Yes, completely agree. Let's handle that in a separate PR.

PR #1454, opened 2021-04-03
cck: make javascript package publishable

davidjgoss:
Not sure if this was intentional - looks to have been added as part of the big workspaces PR.

`@cucumber/compatibility-kit@4.0.0` is not on npm and I suspect this may be why - `npm publish` would have aborted with this in.


aslakhellesoy (approved):
Good find!

PR #1480, opened 2021-04-15
[python] gherkin: run gherkin-python as a module

### LINKED ISSUE #1475: python gherkin parser should not put gherkin.py in 'bin' dir
Please see this https://github.com/pytest-dev/pytest/issues/8549

aurelien-reeves:
## Summary

Fixes #1475 

I moved the scripts away from the bin directory

## Types of changes

- Bug fix (non-breaking change which fixes an issue).

## Checklist:


brasmusson:
The problem is not having python files in the bin directory per se (`bin/gherkin` was a python file in previous releases), the problem is to have both a directory `gherkin` with an file `__init__.py` and a file `gherkin.py` (outside the `gherkin` directory)
`import gherkin`
in python can either mean, import from the directory named `gherkin` with a file `__init__.py`, or import from the file `gherkin.py`. If the paths are setup so that the `import gherkin` finds the file `gherkin.py` first, things will go wrong since then the imports in `gherkin.py` for things located in the directory `gherkin` will not be found (like `gherkin.stream.gherkin_events`) since `gherkin` already have been set up to mean `gherkin.py`.

It is only the files that are part of the release that can cause these kinds of problem, therefore nothing have to be done with `bin/gherkin_generate_tokens.py` (`setup.py` defines which files that are part of the release).

When the released `gherkin-python` is installed the file `bin/gherkin` typically ends up in `~/.local/bin`, whereas the files in the `gherkin` directory typically ends up in `~/.local/lib/<python version>/site-packages/gherkin/`, so the files in `bin` (that are released) cannot make assumptions on where the `gherkin` directory are.

To both have `bin/gherkin` be able to different python version using different executable names (in out docker image, the python-2 executable is `usr/bin/python2` and the python-3 executable is `usr/bin/python3`) and get rid of `bin/gherkin.py`, I think the most sensible thing is to move `bin/gherkin.py` to `gherkin/__main__.py` and let `bin/gherkin` execute `gherkin-python` as a module (that is using the `-m` option to the python executable).

aurelien-reeves:
Thank you for your analysis @brasmusson 
That makes much more sense.

PR #1497, opened 2021-04-26
[Gherkin] Update README snippets

aurelien-reeves:
## Summary

Fixes https://github.com/cucumber/cucumber/issues/1496

Javascript snippet from the gherkin readme is not enough to understand how to use the gherkin library.
Maybe snippets for other languages requires updates too.

## Types of changes

- Documentation


aslakhellesoy (approved):
LGTM!

It would be nice to provide default values so there isn't so much to set up for users of the library, but that can go in a separate issue.

aurelien-reeves:
Some context may still be missing, like some explanation regarding the IdGenerator for example.
Any idea how we could improve documentation about the usage of the gherkin library more than just a basic snippet in the readme? The question may actually concern all sub-projects.

aurelien-reeves:
> 
> 
> LGTM!
> 
> It would be nice to provide default values so there isn't so much to set up for users of the library, but that can go in a separate issue.

I will dig deeper indeed.
I will also check snippets for other languages.

aurelien-reeves:
Just a follow-up: I want to check the java snippet too before requesting for review/merging this.
I am currently in the process of setting-up a Java development environment 😅

Issue #249, opened 2017-08-18
gherkin: Compiling empty scenarios


mattwynne:
## Summary

Currently, the Gherkin pickles compiler removes Gherkin scenarios that don't have any steps, so they don't appear as test cases.

I understand why this was done - what's the point in running a test case that isn't going to test anything? - yet I think we're missing a thing here.

When I watch teams do BDD, they very often start out by creating just the headline or title of a scenario, and defer filling out the Given/When/Then steps until later. I'd like it if Cucumber were reporting that scenario's status as `undefined`. In order to do this, we'd need to pass a pickle out of the compiler.

## Expected Behavior

Pickles compiler outputs pickles (with no steps) for scenarios with no steps.

## Current Behavior

Pickles compiler does not output a pickle for scenarios with no steps.

## Possible Solution

I'd love some pointers here. I've never worked on the pickles compiler, but if someone can show me where to start (and we're agreed on the idea), I'd have a go at this.

## Context & Motivation

I'm thinking about how people like product owners engage with Cucumber, and that I'd guess they think of a scenario as having taken life as soon as they name it. It seems wrong to hide those scenarios from Cucumber.

aslakhellesoy:
Consider there two feature files:

```gherkin
Feature:
  Background:
    Given foo

  Scenario: empty
```

```gherkin
Feature:
  Scenario: not empty
    Given foo
```

The first one should be `undefined` and the second should not. But if we send all pickles through, how would we distinguish between the two?

aslakhellesoy:
Hi @mattwynne

The compiler is pretty simple, [compiler.rb](https://github.com/cucumber/cucumber/blob/master/gherkin/ruby/lib/gherkin/pickles/compiler.rb) is less than 200 LOC.

It doesn't have unit tests, though - only [approval tests](https://github.com/cucumber/cucumber/tree/master/gherkin/testdata/good) (the `.pickles.ndjson` files).

I would start with these files:

* Input: https://github.com/cucumber/cucumber/blob/master/gherkin/testdata/good/incomplete_scenario.feature
* Expected output: https://github.com/cucumber/cucumber/blob/master/gherkin/testdata/good/incomplete_scenario.feature (currently empty - that't what we want to change)

Rather than changing the expected output by hand, I'd make the change you think is correct. Start with the Ruby implementation - you can do the other languages later. The fix is maybe to remove these lines:

[...]

PR #1498, opened 2021-04-27
Add a feature with an empty scenario in the compatibility kit

aurelien-reeves:
## Summary

Following #249, the gherkin parser is considering empty scenarios, but cucumber implementation are not yet align on how to process empty scenarios.

## Details

An empty-scenario has been added to the compatibility kit.

cucumber-ruby is already managing empty scenarios well so fake-cucumber has been updated in order to have the same behavior as cucumber-ruby.

That way, everything remains green without a lot of effort.

No behavior has been changed in any other module. Gherkin remain unchanged.

## Checklist:

- [x] Add a dedicated feature in the compatibility kit
- [x] Implement the desired behavior in fake-cucumber
- [x] Add the corresponding ndjson to the compatibility kit
- [x] Port to the ruby implementation of the cck
- [x] Fix the acceptance tests of the json-formatter (cf. [comment here](https://github.com/cucumber/cucumber/pull/1498/#issuecomment-829464308))
- [x] Fix cucumber-ruby to make the CI green? (cf. [comment here](https://github.com/cucumber/cucumber/pull/1498#issuecomment-829972682))


aurelien-reeves:
The current issue with the CI is:
- the `json-formatter` uses `cucumber-ruby 3` to generate a reference json
- `cucumber-ruby 3` did not managed empty scenarios properly. `cucumber-ruby 4` would
- `cucumber-ruby 4` resulting json has some data in the `match.location` attribute that cannot be used for reference

In order to go further with that PR, I suggest:
- we update `cucumber-ruby` in `json-formatter/ruby-testdata` to at least version 4
- we update the sanitizer scripts to make the resulting json looks like we want to

Any other idea?

Note: the extra data in `match.location` are those multiple line locations which comes from example tables:
```json
"match": {
  "location": "../../compatibility-kit/javascript/features/examples-tables/examples-tables.feature:27:9"
},
```

aurelien-reeves:
The json report from cucumber-ruby seems actually wrong. Instead of resulting in an empty scenario, it results in a background

```json
[
  {
    "description": "  Scenario can be empty.\n  Background and hooks are not executed.\n  The state of the scenario is `undefined`",
    "elements": [
      {
        "description": "",
        "keyword": "Background",
        "line": 6,
        "name": "",
        "type": "background"
      }
    ],
    "id": "minimal",
    "keyword": "Feature",
    "line": 1,
    "name": "minimal",
    "uri": "features/empty-scenario/empty-scenario.feature"
  }
]
```

The result of the json-formatter when converting messages from fake-cucumber looks pretty good to me:
```json
[
  {
    "description": "  Scenario can be empty.\n  Background and hooks are not executed.\n  The state of the scenario is `undefined`",
    "elements": [
      {
        "id": "minimal;empty",
        "keyword": "Scenario,
        "line": 9,
        "name": "empty",
        "steps": [],
        "type": "scenario"
      }
    ],
    "id": "minimal",
    "keyword": "Feature",
    "line": 1,
    "name": "minimal",
    "uri": "features/empty-scenario/empty-scenario.feature"
  }
]
```

aurelien-reeves:
Back to draft as there is no consensus regarding empty scenarios

aurelien-reeves:
Due to lack of consensus, I close this PR.

We actually noticed that such behavior could differ depending the ecosystem it runs on, so I guess each implementation could deal with empty scenarios on their own.

As far as I can tell, cucumber-ruby and cucumber-jvm are already managing empty scenarios. However cucumber-js may still have unhandled exceptions with those.

PR #1511, opened 2021-05-03
Cleanup gherkin dotnet

gasparnagy:
Now this is done theoretically, but to make the build pass, we need to wait for a new version of the cucumber build docker image that is upgraded to use .NET 5.0. 

The changes I did:

* updated .NET Gherkin to use .NET 5.0 SDK for compilation (the target frameworks did not change)
* fixed the .NET Gherkin tests and the parser to match the other gherkin implementation
* included .NET Gherkin to the build
* restructured the .NET Gherkin code to be better compatible with the messages structure (it does not use the real messages yet, but uses classes that are compatible to that)
* fixed the .NET messages compilation for .NET 5 SDK, but did not make any other improvements there


SabotageAndi (approved):

aslakhellesoy:
Great to see progress on this @gasparnagy!

Do you plan to migrate away from protobuf and use code generation from the JSON Schemas in this PR? If not, do you plan to do this in a different PR?

gasparnagy:
@aslakhellesoy that will be another PR. in this one i did not touch messages. .NET Gherkin did not have a reference to the messages so far, so I could do this fixes/cleanup without fixing messages first. But the structures that the improved Gherkin uses are essentially the same as we will have in messages generated from the schema (I manually copied the structure of the java generated classes for now).

aslakhellesoy:
This looks good to me! Can we merge it?

PR #1535, opened 2021-05-11
Only run npm install for javascript modules

aslakhellesoy:
This avoids running npm install when we specify e.g. `LANGUAGES=java make`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -88,7 +88,7 @@ jobs:
     steps:
       - checkout
       - run:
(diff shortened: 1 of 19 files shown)


mattwynne:
Nice one, thanks!

Should we `rm -rf ../../node_modules` in the `clean` target too?

aslakhellesoy:
We could, but I think I prefer to add a `clobber` target for that. It's nice to be able to `clean` just built files without deleting `node_modules`. Will push this change.

PR #1555, opened 2021-05-18
Format GherkinDocuments

aslakhellesoy:
## Summary

Format a `messages.GherkinDocument` as either Gherkin or [MDG](https://github.com/cucumber/common/blob/master/gherkin/MARKDOWN_WITH_GHERKIN.md).

## Motivation and Context

As we're rolling out support for MDG in Cucumber, some users might want to convert all their existing feature files to MDG.
Doing this manually is tedious, so we should provide a migration tool to do this.

The intention is to have a tool that people can run like so:

```shell
# Format .feature files in-place
npx @cucumber/gherkin-utils format "features/**/*.feature"

# Translate all .feature files to .feature.md files and remove the .feature files
npx @cucumber/gherkin-utils format --move "features/**/*.feature" "features/**/*.feature.md"

# Translate all .feature files to .feature.md files and keep the .feature files
npx @cucumber/gherkin-utils format "features/**/*.feature" "features/**/*.feature.md"

# Translate all .feature.md files to .feature files and remove the .feature.md files
npx @cucumber/gherkin-utils format "features/**/*.feature.md" "features/**/*.feature"
```

## How Has This Been Tested?

The `pretty` unit test performs the following transformation:

```
Gherkin Source --parse--> GherkinDocument --pretty--> Markdown Source --parse--> Gherkin Document --pretty--> Gherkin Source
```

We perform a string comparison of the initial and final Gherkin Source.

## Types of changes

## Checklist:


gherkin-utils/javascript/src/pretty.ts
```diff
@@ -1,29 +1,31 @@
 import * as messages from '@cucumber/messages'
 
-export default function pretty(gherkinDocument: messages.GherkinDocument): string {
+type Syntax = 'markdown' | 'gherkin'
+
+export default function pretty(gherkinDocument: messages.GherkinDocument, syntax: Syntax): string {
```
aurelien-reeves, line 5: Maybe we could add a default value 'gherkin' to `syntax` to avoid breaking changes to the interface?

gherkin-utils/javascript/src/pretty.ts
```diff
@@ -1,29 +1,31 @@
...
   const feature = gherkinDocument.feature
-  let s = prettyTags(feature.tags)
+  let s = prettyTags(feature.tags, 0, syntax)
```
aurelien-reeves, line 7: Maybe we could rename that `s` variable in the process?

aurelien-reeves (commented):
I would have created two distinct renderers and call the good one from the pretty method depending the desired syntax.

It may look like duplication as the resulting code would look really close, but the intent behind it is actually distinct.

To illustrate that, I think of another format totally different from those two: another pretifier for it would look totally different, so having distinctive renderers would make more sense.

aslakhellesoy:
> I would have created two distinct renderers and call the good one from the pretty method depending the desired syntax.

I did consider that, but went this solution to avoid too much code duplication. The traversal code is identical, and the only difference is prefixing/indentation of keywords. I think that can be improved by having distinct helper functions for each rendering function rather than using `if` statements.

I'll have a go at that.

gherkin-utils/javascript/src/pretty.ts
aslakhellesoy: Agreed

gherkin-utils/javascript/src/pretty.ts
aslakhellesoy: It's an internal library, so I'm not worried about breaking the interface. I always avoid default argument values unless it's for keeping backwards compatibility (in public interfaces).

aurelien-reeves: The library is published on npm publicly, and the pretty method is publicly exposed
Even if it is not documented, some developer may use it so it remains a breaking change of the public API IMO

aslakhellesoy:
Please remember to label issues and pull requests :-)

gasparnagy:
> I did consider that, but went this solution to avoid too much code duplication. The traversal code is identical, and the only difference is prefixing/indentation of keywords. I think that can be improved by having distinct helper functions for each rendering function rather than using `if` statements.

This reminds me of all the visitor patterns I did already for Gherkin AST... I wanted to make a generic one into Gherkin, but never got the time for that. I wonder how much of the duplication we could avoid with a generic gherkin AST visitor...

laeubi:
@aslakhellesoy would be usefull to have this for Java too (e.g. to format existing feature files), in general it would be nice to support a configuration file where one could configure some aspects e.g. indentation deep, indentation style (tab versus spaces), numer of new lines before/after a given scenario/data table/ ...

aslakhellesoy:
@laeubi this will be published as a command line tool that you can run as `npx @cucumber/gherkin-utils`. This requires Node.js to be installed.

I realise people who don't use Node.js might not have it installed, but it's relatively easy to install. I'm not going to prioritise porting this to Java, but if someone submits a PR for this we'd accept it. The PR would also have to extract the test data to reusable tests so we can ensure consistent behaviour between implementations. A bit like we've done in Gherkin and create-meta.

aslakhellesoy:
@davidjgoss @aurelien-reeves @mattwynne could one of you review this please? Thanks!

aurelien-reeves (commented):
I'll try to find some time to have proper tests
At the moment, here are a few comments after a first quick look at the code :)

gherkin-utils/javascript/src/pretty.ts
```diff
@@ -1,78 +1,228 @@
...
+export type Syntax = 'markdown' | 'gherkin'
+
+export default function pretty(
+  gherkinDocument: messages.GherkinDocument,
+  syntax: Syntax = 'gherkin'
+): string {
   const feature = gherkinDocument.feature
-  let s = prettyTags(feature.tags)
+  if (!feature) return ''
+  let s = ''
```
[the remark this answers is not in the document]
aslakhellesoy: What would you suggest renaming it to @aurelien-reeves?

gherkin-utils/javascript/src/pretty.ts
aslakhellesoy: Done - defaults to 'gherkin'

gherkin-utils/javascript/src/pretty.ts
```diff
@@ -1,78 +1,228 @@
...
+  if (syntax === 'markdown') {
+    return new Array(level + 2).join('#') + ' '
+  } else {
+    return new Array(level + 1).join('  ')
+  }
+}
+
+function indentSpace(level: number): string {
+  return new Array(level + 1).join('  ')
+}
```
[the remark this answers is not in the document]
aslakhellesoy: I've renamed `indent` to `keywordPrefix` and `indentSpace` to `spaces`

gherkin-utils/javascript/src/pretty.ts
```diff
@@ -1,78 +1,228 @@
...
+
+function stepIndent(level: number, syntax: Syntax): string {
```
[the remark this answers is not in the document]
aslakhellesoy: Renamed to `stepPrefix` to be consistent with `keywordPrefix`

gherkin-utils/javascript/src/pretty.ts
aurelien-reeves: `prettifiedFeature`?

mpkorstanje:
The current CLI interface isn't very shell friendly. Consider making it easy to do:

Output formatted std in to std out:  `cat test.feature | gherkin-utils format`

Output formatted feature files to std out:  `gherkin-utils format [ FILE1, FILE2, FILE3.... ]`
 
Format feature files in place:  `gherkin-utils  format --in-place [ FILE1, FILE2, FILE3.... ]`

Format selected feature files in place: `find <some thing> . -exec gherkin-utils format {} \;`

Convert to markdown in place:  `convert --in-place [ FILE1, FILE2, FILE3.... ]`

mattwynne:
I wonder if we could just call this CLI tool `gherkin` instead of `gherkin-utils`. We could still build/release it as a separate library (`gherkin-cli` would follow a common JS idiom for the package name).

gherkin-utils/javascript/src/pretty.ts
```diff
@@ -1,78 +1,227 @@
 import * as messages from '@cucumber/messages'
 
-export default function pretty(gherkinDocument: messages.GherkinDocument): string {
+export type Syntax = 'markdown' | 'gherkin'
+
+export default function pretty(
```
mattwynne, line 5: The pretty formatter for TS/JS was [recently given an official re-write](https://github.com/jbpros/cucumber-pretty-formatter) - could we not re-use that code here?

mattwynne: Or am I reading it wrong and this is for pretty-printing markdown?

aslakhellesoy: This library converts a [GherkinDocument](https://github.com/cucumber/common/blob/main/messages/messages.md#gherkindocument) to a Gherkin or Markdown string, depending on what `Syntax` is passed as the 2nd argument.

The other pretty formatter has a different processing model - it prints lines as results are coming in. This library just prints the whole thing in one go.

I'd rather evolve this codebase to be able to print with results as well, but I'd rather handle that in a separate pull request.

PR #1564, opened 2021-05-23
Require Markdown tables to be indented 2-5 spaces to be recognised

aslakhellesoy:
## Summary

Data Tables and Examples Tables will only be parsed if thet are indented 2-5 spaces. Otherwise they will be ignored.

## Details

See discussion here: https://github.com/cucumber/common/pull/1391#discussion_r637381820

This change also contains some changes to the CCK's `markdown.md` file so that we have a step with all of the following:

* Data Table
* Attachment
* Exception

This should make it easier to fix the double rendering issues mentioned in #1391

## Motivation and Context

As suggested by @davidjgoss in https://github.com/cucumber/common/pull/1391#discussion_r637381820, indentation is needed in order for Markdown parsers (such as `react-markdown`) to consider content as part of a list item.

Requiring indentation for Examples Tables solves another problem - distinguising between "content" tables and "Gherkin" tables. Without this distinction, a GFM table at an "unexpected" location (such as at the beginning of a document) would result in a parse error.

Introducing this distinction would cause unindented tables to be ignored by the Gherkin parser.

The reason for the 2-5 space indentation rule is that this is the range for which markdown parsers consider a table in a bullet to still be a table.

Tables not in a bullet will still be recognised as a table for 2-3 indents (but not 4 or more), so 2-5 seemed a reasonable range. We may want to be more restrictive, say exactly 2 space indent, or 2-3 spaces - I'm not sure...

## How Has This Been Tested?

There are new unit tests as well as acceptance tests in Gherkin.

## Types of changes

## Checklist:


davidjgoss (approved):
LGTM

PR #1575, opened 2021-05-26
JSON Schema consistency

mattwynne:
## Summary

Playing with https://quicktype.io/ we discovered a couple of inconsistencies in the JSON schema docs

## Details

* ParameterType required fields listed `id` twice
* ParameterType required fields listen `children` but it was not defined
* TestCaseFinished required fields listed `testSteps` but it's not needed on that message

## Motivation and Context

Cleaning up the JSON schema so other tools could understand and use it.

## How Has This Been Tested?

Using `npx quickcheck messages/jsonschema/*.json tmp.rb`

## Types of changes

## Schema validation

In order to have continuous validation of the schema consistency, `ajv-cli` as been added and a validation target has been added to messages Makefile. That way json-schema consistency can be checked easily and is now part of the CI.


messages/jsonschema/ParameterType.json
```diff
@@ -29,6 +28,12 @@
     },
     "id": {
       "type": "string"
+    },
+    "children": {
+      "items": {
+        "$ref": "#"
```
aslakhellesoy, line 34: This seems to have broken the generated TypeScript code (see CI failure). I assume ˋ#ˋ means self-reference? Where in the json schema spec is that documented?

aslakhellesoy (changes requested):

aslakhellesoy:
Since generated code is added to git, you have to run ˋmakeˋ from the ˋmessagesˋ directory and commit the modified files (if any).

messages/jsonschema/ParameterType.json
aslakhellesoy: I haven't found any mentions of `#` meaning self-reference. [This](https://stackoverflow.com/questions/35250621/recursive-self-referencing-json-schema) SO suggests the reference needs to go via `definitions`.

And [this](https://stackoverflow.com/questions/20752716/json-schema-recursive-schema-definition/20753545#20753545) post suggests that `#` refers to the root of the document (not "self")

mattwynne: Well, what _is_ the correct type for an item in the `children` array? We assumed it was instances of `ParameterType` so that `#` is meant to refer to the root of the document. Maybe `./ParameterType.json` would be better?

mattwynne: There's some info about JSON pointers [here](https://json-schema.org/understanding-json-schema/structuring.html#id1) and a [spec](https://datatracker.ietf.org/doc/html/rfc6901) which talks about `#` referring to "the whole document".

aslakhellesoy: > Well, what is the correct type for an item in the children array?

Sorry, my bad - I saw `children` and incorrectly assumed this: https://github.com/cucumber/common/blob/1b64380f36cf6bb8534b59af7b064e84299c7b22/messages/messages.proto#L559-L563

Which in JSON Schema is this:

https://github.com/cucumber/common/blob/1b64380f36cf6bb8534b59af7b064e84299c7b22/messages/jsonschema/TestCase.json#L5-L28

The `ParameterType` message doesn't have a `children` property at all:

https://github.com/cucumber/common/blob/1b64380f36cf6bb8534b59af7b064e84299c7b22/messages/messages.proto#L681-L688

A bit of background: I used a `.proto` to `JSON Schema` conversion tool to generate the JSON schema files. The tool didn't convert everything correctly, so I've done many manual edits to the generated schemas. In the process I've apparently made several mistakes (copy paste errors) in the `required` sections. There are probably more.

These mistakes have gone unnoticed until now, because the custom Ruby code generator doesn't generate code based on the `required` array, but rather the `properties` dictionary of the schema:

* https://github.com/cucumber/common/blob/1b64380f36cf6bb8534b59af7b064e84299c7b22/messages/jsonschema/scripts/codegen.rb#L115
* https://github.com/cucumber/common/blob/1b64380f36cf6bb8534b59af7b064e84299c7b22/messages/jsonschema/scripts/codegen.rb#L164

This means mistakes/inconsistencies in `required` are not detected. We should probably change the code generators to iterate over `required` instead. That way we should be able to find the remaining mistakes.

Sorry again for leading you down the wrong path!

aslakhellesoy: While you're working on the schemas - could you also please replace all the `"$schema": "http://json-schema.org/draft-04/schema#"` with `"$schema": "https://json-schema.org/draft/2020-12/schema`? (The conversion tool I used put those old versions in).

FWIW, there isn't currently a non-draft version of the JSON schema spec - see [available versions](https://json-schema.org/specification-links.html)).

messages/Makefile
```diff
@@ -4,8 +4,36 @@ include default.mk
...
 # docker from docker doesn't work).
-default: messages.md
+default: .validated messages.md
+else
+default: .validated
 endif
 
+.validated:
+	cd jsonschema && npx ajv compile --spec=draft2020 --strict=true \
+		-s "Source.json" \
```
aslakhellesoy, line 14: We should generate this list automatically. Fix coming....

.circleci/config.yml
```diff
@@ -96,6 +96,13 @@ jobs:
             npm install --global npm@7
             npm ci
             npm run build
+      - run:
```
aslakhellesoy, line 99: I think this should be done in a separate job - treating the jsonschemas directory as a separate module

compatibility-kit/javascript/Makefile
```diff
@@ -2,13 +2,13 @@ include default.mk
...
+VALIDATED_NDJSON_FILES = $(patsubst features/%.ndjson,acceptance/%.ndjson,$(NDJSON_FILES))
 
 .DELETE_ON_ERROR:
 
-.tested: ndjson_files
-
-ndjson_files: $(NDJSON_FILES)
-.PHONY: ndjson_files
+.tested: $(NDJSON_FILES)
+# TODO: Uncomment next line and remove previous line
```
aslakhellesoy, line 10: @aurelien-reeves @mattwynne I have added a validation utility that validates all of the CCK messages against the schema. It's currently disabled, but if you comment out the next line you can see it fail.

To validate a specific file, run:

    make acceptance/minimal/minimal.feature.ndjson.valid

I think we should make further changes to the schemas to make these validation checks pass. Can you two have a look at that?

aslakhellesoy (changes requested):

.circleci/config.yml
aslakhellesoy: I've fixed this

messages/Makefile
aslakhellesoy: I realised the order is significant, so a fully automated list using `find` won't work until ajv fixes https://github.com/ajv-validator/ajv-cli/issues/43

compatibility-kit/javascript/Makefile
aslakhellesoy: It was an easy enough fix - done!

aslakhellesoy:
I've also added a new Markdown code generator, so our `messages/messages.md` is updated automatically.

PR #1593, opened 2021-06-03
gherkin/ruby: Fix tagging of rules

luke-hill:
## Summary
Fix tagging of Rules

## Details

During the Gherkin Query cache step where it finds all data and relates them to specific AST Node IDs, it wasn't storing tag data. This was likely because similar to features, it needs to look "up" and not down through descendents

## Motivation and Context

Fix Ruby

## How Has This Been Tested?

CI
Altered tests in cucumber-ruby
Discussion during general meeting about executing portions of the regular CCK against all flavours of cucumber

## Types of changes

## Checklist:

### DIFF
--- a/gherkin/ruby/lib/gherkin/query.rb
+++ b/gherkin/ruby/lib/gherkin/query.rb
@@ -27,6 +27,9 @@ module Gherkin
     end
 
     def update_rule(rule)
+      return if rule.nil?
+      store_nodes_location(rule[:tags])
+
       rule[:children].each do |child|
         update_background(child[:background]) if child[:background]
         update_scenario(child[:scenario]) if child[:scenario]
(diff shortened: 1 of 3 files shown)


gherkin/ruby/lib/gherkin/query.rb
```diff
@@ -27,6 +27,9 @@ def update_feature(feature)
     end
 
     def update_rule(rule)
+      return if rule.nil?
```
luke-hill, line 30: Copied this from the method above. I'm not sure this can happen (Same for features), but I'm not really sure how the parser generates the original AST which is then iterated through when we click "Run scenario" so I'll just copy and let it all work.

aurelien-reeves: It looks safer that way.
Good to me, even if useless. It may be useless for now, but that can change.

aurelien-reeves (commented):
Would it be possible to add at east one test in query_spec which would illustrate the fix?

aurelien-reeves (approved):

Issue #1559, opened 2021-05-20
gherkin (Ruby): Undocumented API changes


enkessler:
## Summary

I'm in the process of updating one of my projects to be able to use the 19.x major version of the `cucumber-gherkin` gem and am running into some unexpected API changes that do not appear to be reflected in the [CHANGELOG](https://github.com/cucumber/common/blob/master/gherkin/CHANGELOG.md) at all.

## Expected Behavior

Because the CHANGELOG only mentions some dialect updates (which would not impact my project) and dependency upgrades (whose own changelogs, in turn, had no impacting call outs), I was expecting things to still 'just work' after I upgraded from 18.x to 19.x.

## Current Behavior

I'm seeing several places in the code where data keys in returned AST data got converted from `snake_case` to `camelCase` which, naturally, breaks any code that uses that particular portion of the API. So far, I've run into

`parse_error` -> `parseError`
`media_type` -> `mediaType`
`doc_string` -> `docString`
`data_table` -> `dataTable`

but I'm still in the process of upgrading so there may be several more that I'll run into.

## Possible Solution

It depends on if all of the name changes were intentional or not. It's not the first time that the names of things have changed but I don't know if it was an attempt at standardization across language implementations or it's just that whoever wrote the code was in the habit of writing in a different style.

If the changes were intentional
 - Update the CHANGELOG to reflect the changes

If the changes were not intentional
 - Revert them and probably add some more testing around what the AST structure is so that these kinds of bugs don't slip through in the future

## Context & Motivation

While I understand that an increase in the major version number indicates backwards incompatibility in some fashion, those incompatibilities tend to get called out. Since they weren't mentioned, I figured that I'd check and see if the changes were just as much of a surprise to everyone else as they were to me.

## Your Environment

* Version used: `cucumber-gherkin (19.0.2)`
* Operating System and version: N/A
* Link to your project: https://github.com/enkessler/cuke_modeler

aurelien-reeves:
Hello @enkessler 

Thanks for the report.

This is actually related to messages@16.0. It does not use protocol buffers anymore, but jsonschema to describe the messages.

The impact on AST is more a side effect than an expected change.

We need to improve the documentation, and the changelog, accordingly.

Regarding Ruby, the big difference is that messages are no longer classes, but hashes, and the key are now camelCase as you described.

aurelien-reeves:
CHANGELOGs have been updated
Thanks again for your report

enkessler:
[...]

PR #1597, opened 2021-06-08
Messages: codegen for ruby DTOs

### LINKED ISSUE #1574: Generate Ruby DTOs for messages
## Summary

We're currently using a plain old Ruby Hash to represent message payloads. It will be simpler for consumers if we can de-serialize the messages into simple Ruby value objects, (or DTO - Data Transfer Objects) and pass those around instead.

## Expected Behavior

API consumers should receive messages as Ruby objects instead of hashes.

## Possible Solution

See Aslak's advice [here](https://github.com/cucumber/cucumber-ruby/pull/1544#issuecomment-847734909).

aurelien-reeves:
## Summary

Add a code generator for messages to generate ruby DTOs instead of relying on plain ruby hashes

Refs. #1574 #1559

## Details

Add templates and class to Codegen for ruby DTOs

## Motivation and Context

Cf. #1574

## How Has This Been Tested?

N/A

## Types of changes

- New feature (non-breaking change which adds functionality).


aurelien-reeves:
Before going further - like using those DTOs in cucumber-gherkin - I suggest to review and merge that PR as-is while it is still relatively small and easy to review.

Then we could open a new PR dedicated to using those DTOs into Gherkin.

WDYT?

mattwynne:
I think this is for a separate PR, but the idea occurs to me that we could include some small amount of RDoc in the generated code, perhaps just referencing some other place where we document the message protocol.

aslakhellesoy:
Might as well put that in now I think. The URL to the documentation is https://github.com/cucumber/common/tree/main/messages#readme

mattwynne (approved):

aurelien-reeves:
Great idea!

I have even added the documentation that is already in the json schemas into the generated DTOs.

We may do that for typescript and go.

Some formatting is part of the schemas. We may also clean those a little bit.

But that can be done in other PR, I will merge this one soon to move on the ruby DTOs: using those within gherkin.

PR #1601, opened 2021-06-09
Messages: ruby: Convert DTOs attributes from camelCase to snake_case

aurelien-reeves:
The point of the ruby DTOs was to stick with ruby standard.
We missed to convert cameCase attribute names to snake_case in the previous PR: #1574

### DIFF
--- a/messages/jsonschema/scripts/codegen.rb
+++ b/messages/jsonschema/scripts/codegen.rb
@@ -183,6 +183,17 @@ class Ruby < Codegen
       .map { |description_line| "# #{description_line}" }
       .join("\n#{indent_string}")
   end
+
+  # Thank you very munch rails!
+  # https://github.com/rails/rails/blob/v6.1.3.2/activesupport/lib/active_support/inflector/methods.rb#L92
+  def underscore(camel_cased_word)
+    return camel_cased_word unless /[A-Z-]/.match?(camel_cased_word)
+    word = camel_cased_word.gsub(/([A-Z\d]+)([A-Z][a-z])/, '\1_\2')
+    word.gsub!(/([a-z\d])([A-Z])/, '\1_\2')
+    word.tr!("-", "_")
+    word.downcase!
+    word
+  end
 end
 
 class Go < Codegen
--- a/messages/CHANGELOG.md
+++ b/messages/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+* [Ruby] Generate DTOs from the json schemas to avoid usage of plain ruby hashes
+  ([#1574](https://github.com/cucumber/common/issues/1574))
+
 ### Changed
 
 ### Deprecated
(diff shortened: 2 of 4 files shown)


messages/jsonschema/scripts/codegen.rb
```diff
@@ -183,6 +183,17 @@ def format_description(raw_description, indent_string: "    ")
       .map { |description_line| "# #{description_line}" }
       .join("\n#{indent_string}")
   end
+
+  # Thank you very munch rails!
+  # https://github.com/rails/rails/blob/v6.1.3.2/activesupport/lib/active_support/inflector/methods.rb#L92
+  def underscore(camel_cased_word)
```
aslakhellesoy, line 189: It seems to me that converting to `snake_case` complicates everything a lot. If we weren't doing this, we could generate JSON simply by doing `JSON.stringify`.

I know `snake_case` is a convention in Ruby, but here we're dealing with an external data format, and I think the additional complexity of snake case is difficult to justify.

aurelien-reeves: There is no such thing as JSON.stringify in ruby.
JSON.generate works with hashes only, not with objects. So we will have to write serializers anyway.

snake_case over camelCase was one of the point into generating DTOs over hashes.

mattwynne: I think the message DTOs need to follow Ruby idioms. The whole point of adding the DTOs, for me, is to add a layer of indirection between the raw messages and client code, precisely so that we can do things like this.

Can you same more about the JSON.stringify use case? I didn't realise there were instances where we need to convert these objects back to raw JSON messages.

aslakhellesoy: They need to be converted to JSON / NDJSON for the message formatter, and also for publishing to reports.cucumber.io.

The only users of the Cucumber Messages Ruby API is us. Going the roundabout way of using snake case seems completely unnecessary to me.

aurelien-reeves: > 
> 
> I think the message DTOs need to follow Ruby idioms. The whole point of adding the DTOs, for me, is to add a layer of indirection between the raw messages and client code, precisely so that we can do things like this.
> 
> Can you same more about the JSON.stringify use case? I didn't realise there were instances where we need to convert these objects back to raw JSON messages.

The formatters output messages as ndjson
Also gherkin as a standalone provide a bin which allows to output ndjson from those messages

cf the CI here: https://github.com/cucumber/common/pull/1603

aslakhellesoy: > snake_case over camelCase was one of the point into generating DTOs over hashes.

That wasn't obvious to me. For me, the only benefit was that we can do `envelope.testStepFinished.timestamp` rather than having to do `envelope[:testStepFinished][:timestamp]`.

aslakhellesoy: Another benefit of DTOs is that we can throw an exception in the constructor if you pass an unsupported property.

aslakhellesoy: > There is no such thing as JSON.stringify in ruby.

My bad. We could do this:

```ruby
require 'json'

class Timestamp
  attr_reader :seconds
  attr_reader :nanos

  def initialize(
    seconds: 0,
    nanos: 0
  )
    @seconds = seconds
    @nanos = nanos
  end

  def to_hash
    return {
      seconds: seconds,
      nanos: nanos
    }
  end
end

t = Timestamp.new(seconds: 10, nanos: 33)
puts t.to_hash.to_json
```

PR #1605, opened 2021-06-10
Messages ruby: DTOs serialization and deserialization capabilities

aurelien-reeves:
Add serialization and deserialization capabilities to ruby DTOs

Current status: ready to be reviewed


aurelien-reeves:
Deserialization of messages with arrays of other types of messages won't be easy without the addition of a complex layer which would analysis the DTOs and the schemas to find appropriate classes

aslakhellesoy:
Nah, here is a simple solution:

```ruby
require 'json'

class Envelope
  attr_reader :meta

  def initialize(
    meta: nil
  )
    @meta = meta
  end

  def self.from_hash(hash)
    self.new(
      meta: Meta.from_hash(hash[:meta])
    )
  end
end

class Meta
  attr_reader :protocol_version
  attr_reader :implementation

  def initialize(
    protocol_version: nil,
    implementation: nil
  )
    @protocol_version = protocol_version
    @implementation = implementation
  end

  def self.from_hash(hash)
    self.new(
      protocol_version: hash[:protocol_version],
      implementation: Product.from_hash(hash[:implementation])
    )
  end
end

class Product
  attr_reader :name
  attr_reader :version

  def initialize(
    name: nil,
    version: nil
  )
    @name = name
    @version = version
  end

  def self.from_hash(hash)
    self.new(
      name: hash[:name],
      version: hash[:version]
    )
  end
end

describe '.from_hash' do
  it 'reconstructs a deep object' do
    # Normally we'd construct the hash with JSON.parse(json, {symbolize_names: true})
    hash = {
      meta: {
        protocol_version: '10.2.3',
        implementation: {
          name: 'cucumis',
          version: '55.7'
        }
      }
    }

    envelope = Envelope.from_hash(hash)
    expect(envelope.class).to eq Envelope
    expect(envelope.meta.class).to eq Meta
    expect(envelope.meta.implementation.class).to eq Product
    puts envelope.inspect
  end
end
```

We just need to generate a `.from_hash` class method on each class.

aslakhellesoy:
And for arrays, something like this:

```ruby
def self.from_hash(hash)
  self.new(
    object_array: (hash[:object_array] || []).map {|element| SomeClass.from_hash(element)},
    primitive_array: (hash[:primitive_array] || [])
  )
end
```

mattwynne:
I like the idea of adding (manually-written) specs for this behaviour in the DTO, even if the implementation of the solution is generated code.

aslakhellesoy:
Yeah we need specs for the deserialisation. I didn't modify the code generator to add those `.from_hash` class methods, the code I pasted above if just a proof of concept of what we *could* generate.

aurelien-reeves:
Thanks @aslakhellesoy 
I'll study that today 👍

aurelien-reeves:
Here's a first draft of a generic solution.
The idea remain to generate as less code as possible when generating DTOs.

This is still draft. Some refactoring is needed.

aurelien-reeves:
I've refactored the `Cucumber::Messages::Message` base class. It result in a bunch of small mixins.

I kept generated code as small as possible to avoid overloading the resulting DTOs.

I've added some rdoc to what seems to be a public API.

Unit tests could be added to the new mixins. But the new message_spec.rb covers serialization and deserialization from a higher level.

messages/ruby/lib/cucumber/messages/message.rb
```diff
@@ -0,0 +1,12 @@
...
+
+module Cucumber
+  module Messages
+    class Message
+      include Cucumber::Messages::Message::Deserialization
+      include Cucumber::Messages::Message::Serialization
+
+    end
+  end
+end
```
luke-hill, line 12: not sure if we care about newline formatting? But flagging it here.

messages/ruby/lib/cucumber/messages.dtos.rb
```diff
@@ -18,10 +23,8 @@ module Messages
     #  It is not to be used for runtime errors raised/thrown during execution. This
     #  is captured in `TestResult`.
     #
-    # The code was auto-generated by {this script}[https://github.com/cucumber/common/blob/main/messages/jsonschema/scripts/codegen.rb]
-    #
 
-    class Attachment
+    class Attachment < ::Cucumber::Messages::Message
```
luke-hill, line 27: Stylistic question, flagging here.

Is there a reason we're not opting for the 1 class per file structure (I know this is auto-genned, but it still should be possible / desirable)

messages/ruby/lib/cucumber/messages.dtos.rb
```diff
@@ -588,6 +604,16 @@ def initialize(
         @children = children
         @id = id
       end
+
+      private
```
luke-hill, line 608: just having one private and then a line space is enough (Again aware this is auto-genned).

messages/ruby/lib/cucumber/messages/message/utils.rb
```diff
@@ -0,0 +1,45 @@
...
+          # https://github.com/rails/rails/blob/v6.1.3.2/activesupport/lib/active_support/inflector/methods.rb#L69
+
+          def camelize(term)
+            camelized = term.to_s
+            camelized.gsub!(/(?:_|(\/))([a-z\d]*)/i) { "#{$1}#{$2.capitalize}" }
+            camelized
+          end
+        end
+
+        def self.included(other)
```
luke-hill, line 39: traditionally this would be at the top as this is the first thing you read. Similar to what I have in SitePrism here: https://github.com/site-prism/site_prism/blob/main/lib/site_prism/dsl.rb#L11

messages/jsonschema/scripts/templates/ruby.rb.erb
```diff
@@ -34,6 +37,15 @@ module Cucumber
         @<%= underscore(property_name) %> = <%= underscore(property_name) %>
       <%- end -%>
       end
+      <%- schema['properties'].each do |property_name, property| -%>
+        <%- if property['items'] && property['items']['$ref']  -%>
```
luke-hill, line 41: This can be heavily optimised to use `#dig`

`property.dig('items', $ref)` will return either `nil` or the value from the digging in. It's also slightly more optimal than doing double fetch iirc

luke-hill (commented):
I couldn't review the codegen file proper. But flagged a few style suggestions.

I've not reviewed every bit as a lot looks repeated.

messages/ruby/lib/cucumber/messages.dtos.rb
aurelien-reeves: The codegen script has been built like that, it is easier to remain consistent with other generated languages like typescript, and generating a single file for messages is really easier at the moment.

messages/ruby/lib/cucumber/messages.dtos.rb
aurelien-reeves: Yes I know
Managing that in the ruby codegen erb template would add a lot of noise
I am not sure the benefit of having a proper automaticaly generated file would override that cost

aurelien-reeves:
Thanks @luke-hill for the review!

The important it the PR is the serialization/deserialization capabilities.
I've just pushed refactored spec for those mixins.

messages/ruby/lib/cucumber/messages/message/deserialization.rb
```diff
@@ -0,0 +1,142 @@
...
+          #
+          #   Cucumber::Messages::Message.from_json('{"pickleTag":{"name":"foo","astNodeId":"abc-def"}}')  # => #<Cucumber::Messages::PickleTag:0x00007efda1451a00 @name="foo", @ast_node_id="abc-def">
+          #
+          # If more than one message is part of the JSON document, it will return an array of messages
+          #
+          #   Cucumber::Messages::Message.from_json('{
+          #     "pickleTag":{"name":"foo","astNodeId":"abc-def"},
+          #     "duration":{"seconds":1,"nanos":42}
+          #   }')
+          #   => [
```
aslakhellesoy, line 34: This surprising behaviour.

I would expect:

* JSON primitive => Ruby primitive
* JSON object => Ruby object
* JSON array => Ruby array

Here however, you're turning a JSON object into a Ruby array.

messages/ruby/lib/cucumber/messages/message/deserialization.rb
```diff
@@ -0,0 +1,142 @@
...
+          def prepare_arguments(raw_arguments, dto: self)
+            raw_arguments.each do |key, value|
+              raw_arguments[key] = prepare_argument(key, value, dto: dto)
+            end
+
+            raw_arguments.transform_keys { |key| underscore(key.to_s).to_sym }
+          end
+
+          def find_dto(name)
+            Object.const_get("::Cucumber::Messages::#{name}")
```
aslakhellesoy, line 115: This won't work. You cannot assume that the property name has a corresponding message type. For example, one of the messages has a property called `implementation` (of type `Product`), but here you're trying to access a non-existent constant called `Implementation`

aslakhellesoy (changes requested):
You've gone for an implementation where the deserialisation logic has no knowledge of the schema and tried to implement smart reflection logic in the `Deserialization` module that will figure it out with reflection.

This isn't going to work.

The deserialization logic has to know about the schema - either by loading the schema at runtime, or by generating deserialization code that knows about the schema.

I wouldn't recommend the runtime approach, which is why I suggested the code generation approach in https://github.com/cucumber/common/pull/1605#issuecomment-859731059

messages/ruby/lib/cucumber/messages/message/serialization.rb
```diff
@@ -0,0 +1,66 @@
...
+        #   Cucumber::Messages::Duration.new(seconds: 1, nanos: 42).to_h                                 # => { seconds: 1, nanos: 42 }
+        #   Cucumber::Messages::PickleTag.new(name: 'foo', ast_node_id: 'abc-def').to_h(camelize: true)  # => { name: 'foo', astNodeId: 'abc-def' }
+        #
+        # It is recursive so embedded messages are also processed
+        #
+        #   location = Cucumber::Messages::Location.new(line: 2)
+        #   Cucumber::Messages::Comment.new(location: location, text: 'comment').to_h  # => { location: { line: 2, :column: nil }, text: "comment" }
+        #
+
+        def to_h(camelize: false)
```
aslakhellesoy, line 24: This implementation will return a hash with `nil` values for fields that are `nil`.

This in turn will include all fields in the JSON, which we want to avoid because it increases message size considerably.

Unfortunately `JSON.generate` does not provide any options to not output `nil` values as `null`. We therefore need a `to_hash` implementation that only includes non-nil values.

aslakhellesoy:
I've created https://github.com/cucumber/common/pull/1617 which addresses the feedback I gave in this PR.

aurelien-reeves:
@aslakhellesoy thank you for your feedback.

Indeed I tried implementing the deserialization using reflection. I initially understood very well your proposition to add serialization/deserialization logic as part of the generated code. 

I still would prefer to generate DTOs as light as possible for several reason.

Serialization is working very well with reflection, but yes, deserialization is very difficult that way, and there will always be issues with edge cases.

So, here's a new suggestion. What would you think of:

- generating lightweight DTOs with only the attributes
- using reflection for serialization as it is done in the current PR
- generating dedicated code for deserialization as part of the codegen script - as it is done in your PR - into a dedicated module outside of DTOs. Thanks to Ruby, it is super easy to reopen an existing class. To use the deserialization, a user would just have to opt-in for it requiring it explicitly.

mattwynne:
IMO, if we want to keep the deserialization "clutter" out of the generated DTO classes, we'd be better off putting that code in a module and using `extend` to add it to the metaclass. That way the DTOs are simple to read, but the user doesn't have to think about requiring extra files to get the deserialization behaviour.

aurelien-reeves:
> 
> 
> IMO, if we want to keep the deserialization "clutter" out of the generated DTO classes, we'd be better off putting that code in a module and using `extend` to add it to the metaclass. That way the DTOs are simple to read, but the user doesn't have to think about requiring extra files to get the deserialization behaviour.

Done.
Deserialization are now part of messages per default.
`require 'cucumber/messages'` now allows to use the DTOs, but also the deserializers.

aslakhellesoy:
> Serialization is working very well with reflection, but yes, deserialization is very difficult that way.

It's worse than difficult - it's impossible.

Knowledge of the schema is the only way we can infer the tight types to instantiate.

Good to see the deserialisation logic extracted to separate files! I would have preferred to do the same for the serialization logic (generate a separate file) rather than using reflection. Reflection adds performance overhead and complexity to the library, whereas with the statically generated version, there is no penalty for using reflection and regexp evaluation.

It's a tradeoff between complexity in the code generator and the runtime code, and I think I prefer to keep the complexity out of the runtime code.

messages/ruby/lib/cucumber/messages/message/serialization.rb
```diff
@@ -0,0 +1,72 @@
...
+        #   Cucumber::Messages::PickleTag.new(name: 'foo', ast_node_id: 'abc-def').to_h(camelize: true)  # => { name: 'foo', astNodeId: 'abc-def' }
+        #   Cucumber::Messages::PickleTag.new(name: 'foo', ast_node_id: nil).to_h(reject_nil_values: true)  # => { name: 'foo' }
+        #
+        # It is recursive so embedded messages are also processed
+        #
+        #   location = Cucumber::Messages::Location.new(line: 2)
+        #   Cucumber::Messages::Comment.new(location: location, text: 'comment').to_h  # => { location: { line: 2, :column: nil }, text: "comment" }
+        #
+
+        def to_h(camelize: false, reject_nil_values: false)
```
aslakhellesoy, line 26: What concrete use cases do we have for the `camelize` and `reject_nil_values` arguments?

aurelien-reeves: Those are used for json serialization.
`camelize` for the json using camelCase keys
`reject_nil_values` to avoid serializing null values as requested here: https://github.com/cucumber/common/pull/1605#discussion_r653065795

aslakhellesoy: I understand that, but why do we ever need to set `camelize` to `false` or `reject_nil_values` to `false`?

Was this flexibility added just in case someone else needs it? AFAICT we don't need it ourselves.

aurelien-reeves: > I understand that, but why do we ever need to set `camelize` to `false` or `reject_nil_values` to `false`?
> 
> Was this flexibility added just in case someone else needs it? AFAICT we don't need it ourselves.

Yes.

Because having a ruby hash with camelcase keys is not the ruby standard, and reject_nil_values to avoid data loss per default. So per default, the method has an intuitive use case.

aslakhellesoy: This speculative flexibility comes at a complexity and performance cost.

I assume you went with a reflection based serialization implementation in order to support this use case. If we implemented it without these option (always `camelize` and `reject_nil_values`), we wouldn't need the additional logic, and a static code generation would be sufficient (less complex and faster).

I think this is feature creep, but I'll rest my case here.

aurelien-reeves: > This speculative flexibility comes at a complexity and performance cost.

The whole deserialization thing, which is the big piece in the current PR, is based on speculation. At first we wanted to implement only the serialization, which is required for the DTOs to work properly, then you requested the deserialization just in case someone else would need it, because it would be a more comprehensive developer experience. We actually do not need it ourselves.

The addition of a `to_h` method to messages is at first for our own usage. Make it intuitive and public for third-parties developer is a bonus. Actually those options do not add any complexity to the code as they result in only 2 extra `if` statement.

```ruby
def to_h(camelize: false, reject_nil_values: false)
  resulting_hash = self.instance_variables.to_h do |variable_name|
    h_key = variable_name[1..-1]
    h_key = Cucumber::Messages::Message.camelize(h_key) if camelize

    h_value = prepare_value(
      self.instance_variable_get(variable_name),
      camelize: camelize,
      reject_nil_values: reject_nil_values
    )

    [ h_key.to_sym, h_value ]
  end

  resulting_hash.reject! { |_, value| value.nil? } if reject_nil_values
  resulting_hash
end

def to_h
  resulting_hash = self.instance_variables.to_h do |variable_name|
    h_key = variable_name[1..-1]
    h_key = Cucumber::Messages::Message.camelize(h_key)

    h_value = prepare_value(
      self.instance_variable_get(variable_name)
    )

    [ h_key.to_sym, h_value ]
  end

  resulting_hash.reject! { |_, value| value.nil? }
  resulting_hash
end
```

> I assume you went with a reflection based serialization implementation in order to support this use case. If we implemented it without these option (always camelize and reject_nil_values), we wouldn't need the additional logic, and a static code generation would be sufficient (less complex and faster).

There is no relation between using reflection, and the addition of those two options. I would have proposed the same options with generated static code.

aurelien-reeves:
> It's a tradeoff between complexity in the code generator and the runtime code, and I think I prefer to keep the complexity out of the runtime code.

I stick to my implementation for the serialization using reflection over code generation.

Indeed the runtime code may be a little bit more complex, but it has unit tests / specs that are easy to maintain, and I still find more easy to maintain something like this:

```ruby
##
# Returns a new Hash formed from the message attributes
# If +camelize:+ keyword parameter is set to true, then keys will be camelized
# If +reject_nil_values:+ keyword parameter is set to true, resulting hash won't include nil values
#
#   Cucumber::Messages::Duration.new(seconds: 1, nanos: 42).to_h                                 # => { seconds: 1, nanos: 42 }
#   Cucumber::Messages::PickleTag.new(name: 'foo', ast_node_id: 'abc-def').to_h(camelize: true)  # => { name: 'foo', astNodeId: 'abc-def' }
#   Cucumber::Messages::PickleTag.new(name: 'foo', ast_node_id: nil).to_h(reject_nil_values: true)  # => { name: 'foo' }
#
# It is recursive so embedded messages are also processed
#
#   location = Cucumber::Messages::Location.new(line: 2)
#   Cucumber::Messages::Comment.new(location: location, text: 'comment').to_h  # => { location: { line: 2, :column: nil }, text: "comment" }
#

def to_h(camelize: false, reject_nil_values: false)
  resulting_hash = self.instance_variables.to_h do |variable_name|
    h_key = variable_name[1..-1]
    h_key = Cucumber::Messages::Message.camelize(h_key) if camelize

    h_value = prepare_value(
      self.instance_variable_get(variable_name),
      camelize: camelize,
      reject_nil_values: reject_nil_values
    )

    [ h_key.to_sym, h_value ]
  end

  resulting_hash.reject! { |_, value| value.nil? } if reject_nil_values
  resulting_hash
end
```

Than something that would look like this:

```erb
require 'cucumber/messages.dtos'
require 'json'

# The code was auto-generated by {this script}[https://github.com/cucumber/common/blob/main/messages/jsonschema/scripts/codegen.rb]
#

module Cucumber
  module Messages
  <%- @schemas.sort.each do |key, schema| -%>

    class <%= class_name(key) %>

      ##
      # Returns a new <%= class_name(key) %> from the given hash.
      # If the hash keys are camelCased, they are properly assigned to the
      # corresponding snake_cased attributes.
      #
      #   Cucumber::Messages::<%= class_name(key) %>.from_h(some_hash) # => #<Cucumber::Messages::<%= class_name(key) %>:0x... ...>
      #

      def self.from_h(hash)
        return nil if hash.nil?

        self.new(
        <%-
          schema['properties'].each do |property_name, property|
            ref = property['$ref']
            items_ref = property.dig('items', '$ref')
        -%>
          <%= "#{underscore(property_name)}: " -%>
          <%- if items_ref -%>hash[:<%= property_name -%>]&.map { |item| <%= class_name(items_ref) %>.from_h(item) },
          <%- elsif ref -%><%= class_name(ref) %>.from_h(hash[:<%= property_name %>]),
          <%- else -%>hash[:<%= property_name %>],
          <%- end -%>
        <%- end -%>
        )
      end
    end
  <%- end -%>
  end
end
```

And regarding the performance overhead, at the moment cucumber-ruby is still based on an implementation of messages which embeds protobuf. I think the implementation in these PR is already an improvement.

luke-hill:
I'll get onto this properly on friday afternoon. So I can dedicate a bit of time to it.

messages/ruby/lib/cucumber/messages/message/serialization.rb
aslakhellesoy: > The whole deserialization thing, which is the big piece in the current PR, is based on speculation. At first we wanted to implement only the serialization, which is required for the DTOs to work properly, then you requested the deserialization just in case someone else would need it, because it would be a more comprehensive developer experience. We actually do not need it ourselves.

Fair point. I think we should publish a library that will enable people to write consumers in Ruby. See https://github.com/cucumber/common/issues/1623 where someone was asking how to consume messages in Java.

It's speculative, but it also encouragement.

aurelien-reeves: Thank you @aslakhellesoy 🙂

Issue #1550, opened 2021-05-16
Clean up mirror repos


mattwynne:
When we first moved to the monorepo, many of the package managers we used needed a library to be in it's own repo, so we set up individual read-only mirror repos which are copied to when there are changes in this repo.

The context has changed, and we don't need most of these any more. We might still need them for the go modules.

mattwynne:
@jaysonesmith or anyone else from @cucumber/go can you help me understand whether we still need the subrepos for publishing common go modules?

aurelien-reeves:
We can easily specify a go module which is in a monorepo:
https://golang.org/ref/mod#module-path
https://golang.org/ref/mod#vcs-dir
So in our case, that should be as simple as something like the following: github.com/cucumber/common/messages/go/v42

ciaranmcnulty:
Would be great if this eliminates the rsync'ing

aslakhellesoy:
I disagree with `use symlinks instead of rsync and delete rsync scripts/functions` - symlinks work poorly in git.

We may be able to replace some of the copied files with *relative file paths* within the repo, but we might not be able to do it for all rsync'ed files.

For example, I think `.mocharc.json` files must be in the same directory as `package.json` - there might be more cases like that.

ciaranmcnulty:
> symlinks work poorly in git

Is that a Windows issue?

aslakhellesoy:
I thought it was a Windows issue (used to be at least), but have no simple way of verifying if it works now. Maybe it does now: https://github.community/t/windows-symlinks-do-they-play-well-with-git-projects/730

If symlinks work on Windows too, it would certainly simplify things!

ciaranmcnulty:
I have similar vague memories :)

I think under WSL2 it's fine but no idea how it works on 'real' Windows nowadays

ciaranmcnulty:
If we don't have separate repos in future though I agree that we should minimise the use of symlinks in favour of relative paths

mattwynne:
@gasparnagy @aurelien-reeves @enkessler can you add anything about the symlinks situation on Windows these days?

gasparnagy:
@mattwynne did a quick check. you can create symlinks, but only with admin privileges. so with normal console where you would do a git pull it does not work.

How does git support symlinks. Could anyone point me to a git repo that had symlinks inside - i would try to see what happens if i pull...

btw the win command for creating symlinks is

```
mklink symlink-file target-file
```

gasparnagy:
[...]

PR #1608, opened 2021-06-11
Use VCS module paths in Go reference from Gherkin => Messages

mattwynne:
## Summary

Ref #1550 - switch to using VCS module paths within the monorepo instead of pointing to a subrepo.

TODO:

- [x] gherkin -> messages
- [x] demo-formatter -> messages
- [x] json-formatter -> messages
- [x] merge #1616?
- [x] rename branch refs in all `go.mod` files
- [x] wait for this PR to go green

## Motivation and Context

If we can use module paths that point to the directory within the cucumber/common repo, we can get rid of a whole load of complexity around creating subrepo mirrors.

## How Has This Been Tested?

`cd gherkin/go && make`

## Types of changes

### DIFF
--- a/gherkin/go/go.mod
+++ b/gherkin/go/go.mod
@@ -1,7 +1,7 @@
 module github.com/cucumber/common/gherkin/go/v19
 
 require (
-	github.com/cucumber/common/messages/go/v16 v16.0.0-20210616143200-16b27f1a25b2
+	github.com/cucumber/common/messages/go/v16 v16.0.0
 	github.com/stretchr/testify v1.7.0
 )
 
--- a/gherkin/go/go.mod
+++ b/gherkin/go/go.mod
@@ -1,7 +1,7 @@
 module github.com/cucumber/gherkin-go/v19
 
 require (
-	github.com/cucumber/messages-go/v16 v16.0.1
+	github.com/cucumber/common/messages/v16 v16.0.0
 	github.com/stretchr/testify v1.7.0
 )
 
--- a/gherkin/go/go.mod
+++ b/gherkin/go/go.mod
@@ -1,10 +1,10 @@
 module github.com/cucumber/common/gherkin/go/v19
 
 require (
-	github.com/cucumber/common/messages/v16 v16.0.0
+	github.com/cucumber/common/messages/go/v16 v16.0.0-20210616143200-16b27f1a25b2
 	github.com/stretchr/testify v1.7.0
 )
 
-replace github.com/cucumber/messages-go/v16 => ../../messages/go
+replace github.com/cucumber/common/messages/go/v16 => ../../messages/go
 
 go 1.13
--- a/gherkin/go/go.mod
+++ b/gherkin/go/go.mod
@@ -1,4 +1,4 @@
-module github.com/cucumber/gherkin-go/v19
+module github.com/cucumber/common/gherkin/go/v19
 
 require (
 	github.com/cucumber/common/messages/v16 v16.0.0
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ executors:
   # the build nodes (faster builds).
   docker-cucumber-build:
     docker:
-      - image: cucumber/cucumber-build:0.5.0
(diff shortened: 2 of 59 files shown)


mattwynne:
i had a fiddle around but I can't seem to make this work. @lonnblad @jaysonesmith @aslakhellesoy @aurelien-reeves anyone care to take a look at it?

mattwynne:
The error I see is:

```
go: github.com/cucumber/common/messages/v16@v16.0.0: missing github.com/cucumber/common/messages/go.mod and .../v16/go.mod at revision messages/v16.0.0
```

gherkin/go/go.mod
```diff
@@ -1,7 +1,7 @@
 module github.com/cucumber/gherkin-go/v19
 
 require (
-	github.com/cucumber/messages-go/v16 v16.0.1
+	github.com/cucumber/common/messages/v16 v16.0.0
```
aslakhellesoy, line 4: There is no Git tag named `v16.0.0` in the monorepo. all tags are of the form `library/version`, so you need to use something like this:

```
github.com/cucumber/common/messages/v16 messages/v16.0.0
```

I haven't tested if this is valid go module syntax. If it isn't, we have to start using a different naming scheme for tags - perhaps with a hyphen or dot instead of a slash,

mattwynne: It seems to have ways of inferring the tag from the version string that I can't quite fathom yet. If you see the first error (`...at revision messages/v16.0.0`) it seems to have figured out the tag name and found it, but it has the wrong path to the actual `go.mod` file (`missing github.com/cucumber/common/messages/go.mod`)

mattwynne: In the second case it makes a different guess for the tag name, and can't find it in git: `unknown revision messages/go/v16.0.0`

aurelien-reeves: go look for the module on the master branch.
But it does not exist (yet) because messages/go.mod still have `module github.com/cucumber/messages-go/v16`

with the following in `gherkin/go/go.mod` I was able to have a coherent error message:
```
github.com/cucumber/common/messages/go/v16 e0d826edaa5d24e50910eaa689194a8b3340c907

# go: github.com/cucumber/common/messages/go/v16@v16.0.0-20210611214814-e0d826edaa5d: parsing go.mod:
#         module declares its path as: github.com/cucumber/common/messages/v16
#                 but was required as: github.com/cucumber/common/messages/go/v16
```

I think we will have to update the modules like the following (in `messages/go/go.mod`):
```
module github.com/cucumber/common/messages/go/v16
```

So we may need at least 2 PR to make it work:

The first one updating all go.mod replacing the module path from the subrepos to the appropriate monorepo sub-directory (for example `module github.com/cucumber/messages-go/v16` => `module github.com/cucumber/common/messages/go/v16`)

Then a second PR to update all the go `require` directive to point to the new location.

aurelien-reeves:
Refs. https://github.com/cucumber/common/issues/1550

aurelien-reeves (approved):

aslakhellesoy:
Wooot!!

Don't forget to update CHANGELOG before merging this, with an entry under `Changed` so we remember to make a major release next time.

And then I suppose we can archive all the mirrors and remove the mirroring stuff from the build? Including the executable we run to do the actual mirroring.

aurelien-reeves:
> Wooot!!
> 
> Don't forget to update CHANGELOG before merging this, with an entry under `Changed` so we remember to make a major release next time.
> 
> And then I suppose we can archive all the mirrors and remove the mirroring stuff from the build? Including the executable we run to do the actual mirroring.

Changelog have already been updated as part of #1616 
Maybe just add references to the PR into those?

mattwynne:
I think having the CHANGELOGs reference the main issue #1550 is good enough. Do you agree?

We'll need to do the new major releases of messages and gherkin as part of https://github.com/cucumber/godog/issues/389 I guess.

mattwynne:
@aslakhellesoy I've updated the checklist at the top of #1550 - can you take a look and see if you agree with the remaining steps?

PR #1617, opened 2021-06-16
Messages ruby dtos serialization and deserialization (fixes #1605)

aslakhellesoy:
This PR addresses the feedback I gave in my review of https://github.com/cucumber/common/pull/1605

All serialization/deserialization logic is implemented in generated classes. The message superclass with reflection logic is gone - I don't think it would have been possible to implement this properly with reflection unless we load the JSON schema at runtime to get the types for various fields (we don't want that - it would be complicated).

I've also updated the `NdjsonToMessageEnumerator` to emit proper `Envelope` objects and updated the Ruby demo formatter accordingly.


aslakhellesoy:
When I added support for serialization/deserialization of Array fields, I discovered a bug in the schema for `TestStep`, so I fixed that too (all Array fields should be `required` so that they default to an empty array).

This had some ripple-effect to other parts of the monorepo (compatibility-kit, fake-cucumber, react)

messages/jsonschema/scripts/templates/ruby.rb.erb
```diff
@@ -24,6 +24,29 @@ module Cucumber
       attr_reader :<%= underscore(property_name) %>
     <%- end -%>
 
+      def self.from_json(json)
```
mattwynne, line 27: One thing I liked about Aurellien's iteration was that he'd included RDoc with examples of calls to the API. I know we have RSpec examples but I'd like to see some docs on these methods too.

messages/jsonschema/scripts/templates/ruby.rb.erb
```diff
@@ -24,6 +24,29 @@ module Cucumber
       attr_reader :<%= underscore(property_name) %>
```
mattwynne, line 24: I love that we're keeping the underscore / camelCase stuff out of the runtime code!

messages/jsonschema/scripts/templates/ruby.rb.erb
aslakhellesoy: Yeah put them in. if you want to. I was focussed on making the tests pass :-)

messages/jsonschema/scripts/templates/ruby.rb.erb
```diff
@@ -24,6 +24,29 @@ module Cucumber
...
+
+      def self.from_camel_symbol_hash(hash)
+        return nil if hash.nil?
+        self.new(
+      <%-
+        schema['properties'].each do |property_name, property|
+          ref = property['$ref']
+          items_ref = property.dig('items', '$ref')
+      -%>
+        <%- if items_ref -%>
```
mattwynne, line 39: I can't easily comprehend what each leg of the logic is doing here. Can we extract some methods?

aslakhellesoy: The first one is for arrays of objects
The second one is for objects
The third is for primitives and arrays of primitives.

I agree we should extract a method up into the `Ruby` class

aurelien-reeves:
@aslakhellesoy do you want to keep it opened for further discussion?

aslakhellesoy:
> do you want to keep it opened for further discussion?

No, I think it's good to merge now

aurelien-reeves:
#1605 has been merged
It will be difficult to merge both PR 

My question was more "do we keep it opened or do we close it" 😅

aslakhellesoy:
Ah my bad, I was mixing up the two issues. Closing this one.

PR #1619, opened 2021-06-17
Render attachments in examples tables. Closes #1173

### LINKED ISSUE #1173: Attachments and logs not displayed in Example rows
Attachments attached from a hook/stepdef running as part of a Scenario Outline Examples row are not displayed by `@cucumber/react`.

Reproducible by repo mentioned in this tweet: https://twitter.com/kunalashar25/status/1301719227238289408

aslakhellesoy:
This is a fix for #1173.

Unexpanded:

Screenshot 2021-06-17 at 16 41 07

Expanded:

Screenshot 2021-06-17 at 16 41 37


davidjgoss (commented):
Looks good!

One query around image widths. At the moment our only constraint is:

```css
max-width: 100%
```

So if you have a big image (as in this example, but also common with browser screenshots) it will just go as wide as your screen and dominate the page.

Should we have a reasonable upper width limit e.g. 1200px? If so, should that be different within the examples table?

aslakhellesoy:
Is it a problem that the image takes the full width? It's collapsed by default.

BTW, my change in the CCK seems to have triggered a test failure in `json-formatter` - I need to investigate that.

davidjgoss:
> Is it a problem that the image takes the full width? It's collapsed by default.

Yeah I think that's a fair point, you're opting into it. Sold!

davidjgoss (approved):

PR #1621, opened 2021-06-20
Use better issue templates

mpkorstanje:
The current issue templates are mostly ignored. The ones provided by github
don't tend to be so much.


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,36 @@
...
+title: ''
+labels: Bug
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
```
luke-hill, line 13: Can we mention something here about having either a repo or an MVCE?

mpkorstanje: Unfortunately most Cucumber users have a limited amount of experience with either. I'm afraid that by asking for anything specific will just get us nothing at all instead.

PR #1624, opened 2021-06-21
datatable/java: DataTable print with configurable indent + without escaping

artysidorenko:
## Summary

Updates DataTable with a new print method allowing to configure custom indention, and skip escaping the cell contents.

## Details

- makes the `TablePrinter` class configurable, via a new `TablePrinterBuilder` class.
- adds a new method `printFormatted()` to the `DataTable` class, as it already includes similar publicly-exposed `print()` methods
- above method configures the `TablePrinter` with custom indent + removing cell-escaping.

(I'm still not too familiar with the codebase; the above felt like reasonable decisions, but am happy to incorporate feedback if you prefer doing it different)

## Motivation and Context

Relates to [issue-2320 on cucumber-jvm](https://github.com/cucumber/cucumber-jvm/issues/2320).

Ultimate aim is to include data-tables in the printed test output. E.g. in the following feature, current behaviour prints out only the step definition, but it skips the info in the table.

```gherkin
Given I have a list of people
  | NAME  | AGE |
  | John  | 25  |
  | Sally | 28  |
```

## How Has This Been Tested?

Added 2 unit tests. `can_print_table_with_escape_characters_regular` confirms the escape-cell behaviour of the original `print()` method. And `can_print_table_with_escape_characters_formatted` tests the behaviour of the new `printFormatted()` method.

## Types of changes

## Checklist:

(don't think it requires documentation/porting to other languages as it's a small change?)


datatable/java/datatable/src/main/java/io/cucumber/datatable/DataTable.java
```diff
@@ -642,6 +642,25 @@ public void print(StringBuilder appendable) {
...
+     * table to the {@code appendable},
+     * with custom indentation
+     * and without escaping cell contents
+     *
+     * @param appendable to append the string representation
+     *                   of this table to.
+     * @param indent     level of start indentation
+     * @throws IOException
+     */
+    public void printFormatted(Appendable appendable, String indent) throws IOException {
```
mpkorstanje, line 656: This isn't scaling very well. For each different way to print we'd end up with a new method. Might be better to have the `TablePrinter` as part of the public API and drop the `print*` methods from the data table. 

The table printer could take the `DataTable` as an argument and use `.cells()` to get the cells.

I can try to do this or you can if you want.

artysidorenko: Just pushed an update, if that's more what you had in mind? Am available next couple of days to fix it up further, but equally my ego won't be bruised if you prefer to make some changes directly :)

datatable/java/datatable/src/main/java/io/cucumber/datatable/TablePrinter.java
```diff
@@ -2,30 +2,46 @@
...
+        this.startIndent = startIndent;
+        return this;
+    }
+
+    public TablePrinter escapeCells(Boolean escapeCells) {
+        this.escapeCells = escapeCells;
+        return this;
+    }
+
+    public void printTable(DataTable table, StringBuilder appendable) {
```
artysidorenko, line 26: wasn't sure if your preference was to pass the table in the method, or when first calling the `TablePrinter()` constructor. I figured it's more likely this could be used to print multiple different tables with the same print configuration, so best to keep it as a method argument.

datatable/java/datatable/src/main/java/io/cucumber/datatable/DataTable.java
artysidorenko: (also removed the printer-builder class from the last commit, and allowed for configuring the printer class directly, to minimise number of files, but I don't feel strongly about it)

datatable/java/datatable/src/main/java/io/cucumber/datatable/DataTable.java
```diff
@@ -620,26 +620,32 @@ public String toString() {
     /**
      * Prints a string representation of this
      * table to the {@code appendable}.
+     * 
+     * @deprecated superseded by TablePrinter.printTable()
```
mpkorstanje, line 624: You can use `@link` to link to a method in Java doc.

```suggestion
     * @deprecated superseded by {@link TablePrinter#printTable(DataTable, Appendable)}
```

datatable/java/datatable/src/main/java/io/cucumber/datatable/TablePrinter.java
mpkorstanje: :+1:

datatable/java/datatable/src/main/java/io/cucumber/datatable/DataTable.java
mpkorstanje: Getting close!

The idea is to use the builder pattern to create immutable table printer that can be used to print many different tables. Using the builder pattern to create immutable objects does take a bit more code but makes the objects much easier to work with. Especially in multi-threaded contexts.

So we have a `TablePrinter` with these public methods:

````
TablePrinter
 - static TablePrinter.Builder builder()
 - printTable(DataTable table, StringBuilder appendable)
 - printTable(DataTable table, Appendable appendable)
```

And a `TablePrinter.Builder` with these public methods:

```
TablePrinter.Builder
 - indent(int)
 - escape(boolean)
 - TablePrinter build()
```

I realize this may not be apparent from the surrounding code but a lot of this code was written in a time before these ideas were common.

datatable/java/datatable/src/main/java/io/cucumber/datatable/TablePrinter.java
mpkorstanje: You've got it.

datatable/java/datatable/src/main/java/io/cucumber/datatable/DataTable.java
artysidorenko: cool yeah, that makes sense.

Struggled how best to handle the child `DiffTablePrinter` class, whether it should have its own builder, or call the parent constructor with the parent's builder (for now went with the latter)

mpkorstanje: I had a look. Replaced inheritance with composition by extracting the `indent` as a function that returns a specific indent prefix for each row.

mpkorstanje (commented):
I've rewritten the rendering logic from scratch to make the table printer thread-safe. Coincidently this also simplified the code a whole lot. Added a few tests to confirm I didn't break anything.

Thing I'm still wondering about is the naming. I'm not happy about the `Printer` suffix.

Comparable bits of code in the JDK are named `Formatter` and use the `format(Thing) : String`  `formatTo(Thing, Appendable)`.

artysidorenko:
Looks good, thanks!

The only difference I noticed is that the comparable `java.util.Formatter` tends to keep the appendable as a property of the instance, but I think the `formatTo` method makes the behaviour pretty clear here 👍

aslakhellesoy:
Hi @artysidorenko,

Aslak Hellesøy
Creator of Cucumber

mpkorstanje:
I'll have to find some time to make the release.

mpkorstanje:
Done. Unfortunately there are some breaking changes that precede your change. If you want to continue this in Cucumber JVM. I would suggest branching of [this commit]( https://github.com/cucumber/cucumber-jvm/commit/b73b38b5acea8db436b5ddcc41e82dc278087d3d ) in the `v7.x.x` branch.

mpkorstanje:
Nevermind that. Branch of `main`. v7 is too broken.

PR #1639, opened 2021-07-01
cck: use predictable ids and timestamps

davidjgoss:
## Summary

For generating the golden master for compatibility-kit, use `--predictable-ids` when calling fake-cucumber.

## Motivation and Context

Every time we add/change something in the compatibility kit, every line of every file in the golden master changes because of the ids and timestamps - this makes it difficult to see the actual changes and also to find any meaning in the git history.

We wouldn't expect Cucumber implementations to match these when validating against the golden master messages - it's just about reducing the churn in the source.

## Checklist:


aslakhellesoy (approved):
Thanks @davidjgoss - a simple and very useful improvement!

PR #1648, opened 2021-07-08
Fix @types/node to 14

aurelien-reeves:
@types/node should match the version of nodejs we are using


davidjgoss (approved):

aurelien-reeves:
I just want to wait for the CI to be green before merging it. To be sure it is actually ok for real 😅

PR #1652, opened 2021-07-10
Use native RegExp Match indices

aslakhellesoy:
## Summary

Use [RegExpExecArray#indices](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec) to get the position of capture groups.

## Details

This PR removes the dependency on the [becke-ch--regex--s0-0-v1--base--pl--lib](https://www.npmjs.com/package/becke-ch--regex--s0-0-v1--base--pl--lib) library and uses `RegExpExecArray#indices` instead.

[RegExp Match Indices](https://github.com/tc39/proposal-regexp-match-indices) were introduced in v8 v9.0.259. The first Node.js release to use a v8 newer than that is [16.4.0](https://nodejs.org/en/download/releases/).

Since we cannot drop support for Node.js < 16.4.0 until 2023/2024, the implementation currently uses a [polyfill](https://github.com/rbuckton/regexp-match-indices).

## Motivation and Context

The main motivation for this change is to stop depending on http://www--s0-v1.becke.ch/tool/becke-ch--regex--s0-v1/becke-ch--regex--s0-0-v1--homepage--pl--client/, which seems to be a liability to depend on because it doesn't seem to have a public Git repo, except for [this fork](https://github.com/TheWorldTable/regexp-plus).

I would also expect this to be slightly more performant since it's based on a native implementation, but I doubt it will be significant for end users.

Also see https://github.com/cucumber/common/issues/300

## How Has This Been Tested?

The tests pass on Node v16.4.2, but I doubt it will pass on Node < 16.4.0 as explained above.

## Types of changes

## Checklist:

### DIFF
--- a/cucumber-expressions/CHANGELOG.md
+++ b/cucumber-expressions/CHANGELOG.md
@@ -20,6 +20,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+* Use native RegExp Match indices (currently relying on a polyfill)
+  ([#1652](https://github.com/cucumber/common/pull/1652)
+   [aslakhellesoy])
+
 ## [12.1.1] - 2021-04-06
 
 ### Fixed
--- a/cucumber-expressions/javascript/package.json
+++ b/cucumber-expressions/javascript/package.json
@@ -34,6 +34,7 @@
     "typescript": "4.3.5"
   },
   "dependencies": {
+    "regexp-match-indices": "1.0.2"
   },
   "directories": {
     "test": "test"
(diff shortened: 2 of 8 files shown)


aslakhellesoy:
Update: I added the polyfill so we don't have to wait until 2023/2024 to merge this.

aurelien-reeves (approved):
👍

mattwynne:
👍 

If we're using the polyfill let's edit the text in the PR to reflect that and remove the warnings.

aslakhellesoy:
PR text updated

Issue #1306, opened 2021-01-13
Gherkin language server support


laeubi:
## Summary

Gherkin already provides a process to parse feature files from the commandline as well as parsing "in code". This still needs adaption when used, e.g. in an IDE. For some time now there is a [Language Server Protocol](https://en.wikipedia.org/wiki/Language_Server_Protocol) that allows to support the easy integration in multiple IDEs with less code.

## Expected Behavior

The process can be started with 'gherkin --server' behaving according to the LSP4 protocol using the std input/output stream for communication.

## Current Behavior

currently it is only possible to parse feature files one-time with a cucumber specific json response

## Context & Motivation

See summary, this would allow to integrate advanced editing capabilities to a majority of IDEs without "reinvent the wheel"

## Your Environment

The users operating system that is capable of executing the gherkin command.

aslakhellesoy:
Hi @laeubi - I think this could be incredibly useful.

Is this something you'd be willing and able to contribute, or kick-start? I personally do not have capacity to do this now.

laeubi:
@aslakhellesoy I'm currently doing some investigations on how this could be archived using LSP4J (java) but if someone else can do it better and faster (e.g. GO, C+++, ...) feel free to raise a hand :-)

I can't promise when I could came up with a prototype but will keep this in mind.

aslakhellesoy:
@SabotageAndi where is that VSCode LSP implementation you mentioned?

SabotageAndi:
This is the mentioned VS Code Extension: https://marketplace.visualstudio.com/items?itemName=alexkrechik.cucumberautocomplete
Everything with autocomplete is done in VS Code via LSP.

aslakhellesoy:
This is in progress: https://github.com/cucumber/common/pull/1689

PR #1653, opened 2021-07-12
Suggest

aslakhellesoy:
## Types of changes

## Checklist:


laeubi:
@aslakhellesoy I have written a very basic Gherkin autocomplete for the eclipse-java plugin based on the eclipse-api, so an official API would be great, even in the form of some rules that could be parsed and translated to some other techniques.

laeubi:
This is also related to https://github.com/cucumber/common/issues/1306

aslakhellesoy:
What does the eclipse plugin use for suggestions @laeubi?

The other Gherkin auto-complete plugins I have seen suggest regular/cucumber expressions from step definitions that the user has to edit into a step afterward. Does the Eclipse plugin do the same thing? If so, I think a Java port of this library could improve suggestions considerably for Eclipse.

gasparnagy:
I'm on vacation, so just a sort feedback: I like the concept. Two things came to my mind based on my auto-complete implementations:

1. For SpecFlow we would need to also record whether the particular step definition can be used as Given, When, Then or all -- but this can be added later.

1. In large codebases the parameter variations can be too big. So I would apply some maximum threshold for the possible parameter values (like take the top 10).

aslakhellesoy:
Thanks for the feedback @gasparnagy! I’ll add the max thing. I might as well add the GWT bit too.

unless there are objections I’ll merge and release 0.0.1 of this tonight. Dying to try it out on some real data (in Cucumber Reports).

mattwynne:
This looks fantastic!

When handling parameters (e.g. ` ${2|belly,suitcase,table|}`), does the LSP allow us to express constraints on custom values? We have the regex from the parameter type, so theoretically that could allow the editor validate a custom value, if the LSP allows for that.

aslakhellesoy:
> This looks fantastic!

Much obliged :-)

> When handling parameters (e.g. ${2|belly,suitcase,table|}), does the LSP allow us to express constraints on custom values?

I don't think the LSP allows constraints to be defined on choices, nor do I think it is a good idea. An editor should never *prevent* users from typing something, even if it is wrong. That would be a patronising and annoying user experience.

Instead, it should *highlight* that you've typed something invalid, and this is what editors do, using syntax validation. So if the user doesn't want `42` or `54` (from the `${1|42,54|}` parameter), they should be allowed to press `ESC` and insert `horse`, even if the underlying parameter is a Cucumber Expression `{int}` or a Regular Expression `\d+`.

In this case the editor should simply highlight the line indicating a syntax error (it no longer matches any step definitions).

You can read more about the LSP snippet syntax [here](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#snippet_syntax).

aslakhellesoy:
@gasparnagy wrote:

> For SpecFlow we would need to also record whether the particular step definition can be used as Given, When, Then or all -- but this can be added later.

I think the best simplest way to handle this is to build three sets of `StepDocument` - one for each of Given, When and Then.

These three sets of documents can then be added to the four indexes you describe. When the editor sends an auto-complete request, it can send context about what index it wants to search, and the LSP server can use that information to look in the corresponding index.

If the document sets are to be built from Cucumber Messages, we'll have to add a new `given|when|then` enum field on the [StepDefinition](https://github.com/cucumber/common/blob/main/messages/messages.md#stepdefinition) message. The [Step](https://github.com/cucumber/common/blob/main/messages/messages.md#step) message already has a `keyword` field (which would have to be mapped back to the `given|when|then` enum.

davidjgoss:
Nice!

jbpros:
I finally took the time to look at this. It is really cool (I also love the executable README, btw 😉)!

I can't wait to get auto completion in VScode!

aurelien-reeves:
Finally had some time to take a look

Indeed, the executable readme, this is really great!

Regarding `suggest` itself, I am concerned about the memory consumption for StepDocuments.

I may say something silly: maybe we could bring sqlite as storage engine? Would that be doable?

aslakhellesoy:
At what memory size would you consider consumption to be problematic @aurelien-reeves?

Roughly how many (by order of magnitude) `StepDocument` objects would you estimate would be needed to reach that critical level?

aurelien-reeves:
We would have to benchmark to have proper idea actually

That was just a rough idea

aurelien-reeves:
> At what memory size would you consider consumption to be problematic @aurelien-reeves?
> 

That's a difficult question actually.
If we consider only `suggest`, we may consider that we could easily take benefit from GB of user-memory and that it won't be an issue.

But actually it will be part of bigger tools like IDEs, or in-browser code editors. Those tools are already huge memory consumer. So I think we may try to reduce that impact on user-memory if this is possible and not too difficult.

As this is just the beginning of the project, that would be a good opportunity to think about that right now as it may be easier than with a more mature project.

gasparnagy:
Having slept on that and reading the questions of @aurelien-reeves made me think... Somehow I miss the link of this service to the real usage. Maybe you have had concrete discussions about that, but now I feel that without any concrete plan/need for using it, we might go into a wrong direction. Now I remember, that when I did this for the SpecFlow Visual Studio extension, the real challenge was not to find and collect the usages, but how this can be integrated into a real IDE. And those challenges were not foreseeable. I remember that I was also worried by the memory consumption, but finally the two key challenges were:

1. How to build the StepDocuments in a reasonable time. (In a larger project parsing through all feature files takes a few seconds, so you definitely cannot do this when the auto-complete is triggered.)
2. If you want to cache the StepDocuments (because of the challenge above), how can you (partially) invalidate/update this cache. 

I have never implemented an LSP server, but probably doing so would bring up a few other questions as well. So probably before we move forward with this, we should really find a concrete project that needs this (maybe Cucumber Studio, maybe something else) and try to integrate it there and see how it works on a project with 500+ feature files.

gleiderr:
It would be great to see a decent autocomplete in VS Code.
Unfortunately I don't have enough time right now to implement using this solution.

aslakhellesoy:
@aurelien-reeves @gasparnagy I have updated the documentation with an [architecture](https://github.com/cucumber/common/tree/main/suggest/javascript#architecture) section.

A full suggest system would need an ETL process to build `StepDocument` objects and store them in persistent storage. This process would not run in the editor - it would be a separate process running in the background. It could be triggered whenever the step definitions or gherkin steps change.

The ETL process is probably the most memory hungry part of the system, since it would need to load all step definitions and gherkin steps into memory. For an online editor, it could run on a server, and would therefore not affect memory consumption on the machine running the editor.

For a desktop editor (such as VS Code) it would have to run on the same machine, but still in a separate process - either in the LSP server, or even as a completely separate process.

To get a rough idea of memory consumption, we could consider a system where there are 1000 feature files with 200 steps in each. Let's assume the average length of a step is 80 bytes - that would amount to 16Mb. Maybe there are 1000 step definitions, each with an average length of 100 bytes - that's another 100Kb (negligible compared to steps).

I think some benchmarks are needed in order to get a sense of how much memory be consumed, as well as how long it would take.

Regarding real usage - I have already built an editor prototype using Monaco Editor and a custom plugin with an in-memory lucene-like index. It seems to work pretty well, but I have not tried with a large index.

Regarding point 1 @gasparnagy - I believe that would be addressed by building the index in a background process. There may be a few seconds lag to update the index when new step definitions are added, but I doubt the users would notice - most of the time they would be writing steps using data that is already in the index.

Regarding point 2 I am not sure partial update of the index is needed. As I said, I don't think usability will suffer significantly if the index is only updated sporadically, for example with a debounce of 10 seconds inactivity and max once a minute for example.

Regarding a concrete project that needs this - we are currently planning on using this in Cucumber Studio and/or Cucumber Reports (that's why I started this library), so I expect integrating early versions of this system in those products will quickly give us more feedback. And we'll open source as much as we can (the LSP stuff and/or the Monaco plugin).

aurelien-reeves:
Thanks for all those explanation @aslakhellesoy 👍

PR #1655, opened 2021-07-13
Convert ruby messages duration.nanos to integer

aurelien-reeves:
`duration.nanos` is described in the schema as being an integer. It was already the case with protobuf.

With the new DTOs, `duration.nanos` became a float. When using the json-formatter to convert ndjson to "legacy" json, an error is raised:

```
ERROR: json: cannot unmarshal number 127100.0 into Go struct field Duration.testStepFinished.testStepResult.duration.nanos of type int64
```

Converting nanos to integer should fix this.

## How Has This Been Tested?

As durations are removed when validating messages as part of the neutralize-json scripts, this is not properly tested yet in the mono-repo.
I encounter this issue while trying to reactivate the CCK tests with cucumber-ruby. So there will be at least that way to test it.

## Types of changes

## Checklist:


aslakhellesoy:
I think we should also patch json-formatter to handle floats, so it works with those ruby versions that output a float.

aurelien-reeves:
> I think we should also patch json-formatter to handle floats, so it works with those ruby versions that output a float.

For now no ruby version output float at the moment.
However I check how difficult that would be

aurelien-reeves:
From some tests I did on cucumber-ruby main branch but not released yet

aurelien-reeves:
> 
> 
> I think we should also patch json-formatter to handle floats, so it works with those ruby versions that output a float.

Due to the strong type system of Golang, that may be doable but it looks pretty difficult.
As the issue has not been released yet, would it be ok to merge and release this PR quickly in order to avoid releasing this known bug with cucumber-ruby? That way, we would not have to patch json-formatter

Issue #871, opened 2020-01-22
Message based cucumber-junit-formatter


aslakhellesoy:
We need a `cucumber-junit-formatter` executable that reads messages from `STDIN` and writes JUnit XML to `STDOUT`. Any Cucumber implementation with a `message` formatter can then pipe generated messages to this formatter to generate JUnit XML.

Older Cucumbers with built-in JUnit formatters will have those removed when they support messages and people will have to use the `cucumber-junit-formatter` executable instead. The existing ones are:

* [JUnitFormatter.java](https://github.com/cucumber/cucumber-jvm/blob/master/core/src/main/java/io/cucumber/core/plugin/JUnitFormatter.java)
  * [test data](https://github.com/cucumber/cucumber-jvm/tree/master/core/src/test/resources/io/cucumber/core/plugin)
* [junit.rb](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/formatter/junit.rb)
  * [test data](https://github.com/cucumber/cucumber-ruby/blob/master/features/docs/formatters/junit_formatter.feature)

As pointed out in https://github.com/cucumber/cucumber-ruby/issues/1384 the generated XML should contain the steps (like the Java implementation does).

In order to be portable, `cucumber-junit-formatter` should be implemented in Go. Look at `demo-formatter/go` for inspiration.

vincent-psarga:
I don't think it should be a stand alone tool. I think there should be a `cucumber/junit-formatter` in the mono-repo, written in all languages currently supported (Java, Ruby and JS). We could of course take advantage of the messages portability to have a common set of tests.
JUnit is one of the most widely used output format, and I definitely think it should built-in all cucumber releases, without having to add an extra tool

mpkorstanje:
Might be of interest that the JUnit Team is currently working on a new format. I would wait too see how that turns out before we spend too much time implementing a structural solution. 

https://github.com/junit-team/junit5/issues/373

mpkorstanje:
Also worth nothing that the JUnit formatter works just fine, even with the message protocol. So I don't see a need to remove it. I can see a use case for implementing it as a polyglot component with a common acceptance test suite.

denim2x:
@aslakhellesoy Please explain why this tool should be standalone; wouldn't a transpiled multi-language (like [Haxe](https://haxe.org/manual/introduction-what-is-haxe.html)) module do?

aslakhellesoy:
[...]

PR #1659, opened 2021-07-15
cck: add cases for pending, skipped and undefined

davidjgoss:
## Summary

Add cases to the compatibility kit for:

- Pending steps
- Skipped scenarios via hooks/steps
- Undefined steps

## Motivation and Context

We want to document and standardise the behaviour of these features. We'll also need them in order to use the CCK as the reference point for a JUnit formatter (see #871).

## How Has This Been Tested?

Test added to fake-cucumber which now supports `skipped`.

## Types of changes

## Checklist:


compatibility-kit/javascript/features/pending/pending.feature
```diff
@@ -0,0 +1,13 @@
...
+  not yet implemented (or "pending") by returning or throwing a particular
+  value.
+
+  This causes subsequent steps in the scenario to be skipped, and the overall
+  result to be treated as a failure.
+
+  Scenario: Unimplemented step signals pending status
+    Given a step
+    When a step that isnt implemented yet
+    Then a step
```
mpkorstanje, line 13: Dropping the other two steps would simply this.

compatibility-kit/javascript/features/skipped/skipped.feature
```diff
@@ -0,0 +1,16 @@
...
+  for running the scenario.
+
+  @skip
+  Scenario: Skipping from a Before hook
+    Given a step
+
+  Scenario: Skipping from a step
+    Given a step
+    When a step that skips
+    Then a step
```
mpkorstanje, line 16: Same as before. There are a few more.

compatibility-kit/javascript/features/pending/pending.feature
davidjgoss: I wanted the messages that come out of this to show that the step before runs and passes, and the step after is skipped - I think that's key to properly implementing this?

davidjgoss: Maybe I should make this explicit in the wording of these steps?

mpkorstanje: > I wanted the messages that come out of this to show that the step before runs and passes, and the step after is skipped - I think that's key to properly implementing this?

I think that should be a separate test case then. Could be much more generic. Any step after a non-passing step is skipped.

davidjgoss: I kind of prefer to have the test cases for pending etc be fairly complete even if there is a little duplication overall with the skip-subsequent-steps concept, especially as there is some patchiness in our implementations (e.g. ruby doesn't support pending or skipped it seems).

mpkorstanje: I understand, but when the actual and expected result differ it is very hard to to see what the problem is when there is extra fluff laying around. 

Consider instead using multiple examples:

```
  Scenario: Unimplemented step signals pending status
    Given a step that isnt implemented yet

  Scenario: Steps after unimplemented steps are skipped
    Given a step that isnt implemented yet
    Then a skipped step

  Scenario: Steps before unimplemented steps are executed
    Given an implemented step
    Then a step that isnt implemented yet
```

davidjgoss: Thanks - I've refactored based on your suggestion above.

aurelien-reeves (approved):
Looks good to me

Issue #1592, opened 2021-06-03
react: adopt styling from cucumber reports


davidjgoss:
Current:

![image](https://user-images.githubusercontent.com/3192745/120641691-ce806d80-c46b-11eb-97b3-40be49960efb.png)

https://reports.cucumber.io/

![image](https://user-images.githubusercontent.com/3192745/120641721-db04c600-c46b-11eb-99f8-3055ba0094b3.png)

Will need to:

- [x] Get #1302 updated and merged first
- [x] Copy styling from Cucumber reports for:
  - [x] Statuses summary (counts and colours)
  - [x] Execution summary (stats, timing, meta stuff)
- [x] Come up with some good styling for the ci and git stuff
- [x] Come up with some good styling for the search/filter bit

This is also a chance to finish modularising the styling and finally remove the global stylesheet - this should hopefully make it a bit easier for people to use these components in React apps.

davidjgoss:
(working on this now)

PR #1660, opened 2021-07-15
gsModularise cucumber react

aslakhellesoy:
## Summary

Ensure all components in `@cucumber/react`

## Details

This changes most components to `export const SomeComponent = ...` and removes the `export default SomeComponent`.
The motivation for this is to more easily export everything in a directory, and do it in a nested fashion.

From the consumer's perspective, this means components are now imported like this:

```typescript
import { components } from '@cucumber/react'
const { DocString, Tags } = components.gherkin
```

This is a bit more verbose than the previous:

```typescript
import { DocString, Tags } from '@cucumber/react'
```

But since this library has grown quite big, I feel it's a bit tidier to have a bit more namespacing rather than cramming everything into the top-level import. Trying to do that also makes it easier to forget to export something, whereas now, the various directories' `index.ts` spreads this responsibility over multiple files, so it's easier to spot an omission.

I've also made the `<Feature>` component styleable.

## Motivation and Context

Make sure everything is exported and make things a bit more organised

## How Has This Been Tested?

With StoryBook and tsc - and existing tests are still passing.

## Types of changes

## Checklist:


react/javascript/src/styles/styles.scss
```diff
@@ -28,12 +28,6 @@ $summary-border-color: rgba(0, 0, 0, 0.1);
   }
 }
 
-// TODO encapsulate in a component
-.cucumber-children {
```
davidjgoss, line 32: I think this may be used in more places than the `<Feature>` component?

davidjgoss (commented):
Looks good, agree with the approach. We could also look at enabling something like:

```typescript
import { DocString } from '@cucumber/react/components/gherkin'
```

I think that would require us being very deliberate about the file structure of the compiled output and the `.d.ts` files but would probably be possible.

> Make all components customisable with component overrides

Do we really mean every component in here? Some are a bit in flux (especially the "chrome" stuff that we have yet to restyle/redesign, plus the accordion might be replaced as you've said before) and I'd be wary of effectively expanding the public API surface more than necessary if there's not an obvious use case.

> Demonstrate how to override component styles with TailwindCSS by defining classes with @apply in a Storybook story

My next task on #1592 is to add tailwind to the tooling - I can do a standalone PR for that unless you have already started?

react/javascript/src/styles/styles.scss
aslakhellesoy: Fixed in 53cc7f6b27720d4d9aa9fff14cde50bf7097f306 - added a `<Children>` component

aslakhellesoy:
> My next task on #1592 is to add tailwind to the tooling

Ah yes, removed the tailwind TODO from this PR

PR #1665, opened 2021-07-18
[gherkin perl] general cleanup by upgrading the Perl dependency

ehuelsmann:
## Summary

Miscelaneous cleanup made possible after upgrading the Perl minimum requirement to 5.12.0 (which was released in April 2010).

## Details

* Since 5.12, Perl can open filehandles to in-memory byte sequences. Remove an external depenency (`IO::Scalar`) by taking advantage of that. 
* Additionally, since 5.12, all file handles are blessed references into `IO::File`, which removes the need to explicitly load and create `IO::File` instances.
* Close the input file as soon as the last line is read, instead of when the TokenScanner goes out of scope; it's simply better form to do that early

## Motivation and Context

The resulting code is overall more clean and more modern Perl.

## How Has This Been Tested?

The test suite runs on this change as expected.

## Types of changes

- [x] General cleanup.
- [x] Dependency update.

## Checklist:

- [x] The change does not need to be ported to other implementations
- [x] Existing tests cover this change.
- [x] My change does not require a change to the documentation.
- [x] No documentation changes are required.


aurelien-reeves:
@ehuelsmann is this PR ready to be merged?

ehuelsmann:
@aurelien-reeves yes, it is. I was waiting for an answer to the effect on version numbering; @olleolleolle explained that upping the minimum version should be only a minor version upgrade. So, lets merge!

ehuelsmann:
Merging with the implied consent.

PR #1670, opened 2021-07-21
Initial implementation of Cucumber messages for Perl

aurelien-reeves:
## Types of changes

## Checklist:


messages/jsonschema/scripts/templates/perl.pm.erb
```diff
@@ -0,0 +1,96 @@
...
+
+use constant
+<%- property['enum'].to_a.each_with_index do |value, index| -%>
+   <%= property_name.upcase %>_<%= enum_constant(value) %> => '<%= value %>',
+<%- end -%>
+   ;
+
+<% end -%>
+has <%= underscore(property_name) %> =>
+    (is => 'rw',
```
aurelien-reeves, line 80: For info, with the Ruby DTOs, it has been chosen to use read-only attributes

messages/jsonschema/scripts/templates/perl.pm.erb
```diff
@@ -0,0 +1,137 @@
...
+       map { blessed( $_ ) ? $_ : Cucumber::Messages::<%= class_name(items_ref) %>->new( %$_ ) }
+       @{ $args{<%= underscore(property_name) %>} }
+       ]
+       if $args{<%= underscore(property_name) %>};
+
+<%-    end
+    end -%>
+<%- if property['$ref'] -%>
+    $args{<%= underscore(property_name) %>} =
+       Cucumber::Messages::<%= type_for(class_name(key), property_name, property) %>->new( %{ $args{<%= underscore(property_name) %>} } )
```
aurelien-reeves, line 78: This seems to generate things like the following:
```perl
Cucumber::Messages::Cucumber::Messages::Location->new()
```

Is this expected?

ehuelsmann: No. Thansks for spotting that. The serializer/deserializer test doesn't capture this problem, so I'll need to implement more tests :-).

This error is a consequence of having changed the definition of `class_name()` and not working the consequences fully through. ETOOMUCHLATENIGHTCODING?

I'll fix in a sec.

aurelien-reeves:
Caution: there is already a `default.mk` for perl in the templates.

At some point we will have to move your code from the `default.mk` you have created into the one in `templates/perl`, and `rsync` the files to other modules which have perl implementations.

ehuelsmann:
> Caution: there is already a `default.mk` for perl in the templates.
> 
> At some point we will have to move your code from the `default.mk` you have created into the one in `templates/perl`, and `rsync` the files to other modules which have perl implementations.
> 
> Edit: I noticed that there may have been an attempt in #1673?

Yes,  that was an attempt, based on misunderstanding the situation. I've since merged a better version and rebased this branch. By setting up a '.rsync' file to ensure the shared infrastructure remains current, this new library should now be integrated in the regular build infrastructure.

messages/jsonschema/scripts/templates/perl.pm.erb
```diff
@@ -0,0 +1,146 @@
...
+
+__END__
+
+=head1 LICENSE
+
+Please see the included LICENSE.txt for the canonical version. In summary:
+
+The MIT License (MIT)
+
+Copyright (c) 2021 Erik Huelsmann
```
aslakhellesoy, line 138: I think we might need to change this to something else that represents to Cucumber Open community.

We’ve been using Cucumber Ltd for other repos, but that company is defunkt. I’m not sure we ahould use SmartBear either, since it might deter people from contributing.

@mattwynne @aurelien-reeves what do you think? Time to revisit the license owner of all the cucumber/* repos?

aslakhellesoy (changes requested):

messages/jsonschema/scripts/templates/perl.pm.erb
ehuelsmann: Let me know what it should be. if there were a legal body, we would be able to work with ICLA or CCLA constructs like the Apache Foundation does. That would be the  most sound solution. I can change it to "The Cucumber Project", but maybe that's just an empty shell/meaningless term at the moment?

ehuelsmann: @mattwynne @aslakhellesoy maybe it's consistent for now to (at least) add Cucumber Ltd for now? That way, it's consistent with the other repositories.

aslakhellesoy: That would be great - thanks. Then we can address the license owner issue on a broader scale later.

ehuelsmann: Ok. I actually already set the copyright holder in the metadata to Cucumber Ltd: https://github.com/cucumber/common/pull/1670/files#diff-e4960b209d869c770dab6dae20e4e07a5823e193b23d01967012993cde5c907eR12 I'll submit an additional commit to add Cucumber Ltd to the author/copyright in the generated messages file.

ehuelsmann: How about this instead:

```perl
=head1 LICENSE

Please see the included LICENSE for the canonical version. In summary:

The MIT License (MIT)

  Copyright (c) 2021 Erik Huelsmann
  Copyright (c) 2021 Cucumber Ltd

This work is loosely derived from prior work of the same library for Ruby,
called C<cucumber-messages>.

=cut
```

?

ehuelsmann: I guess my question is: this seems to be in line with the Gherkin/Perl module; maybe that was added in another era though?

mattwynne: Cucumber Ltd is wholly owned by SmartBear, so really they're interchangeable. I can't comment on whether it would deter people or not.

I don't really understand what it means to copyright something with an MIT license anyway. Does it enable you to sue someone if they tried to claim they made it instead of you?

ehuelsmann: As the copyright holder, you're allowed to change the license as you own the creative rights to the code. With the MIT license, the distributor and user can do with it what they want (even create derived works), but they can't change the license without the consent of the original copyright holders.

messages/jsonschema/scripts/templates/perl.pm.erb
```diff
@@ -131,12 +131,12 @@ __END__
...
 
-Please see the included LICENSE.txt for the canonical version. In summary:
+Please see the included LICENSE for the canonical version. In summary:
 
 The MIT License (MIT)
 
-Copyright (c) 2021 Erik Huelsmann
-
+  Copyright (c) 2021 Erik Huelsmann
+  Copyright (c) 2021 Cucumber Ltd
```
ehuelsmann, line 139: The code on which this implementation is loosely based (the ruby messages implementation), lists an MIT license with Cucumber Ltd copyrights.

messages/jsonschema/scripts/templates/perl.pm.erb
mattwynne: I think this is good enough for now.

messages/jsonschema/scripts/templates/perl.pm.erb
aurelien-reeves: As https://github.com/cucumber/common/pull/1670#discussion_r683503958 and as discussed during our last community meeting: this is good for now as part of this PR.

PR #1673, opened 2021-07-22
Expand template build infrastructure for Perl components

ehuelsmann:
Simply add generic configuration to spawn additional Perl subprojects off from. Using a PR because I can't commit directly to `main`.

### DIFF
--- a/.templates/perl/Makefile
+++ b/.templates/perl/Makefile
@@ -1,0 +1,10 @@
+include default.mk
+
+default: test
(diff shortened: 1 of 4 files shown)


ehuelsmann:
Removing skip ci because the tests need to run before I can merge...

ehuelsmann:
Based on the failure, this was a bad idea. I don't want Makefile, dist.ini or cpanfile to be overwritten.

PR #1689, opened 2021-08-11
Cucumber LSP server

aslakhellesoy:
## Summary

An [LSP](https://langserver.org/) server for Cucumber

## Motivation and Context

Developer nowadays expect excellent language support in IDEs, and with the rise of VSCode, an LSP server is the way to provide this.

There is already a VSCode / LSP server at https://marketplace.visualstudio.com/items?itemName=alexkrechik.cucumberautocomplete that I considered improving, but I think it's important to be in control over the implementation.

## Types of changes

## Checklist:


language-server/javascript/src/makeCucumberInfo.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  const stderr: Buffer[] = []
+
+  return new Promise((resolve, reject) => {
+    cucumber.on('error', reject)
+    cucumber.on('close', (code, signal) => {
+      if (signal) {
+        reject(new Error(`Received signal ${signal}`))
+      }
+      // https://github.com/cucumber/cucumber-js/issues/1768
```
davidjgoss, line 24: Hopefully we can address this with the [new API](https://github.com/cucumber/cucumber-js/issues/1711) anyway (I'm finally starting to work on that now).

Interested in thoughts about - in addition to a `runCucumber` that does the full test run lifecycle - exposing a few more targeted functions that can help with things like this.

davidjgoss (commented):
Very nice @aslakhellesoy!

I don't have a ton of context around LSP and VSCode so not much specific feedback, but looks good to me.

One thought: given the issues around long build times etc, and the fact that these modules don't feed into a cucumber implementation, would they live better as their own repo?

aurelien-reeves:
> Very nice @aslakhellesoy!
> 
Yeah, really nice! 👏

Regarding the review, should we focus on something specific?

PR #1709, opened 2021-08-24
ESM modules

aslakhellesoy:
This PR is an attempt to switch to ESM modules. The motivation for doing this is than an increasing number of npm modules are ESM only, and we cannot use them unless our own modules are ESM.

In order to be able to import e.g. `@cucumber/messages` from `@cucumber/message-streams` (inter-module dependencies) at runtime (using ts-node so we don't need to pre-compile all ts code), we can no longer depend on [tsconfig-paths](https://github.com/dividab/tsconfig-paths).

Instead, I have written a custom ESM loader that extends the `ts-node/esm` loader and modifies code like `import from '@cucumber/messages'` to `import from '@cucumber/messages/typescript'`. This works as long as all our modules export a `./typescript` module in the package.json `exports` section.

There are still some issues to sort out, but I've got this working for the `@cucumber/messages` and `@cucumber/message-streams` modules.


davidjgoss:
I think it would be worth discussing the downstream implications of this. Once all the common blocks are ESM only, cucumber-js has to do the same, and that will have an impact on end users:

- Support code will need to be rewritten to use `import` instead of `require`, since they will no longer be able to `require('@cucumber/cucumber')`
- Use of JIT transpilers (ts-node, coffee, babel) may be broken to some degree

Somewhat related: https://github.com/cucumber/cucumber-js/pull/1649

(Not saying we shouldn't do it - library maintainers are basically forcing the issue anyway. But I think we need to communicate it will and make it not disruptive to users.)

aslakhellesoy:
I've seen a few npm modules that are published with both esm and commonjs exports. If we do that too, won't we be able to support both commonjs and esm?

davidjgoss:
> ...modules that are published with both esm and commonjs exports

The [strategies](https://nodejs.org/api/packages.html#packages_approach_1_use_an_es_module_wrapper) for doing this - having two compile targets, or providing an ESM facade around CommonJS code - both require that packages you _consume_ are still interoperable, which is the problem we're running into anyway.

aslakhellesoy:
Oh I see.

It seems to me this recent push to ESM is viral - everyone is going to have to do it, or be stuck on old modules. All it takes is for one npm module to be ESM-only, and then it will ripple through anything that depends on it...

aurelien-reeves:
I agree with @aslakhellesoy: ESM will be force-pushed really quickly. So I think we could consider moving to ESM only our-self.

Regarding the following PR however:
- There is a major changes in the APIs - actually how they are exposed - due to the new `exports` instruction in the `package.json`. I think this is a breaking change. Not by choice, not by design, but for our own convenience. Are we sure we want to do that?
- With your last commit, you hit an issue of the experimental ts-node ESM loader: the requirement for `.js` extension. To actually load `.ts` files. I think that could be ok for a small project considering that it should not remain like that in the future. For a big project like Cucumber, I am not fan of it

aurelien-reeves:
@aslakhellesoy it looks like #1743 and #1756 are promising

Do you want to keep this PR opened?

aslakhellesoy:
Nope :-)

PR #1728, opened 2021-09-06
[perl] build, test and release improvements

ehuelsmann:
## Summary

Main improvement is the release improvement where CHANGELOG.md wasn't included in Gherkin-21.0.0 because it's an unversioned file. On the Cucumber::Messages release, it was included because it was a versioned file until the post-release step removed it.

## Details

1. On the next Cucumber::Messages release, the CHANGELOG.md file would not be included without this change.
2. To improve understanding for occasional reviewers, remove the MANIFEST.SKIP file as a second source for what might (not) be included in a distribution archive.
3. Align the `clean` targets between Gherkin and Cucumber::Messages, where `clean` does not remove generated files (and `clobber` *does*)
4. Centralize PERL5LIB setting in anticipation of making Gherkin depend on Cucumber::Messages and depend on the version in the repository for CI builds

## Motivation and Context

Correct releases, Alignment of infrastructure between various Perl libraries inside a single repo and preparation to integrate Cucumber::Messages into Gherkin.

## How Has This Been Tested?

Rerunning `make distribution` and verifying the content of the distribution tarball.

## Types of changes

## Checklist:


messages/perl/dist.ini
```diff
@@ -29,8 +29,13 @@ repository.type   = git
 [PkgVersion]
 [Prereqs::FromCPANfile]
 [Git::GatherDir]
-exclude_filename=VERSION
 exclude_filename=default.mk
+exclude_filename=Makefile
+exclude_filename=VERSION
+
+[GatherFile]
+; explicitly add unversioned files
```
aurelien-reeves, line 37: CHANGELOG.md should be versioned 🤔
Isn't it?

gherkin/perl/dist.ini
```diff
@@ -27,8 +27,17 @@ repository.type   = git
 [PkgVersion]
 [Prereqs::FromCPANfile]
 [Git::GatherDir]
-exclude_filename=VERSION
+exclude_match=^helper-scripts/
+exclude_filename=gherkin.berp
+exclude_filename=gherkin-perl.razor
+exclude_filename=gherkin-languages.json
```
aurelien-reeves, line 33: Are we sure we want to excelude `gherkin-languages.json`? Perl implementation do not use it at runtime?

aurelien-reeves (commented):
A couple of question related to my un-knowledge of perl 😅

gherkin/perl/dist.ini
ehuelsmann: That's correct: the file only needs to exist at build time: https://github.com/cucumber/common/blob/main/gherkin/perl/helper-scripts/build_languages.pl generates the lib/Gherkin/Generated/Languages.pm file which is a compressed Perl representation of the JSON file's content. The idea is that it's much easier and more standard to deploy Perl files than "other" files.

messages/perl/dist.ini
ehuelsmann: It *was* versioned in the `messages/perl/` directory, but after the Cucumber::Messages 17.0.1 release, it was removed as a versioned file. The release procedure copies it into the module directory as an unversioned file. Same procedure is used for the `gherkin/perl/` directory. (Where Gherkin 21.0.0 doesn't have a CHANGELOG.md file in the distribution archive because it was unversioned at the time of the release.)

aurelien-reeves: Ok, got it, thanks

aurelien-reeves (approved):
As far as I can tell, it looks good ;)

ehuelsmann:
Thanks for the review! Merging.

PR #1740, opened 2021-09-08
[gherkin-perl] Add module documentation for public API

ehuelsmann:
## Summary

Add missing API documentation.

## Details

Now that the Gherkin/Perl modules are on target (i.e. based on Cucumber::Messages), we'd like people to actually be able to use these modules. Documentation is paramount to that goal.

## Motivation and Context

Why, if we don't want to document the modules, would we even go through the trouble publishing them?

## How Has This Been Tested?

The documentation itself has not; the tests have been run to verify no new regressions.

## Types of changes

## Checklist:


aurelien-reeves (approved):
Very nice! 🤩

We should consider documenting all implementations this way!

ehuelsmann:
Thanks for the review!

PR #1741, opened 2021-09-08
Change messages and Gherkin parser/pickle compiler to retain step keyword

ehuelsmann:
## Summary

Re #768: As discussed, for further discussion, implement a proposal for retention of step keywords Given/When/Then/But/And into the pickle.

Please note that the tests on this PR can't succeed until the test-data has been patched to include the new fields defined in the JSON schema. I'll take that task when the general direction has been decided to be acceptable/refined.

## Details

Considerations (summarized from the issue as well as implementation-wise):

 * Some implementations use keyword-specific dispatching of step functions; without the keyword in the pickle, these implementations can't migrate to Gherkin
 * The keywords `But` and `And` are resolved to the `Given`/`When`/`Then` keywords they inherit their context from, because:
   * Pickles contain execution plans, i.e. with variables resolved (readily available for execution)
   * Along the same lines, the `But` and `And` keywords have been resolved to eliminate variability in intepretation
 * The step dispatch environment doesn't have the parser available to determine the meaning of the keyword as mentioned in the parsed document; therefore the GherkinDocument mentions both the keyword from the parsed text (`keyword`) and the key as it is used in the translations (Given/Then/When/And/But).

## Motivation and Context

1. To allow implementations to move to the Gherkin parser
2. To retain information currently, literally, "lost in translation"
3. Strengthen separation of concerns (remove the need to mix parsing, execution and formatting stages)

## Screenshots

Not screenshots, but rendered output nonetheless.

(the first scenario of the) AST output from `gherkin --predictable-ids --no-source --no-pickles testdata/good/i18n_fr.feature`:

```json
{
  "gherkinDocument": {
    "comments": [],
    "feature": {
      "children": [
        {
          "scenario": {
            "description": "",
            "examples": [],
            "id": "3",
            "keyword": "Scénario",
            "location": {
              "column": 3,
              "line": 4
            },
            "name": "Support des caractères spéciaux",
            "steps": [
              {
                "id": "0",
                "keyword": "Soit ",
                "keywordType": "Given",                                        ******
                "location": {
                  "column": 5,
                  "line": 5
                },
                "text": "un exemple de scénario en français"
              },
              {
                "id": "1",
                "keyword": "Quand ",
                "keywordType": "When",                                        ******
                "location": {
                  "column": 5,
                  "line": 6
                },
                "text": "j'ai 1 gâteau"
              },
              {
                "id": "2",
                "keyword": "Alors ",
                "keywordType": "Then",                                       ******
                "location": {
                  "column": 5,
                  "line": 7
                },
                "text": "je suis heureux"
              }
            ],
            "tags": []
          }
        },
```
And the pickle output for said first scenario (`gherkin --predictable-ids --no-source --no-ast testdata/good/i18n_fr.feature`):

```json
{
  "pickle": {
    "astNodeIds": [
      "3"
    ],
    "id": "53",
    "language": "fr",
    "name": "Support des caractères spéciaux",
    "steps": [
      {
        "astNodeIds": [
          "0"
        ],
        "id": "50",
        "keyword": "Given",                                                     ******
        "text": "un exemple de scénario en français"
      },
      {
        "astNodeIds": [
          "1"
        ],
        "id": "51",
        "keyword": "When",                                                     ******
        "text": "j'ai 1 gâteau"
      },
      {
        "astNodeIds": [
          "2"
        ],
        "id": "52",
        "keyword": "Then",                                                     ******
        "text": "je suis heureux"
      }
    ],
    "tags": [],
    "uri": "testdata/good/i18n_fr.feature"
  }
}
```

See the lines with the `******` markers for the new output.

## How Has This Been Tested?

It's only lightly tested, because this is mostly for discussion purposes.

## Types of changes

## Checklist:

- [x] The change has been ported to .NET.

@aslakhellesoy indicated this PR should not be blocked on the `.NET` implementation, as it's not ported to JSONSchema yet (still running on the old protobuf schema).


gherkin/perl/lib/Gherkin/Pickles/Compiler.pm
```diff
@@ -68,9 +77,17 @@ sub _compile_scenario {
                 $class->_create_pickle_arguments(
                     $step,
                     $variables, $values );
+
+            $last_keyword =
+                $inheriting_keyword{$step->keyword_type} ?
+                $last_keyword : $step->keyword_type;
+
+            #TODO: Throw an exception here when $last_keyword is empty
```
aurelien-reeves, line 85: Why throw an exception?

If last_keyword is empty, I would just consider it as the non specified keyword `*`

aurelien-reeves:
It looks promising 👍 

Is the "non-keyword" `*` missing?
How would you manage the following:

```gherkin
Feature:
  Scenario:
    * first step
    * second step
```

gherkin/perl/lib/Gherkin/Pickles/Compiler.pm
ehuelsmann: `$last_keyword` is empty when there was no prior `Given`/`When`/`Then` keyword and the current keyword is `And` or `But`.

To be honest, I completely forgot about the non-specified keyword `*`. I think you're right that making `And` and `But` resolve to the non-specified keyword when there was no prior keyword specified, makes the most sense.

ehuelsmann:
Yes. The "non-keyword" is missing. If you look at

https://github.com/cucumber/common/pull/1741/files#diff-c07661832329db6cbfc36ea70c7dde480191c281e6e134531ef7b7638d144cfbR182

you'll see that the `*` keyword isn't being considered in the current Perl implementation of the Gherkin parser, which is why I overlooked it. There's no failure in the parser's test suite, which makes me wonder: is `*` still a supported keyword? Or is it there in some implementations for backward compatibility?

aurelien-reeves:
> 
> 
> Yes. The "non-keyword" is missing. If you look at
> 
> https://github.com/cucumber/common/pull/1741/files#diff-c07661832329db6cbfc36ea70c7dde480191c281e6e134531ef7b7638d144cfbR182
> 
> you'll see that the `*` keyword isn't being considered in the current Perl implementation of the Gherkin parser, which is why I overlooked it. There's no failure in the parser's test suite, which makes me wonder: is `*` still a supported keyword? Or is it there in some implementations for backward compatibility?

Good question for our community meeting 😁

ehuelsmann:
> > is `*` still a supported keyword? Or is it there in some implementations for backward compatibility?
> Good question for our community meeting grin

So, in preparation of that meeting, I took a stab at the Ruby parser, trying to find whether it supports the `*` non-keyword. I found:

https://github.com/cucumber/common/blob/main/gherkin/ruby/lib/gherkin/token_matcher.rb#L120-L124

which suggests that it also doesn't. So, definitely something to discuss!

brasmusson:
My take on the "*" keyword is that when using it, the step should be allowed to match step definitions defined with all of "Given", "When", and "Then".

From the original discussions of "unambiguous keywords", I think I remember the view that defining both step definitions for "Given <something>" and "When <something>" should (still?) be an error, so the keywords in the pickles should rather be for verifying the keyword of the step definition, than for selecting one of several candidates of step definitions.

brasmusson:
As long as the "* " keyword is defined in the [gherkin-languages.json](https://github.com/cucumber/common/blob/main/gherkin/gherkin-languages.json) is is supported, isn't it?

mpkorstanje:
> The keywords But and And are resolved to the Given/When/Then keywords 

I would suggest deferring this to a the cucumber implementation. It's non trivial, especially with `* ` keyword is used. The pickle may also be used in representations in which case it would be confusing to see it replaced.

ehuelsmann:
This is how `* ` keyword is supported: https://github.com/cucumber/common/blob/main/gherkin/gherkin-languages.json#L4 shows that `* ` is a translation/dialect for the `And` keyword.

sebrose:
- It looks like '*' is supported in some Cucumber Rubys (I tried it in 4.0.0)
- By converting the actual keyword used in the scenario into G/W/T in the pickle, we lose the ability to reconstitute the scenario as written
- There's nothing to prevent a scenario starting with a conjunction (And / But)

aurelien-reeves:
> 
> 
> This is how `* ` keyword is supported: https://github.com/cucumber/common/blob/main/gherkin/gherkin-languages.json#L4 shows that `* ` is a translation/dialect for the `And` keyword.

Yeah
Actually it is a translation for all given/when/then/and/but keyword.
I guess it fallback as "Given" if needed?

gasparnagy:
On the other hand, we realized with @sebrose that treating `*` just as a simple synonym of `And` is not perfect... e.g. a scenario with only `*` steps might be valid on the other hand having a scenario with only `And`s does not look good...

So if we do this change it would be good if we could semantically detach the `*` from the `And`s...

davidjgoss:
> By converting the actual keyword used in the scenario into G/W/T in the pickle, we lose the ability to reconstitute the scenario as written

I think the intent was to have this as well as the verbatim keyword, not instead?

brasmusson:
The "* " is a valid keyword for Given, When, Then, And, But (see [gherkin-languages.json](https://github.com/cucumber/common/blob/main/gherkin/gherkin-languages.json)). So any parser that passes the test supports "* " (see [this](https://github.com/cucumber/common/blob/main/gherkin/testdata/good/incomplete_background_1.feature.ast.ndjson) ast.ndjson). Currently its not a problem that "* " is a valid keyword for all types, but when converting to G/W/T then the question of how to convert "* " occurs. My suggestion in that case, it should be converted to the "*" type (which would be valid for all step definitions).

ehuelsmann:
@gasparnagy writes:

> On the other hand, we realized with @sebrose that treating `*` just as a simple synonym of `And` is not perfect... e.g. a scenario with only `*` steps might be valid on the other hand having a scenario with only `And`s does not look good...

I'd agree that a scenario of pure conjunctions (And / But) doesn't look good. However, if I'm understanding correctly the way Cucumber Ruby/Java work now, it's still allowed. To that extent, I'd say that the argument @mpkorstanje and @mattwynne (?) made to leave the intepretation of the keywords to the actual executor and not hard-code it in the Pickle compiler, does allow the executor to object to scenarios like these (or simply assume "* ").

> So if we do this change it would be good if we could semantically detach the `*` from the `And`s...

I think that if I change the PR to have the same fields `keyword` and `keywordType` in `PickleStep` the way it's now done in the `Step`, with the addition of the new keywordType `*` (in addition to the keywordTypes `Given/When/Then//And/But`), the code will be able to handle all comments made so far.

brasmusson:
> We can map *, And and But to Ambiguous

@mattwynne I disagree with that `*` is mapped the same way `And` and `But`. The current meaning of `And` and `But` is "the same as the previous step". But that is currently not the meaning of `*`, as it is a valid keyword for `Given`, `When`, `Then`, `And` and `But`.

gasparnagy:
I like the abstraction of phase suggested my @mattwynne. What about these?

* `Context` -- like `Given`
* `Action` -- like `When`
* `Outcome` -- like `Then`
* `Conjunction` or `Inherited` -- like `And` and `But`
* `Step` -- for the `*`

mattwynne:
> But that is currently not the meaning of *, as it is a valid keyword for Given, When, Then, And and But.

Good point. In fact we might find (now or in the future) that there are other strings mapped in the gherkin language spec that could be mapped back to more than one keyword.

ehuelsmann:
> > But that is currently not the meaning of *, as it is a valid keyword for Given, When, Then, And and But.
> 
> Good point. In fact we might find (now or in the future) that there are other strings mapped in the gherkin language spec that could be mapped back to more than one keyword.

Assuming that more of these exist *today*, should all of these be mapped to "General" in that case? Doing so will at least give us the desired behaviour for `* ` without special-casing.

ehuelsmann:
Based on the comments above, I've updated the PR for further discussion. The output of the AST now looks like:

```json
{
  "gherkinDocument": {
    "comments": [],
    "feature": {
      "children": [
        {
          "scenario": {
            "description": "",
            "examples": [],
            "id": "3",
            "keyword": "Scénario",
            "location": {
              "column": 3,
              "line": 4
            },
            "name": "Support des caractères spéciaux",
            "steps": [
              {
                "id": "0",
                "keyword": "Soit ",
                "keywordType": "Context",
                "location": {
                  "column": 5,
                  "line": 5
                },
                "text": "un exemple de scénario en français"
              },
              {
                "id": "1",
                "keyword": "Quand ",
                "keywordType": "Action",
                "location": {
                  "column": 5,
                  "line": 6
                },
                "text": "j'ai 1 gâteau"
              },
              {
                "id": "2",
                "keyword": "Alors ",
                "keywordType": "Outcome",
                "location": {
                  "column": 5,
                  "line": 7
                },
                "text": "je suis heureux"
              }
            ],
            "tags": []
          }
        },
```

And the pickle output:

```json

  "pickle": {
    "astNodeIds": [
      "3"
    ],
    "id": "53",
    "language": "fr",
    "name": "Support des caractères spéciaux",
    "steps": [
      {
        "astNodeIds": [
          "0"
        ],
        "id": "50",
        "keyword": "Soit ",
        "keywordType": "Context",
        "text": "un exemple de scénario en français"
      },
      {
        "astNodeIds": [
          "1"
        ],
        "id": "51",
        "keyword": "Quand ",
        "keywordType": "Action",
        "text": "j'ai 1 gâteau"
      },
      {
        "astNodeIds": [
          "2"
        ],
        "id": "52",
        "keyword": "Alors ",
        "keywordType": "Outcome",
        "text": "je suis heureux"
      }
    ],
    "tags": [],
    "uri": "testdata/good/i18n_fr.feature"
  }
}
```

Note that the pickle now has the `keywordType` field as well as the `keyword` field with the same values as present in the `Step` in the GherkinDocument. The code currently specializes `*`, but that's work in progress: I think that all keywords that a dialect associates with more than one step keyword (given/then/when/and/but), should have the "General" keywordType associated, based on the discussion above. That will be part of a further update to this PR.

ehuelsmann:
With this latest commit, we don't need to restructucture `gherkin-languages.json`. It will extract all translations ("dialect words") that are associated with more than one step keyword and declare that translation to then be in the "General" group. Because "* " has been associated with *all* keywords, it automatically goes into the "General" type without special-casing it in the code.

luke-hill:
@gasparnagy Like yours. My only suggestion is "Any" for "*"

ehuelsmann:
> @gasparnagy Like yours. My only suggestion is "Any" for "*"

@luke-hill , given that [the comment above](https://github.com/cucumber/common/pull/1741#issuecomment-916697337), I'd say that "Unspecified" may be a good value, because it won't be just mapped to by "* ", but all keywords having more than one association with Given/Then/When/But/And. ("Unspecified" is better than "Unknown", because actually the keyword type *is* known, but it's not a strict mapping to a single type).

gasparnagy:
> I'd say that "Unspecified" may be a good value, because it won't be just mapped to by "* ", but all keywords having more than one association with Given/Then/When/But/And. 

I think what we want to say here is that these are neutral to whether they belong to the context, action or outcome. So it _is_ specified and we say that it is neutral. If in the future there will be a new keyword that works as Given and When, but not as Then, we would anyway need to create a new category for that.

I think the point is that the processing logic should be able to make decisions based on the `keyword_type` without checking the concrete keyword. (So we don't want to have the `* ` constant in the code.) 

From this perspective, `Any` or `General` might be better IMHO.

mattwynne:
What about having `gherkinKeywords` as an array, and list all the "root" keywords that the string used in the parsed document maps back to? Maybe a bit too low level?

gasparnagy:
@mattwynne I don't understand your proposal. Could you please give an example?

messages/jsonschema/GherkinDocument.json
```diff
@@ -347,8 +348,20 @@
           "description": "The location of the steps' `keyword`"
         },
         "keyword": {
+          "description": "The actual keyword as it appeared in the source.",
           "type": "string"
         },
+        "keywordType": {
```
aslakhellesoy, line 354: I like this

messages/jsonschema/GherkinDocument.json
```diff
@@ -347,8 +348,20 @@
...
+          "description": "The test phase signalled by the keyword: Contex definition (Given), Action performance (When), Outcome assertion (Then). Other keywords signal Continuation (And and But) from a prior keyword. Lastly, the `*` 'keyword` isn't associated with a specific phase; hence the association to 'General'.",
+          "type": "string",
+          "enum": [
+            "Context",
+            "Action",
+            "Outcome",
+            "Conjunction",
+            "General"
```
aslakhellesoy, line 362: I would prefer `Uknown` here to emphasise that we don't know the semantic keyword type.

messages/jsonschema/Pickle.json
```diff
@@ -25,6 +25,8 @@
       "required": [
         "astNodeIds",
         "id",
+        "keyword",
```
aslakhellesoy, line 28: What do we need the `keyword` for? Isn't the `keywordType` sufficient?

messages/jsonschema/Pickle.json
```diff
@@ -25,6 +25,8 @@
...
+        "keywordType",
```
aslakhellesoy, line 29: If we remove the `keyword` (see comment above), perhaps a better name for this field would be simply `type`? It describes the step just as much as it describes the keyword.

Example: This is a `Context` step.

messages/jsonschema/Pickle.json
```diff
@@ -42,6 +44,21 @@
...
           "type": "string"
         },
+        "keyword": {
+          "description": "The keyword as it appears in the feature file",
+          "type": "string"
+        },
+        "keywordType": {
+          "description": "The context in which the step was specified: context (Given), action (When) or outcome (Then).\n\nNote that the keywords `But` and `And` inherit their meaning from prior steps and the `*` 'keyword' doesn't have specific meaning (hence General)",
+          "type": "string",
+          "enum": [
```
aslakhellesoy, line 54: I think the following enum values here are sufficient:

* `Context`
* `Action`
* `Outcome`
* `Unknown`

At the pickle level, it should make no difference wherher an outcome step had the keyword `Then`, `And` or `But` in the source. The pickle compiler should translate `And` and `But` to one of `Context`, `Action`, `Outcome` (or `Unknown` in the unlikely event that the preceding source keyword was a `* `.

Not having `Conjunction` at the pickle step level also makes interpretation easier, since the interpretation can be encapsulated inside the compiler so that pickle consumers don't need to do it.

aslakhellesoy (changes requested):

messages/jsonschema/GherkinDocument.json
ehuelsmann: Here I think that `Unknown` is actually better than `Neutral` because it conveys that information isn't avaliable: there's no explicit mapping to exactly one of the enumerated types. So, if I get to point the bikeshed, it's `Unknown`.

messages/jsonschema/Pickle.json
ehuelsmann: When the pickle is focused on execution, I think you have a strong point. I also started out like that (although the names of the keywords were different). However, @mpkorstanje and @sebrose(?) suggested in [prior discussion](https://github.com/cucumber/common/pull/1741#issuecomment-916129760) that by compiling conjunctions into the keywords they refer to, we loose information. Between the Pickle and the GherkinDocument, we don't, their point is that the scenario can't be rendered from the Pickle alone.

ehuelsmann:
Adjustments I'll make to this PR:

* Change the enum in the Pickle as per your last comment
* Add translation of But/And to the preceeding keyword in the pickle compiler
* Rename the "keywordType" field in the Pickle to `type`

@aslakhellesoy does that encompass all change you intended to ask for?

ehuelsmann:
The latest push rebases on top of current `origin/main` and adds the changes requested by @aslakhellesoy.

messages/jsonschema/GherkinDocument.json
gasparnagy: I think for `* ` there is a kind of consensus, that it should be neutral to the Context/Action/Outcome phases. The `Unknown` is hiding this fact, so from Aslak's choices my vote is for `Netural`, but I even better like `Any` suggested by @luke-hill.

gasparnagy: We can still have an additional `Unknown` as well, of course, but I would not use it for the `* `.

messages/jsonschema/Pickle.json
aslakhellesoy: @sebrose said:

> By converting the actual keyword used in the scenario into G/W/T in the pickle, we lose the ability to reconstitute the scenario as written.

We have already lost the ability to reconstitute the scenario as written from a pickle, so this is a moot point (pickles inline backgrounds, examples tables etc). We never do this either. If we want to reconstitute the source, we'll do it from the `GherkinDocument` AST, not the pickle.

If we *really* want to obtain the original source keyword, we can do this by looking up the AST step from the [pickle step](https://github.com/cucumber/common/blob/main/messages/messages.md#picklestep)'s `astNodeIds`.

I think we should leave it out for now, and only add it later if we have a specific need for it.

messages/jsonschema/Pickle.json
aslakhellesoy: Pickles are only meant to be used for execution - not reporting. By delegating the responsibility to the pickle consumer to figure out whether a step is context/action/outcome, we're opening up for multiple inconsistent interpretations for no obvious benefit.

We do lose information, but it's information we don't need for execution. The purpose of this PR (as I see it) is simply to be able to prevent the use of a `Given` step definition from a `When` step etc, and I think we should make this as easy as possible to implement on the Cucumber side.

mattwynne:
> @mattwynne I don't understand your proposal. Could you please give an example?

In `gherkin-languages.json` in the entry for the `en` language, the translations for the `"given"` keyword are `"Given " ` and `"* "`. For the `"when"` keyword they are `"When "` and `"* "`, etc.

I was suggesting, though I think the conversation has moved on, that we simply model this mapping in the `keywordType` field, so if the user puts a `*` as their step keyword, we write the pickle's `keywordType` as an array of all those map keys in the language object, i.e.

````json
{ 
  ...
  "keyword": "*",
  "keywordTypes": [ "and", "but", "given", "when", "then" ]
  ...
}
```

In most cases, there will be only one entry in the `keywordTypes` array, but for a keyword that has an ambiguous / general mapping there will be more than one entry.

messages/jsonschema/Pickle.json
mpkorstanje: > Pickles are only meant to be used for execution - not reporting. By delegating the responsibility to the pickle consumer to figure out whether a step is context/action/outcome, we're opening up for multiple inconsistent interpretations for no obvious benefit.

Fundamentally this isn't true. The feature format is not an accurate representation of the what is executed.

When reporting on what will be excuted a reasonable mapping to the feature file is needed. This includes the keyword as it occurs in the feature file.

ehuelsmann: @mpkorstanje Isn't it the case that @aslakhellesoy tries to separate the "reporting side" (including the "this is what I'm executing" output) from the actual execution ("please execute this pickle step with this keyword type, this argument and these parameters")?

I do see why that then later becomes a challenge to correctly report to the user *if* you're doing that based on the pickle content. I'm thinking that @aslakhellesoy would want the progress report to be based on the "testCase" and the AST, so that you don't need the pickle. (The AST has the keyword, keyword type, argument and parameters, all of which you need to do the reporting; the testCase has the test outcome, which would help render in the correct color?)

messages/jsonschema/GherkinDocument.json
ehuelsmann: @aslakhellesoy I'm continuing my work on this change. I've now picked up your preference to associate `* ` with `Unknown`. Let me know if that's not going to fly.

messages/jsonschema/Pickle.json
ehuelsmann: `keyword` is no longer in this change.

messages/jsonschema/Pickle.json
ehuelsmann: Yes. `keyword` is no longer in this change.

aurelien-reeves:
The CI is failing because the `Step.keywordType` and `PickleStep.type` are defined as "required" in the json schema.
Yet the gherkin implementation has not been updated accordingly.

Do we need those properties to be mandatory? Doing so will make the implementation of the keyword retention more difficult

aurelien-reeves:
> The CI is failing because the `Step.keywordType` and `PickleStep.type` are defined as "required" in the json schema. Yet the gherkin implementation has not been updated accordingly.
> 
> Do we need those properties to be mandatory? Doing so will make the implementation of the keyword retention more difficult

Just tested locally: removing the keyword type from the "required" attributes make the build go further.
It would be more easier to move on incrementally with this feature without those new fields to be required

aslakhellesoy:
When we add a new field, it should *never* be marked as required. This is to ensure *forward* compatibility: Consumers using a new version of the lib can still process messages produced by an old version of the lib.

ehuelsmann:
> When we add a new field, it should _never_ be marked as required. This is to ensure _forward_ compatibility: Consumers using a new version of the lib can still process messages produced by an old version of the lib.

Maybe this is a bridge too perfect, but: I'd expect schemas to adhere to semantic versioning similar to the way libraries do: new APIs/fields lead to new version numbers, which, when not backward compatible, leads to a new version number. While schemas and input parsers need to be forward compatible by ignoring any unknown data that's provided but not recognized, the software to parse the input data could be made backward compatible by recognizing the schema version of the data provided and e.g. treating `keywordType` as optional or even ignore it - thereby supporting (mimicing) older versions.

Having said that, I'll agree that it's probably an approach that 's too complex to maintain long-term for the amount resources available to the project. In which case, yes, the best solution is to mark these fields as optional.

There is a bit of a problem that I'm left with after that conclusion though: for e.g. the Perl implementation of Cucumber that I maintain, I'll *need* the `keywordType` and corresponding `type` fields (on the step and picklestep respectively) to be provided. What declaration/level of conformance would an implementation require to be guaranteed that type of input?

messages/jsonschema/GherkinDocument.json
```diff
@@ -338,6 +338,7 @@
       "required": [
         "location",
         "keyword",
+        "keywordType",
```
ciaranmcnulty, line 341: In terms of interoperability, I'm wary of making this required and breaking backwards compatibility with older streams (e.g. an older version of cucumber reporting test results to a newer version of a formatter).

Is it strictly necessary, or is there a path where we can introduce it as optional?

ciaranmcnulty: I appreciate this shifts some complexity into the consuming cucumber implementations, but they're already handling streams without these extra fields in, so can presumably fall back to existing behaviour

aurelien-reeves: This is actually discussed :D

ciaranmcnulty: Ah can you point me to the right place?

ciaranmcnulty: Ah indeed @aslakhellesoy made the point

ciaranmcnulty:
@ehuelsmann There's a slight semver complexity in both Gherkin and Messages, in that the major version will get bumped when API changes happen in the various implementations, without it reflecting a backwards compatibility break in the serialisation format

For instance, Gherkin is very broadly backwards compatible since 6.0 but the libraries have changed methods, etc, so the major version increases.

I think if you want to require the new fields for cucumber-perl:

 * messages _supports_ the new fields from a particular version X
 * gherkin to _always populates_ the fields from version Y
 * cucumber-perl requires messages >=X gherkin >= Y
 * cucumber-perl checks the fields are populated and shows an appropriate error message if not (which won't really happen, it's only if someone's feeding it gherkin from a different source than the correct gherkin-perl version)

aurelien-reeves:
> Maybe this is a bridge too perfect, but: I'd expect schemas to adhere to semantic versioning similar to the way libraries do: new APIs/fields lead to new version numbers, which, when not backward compatible, leads to a new version number. While schemas and input parsers need to be forward compatible by ignoring any unknown data that's provided but not recognized, the software to parse the input data could be made backward compatible by recognizing the schema version of the data provided and e.g. treating `keywordType` as optional or even ignore it - thereby supporting (mimicing) older versions.
> 
> Having said that, I'll agree that it's probably an approach that 's too complex to maintain long-term for the amount resources available to the project. In which case, yes, the best solution is to mark these fields as optional.
> 

> There is a bit of a problem that I'm left with after that conclusion though: for e.g. the Perl implementation of Cucumber that I maintain, I'll _need_ the `keywordType` and corresponding `type` fields (on the step and picklestep respectively) to be provided. What declaration/level of conformance would an implementation require to be guaranteed that type of input?

Actually, the generated code issue from the schema will set fields that are not marked as "required" with default values. Thus the field is always there and have a correct value next in the stack

It is then up to gherkin to manage exception if the data is missing in the source being parsed, which is already the case I guess. The "required" attribute in the schema does not mean a "required" field in the gherkin syntax.

aslakhellesoy:
Let's assume I have a server that always uses the latest version of messages. This server parses messages sent to it, and these messages can be produced by many different versions of messages (different cucumber versions).

If a message field is required in the new version, the code in the server will not check whether it has a value or not. Checking if a field is null is *only* done for fields that are not required.

So the server might blow up, because it receives an old message that doesn't set the (newly) required field. This is what I mean by forward compatibility.

This is explained in more detail [here](https://stackoverflow.com/questions/31801257/why-required-and-optional-is-removed-in-protocol-buffers-3)

gherkin/java/src/main/java/io/cucumber/gherkin/Token.java
```diff
@@ -10,6 +12,7 @@ class Token {
     public List<GherkinLineSpan> matchedItems;
     public int matchedIndent;
     public GherkinDialect matchedGherkinDialect;
+    public StepKeywordType keywordType;
```
ciaranmcnulty, line 15: I think this would be better named `matchedKeywordType` as all the other fields that 'only exist after a match' are named that way. Also note https://github.com/cucumber/common/issues/1941

gherkin/java/src/main/java/io/cucumber/gherkin/GherkinDialect.java
```diff
@@ -60,6 +64,18 @@ public List<String> getStepKeywords() {
         return result;
     }
 
+    public Map<StepKeywordType, List<String>> getStepKeywordsByType() {
```
ciaranmcnulty, line 67: I think it would be most useful to return `Map<String, StepKeywordType>` here, as it would make the lookup in TokenMatcher simpler

Or possibly it would make more sense to have `public StepKeywordType getKeywordType(String keyword)` here?

gherkin/java/src/main/java/io/cucumber/gherkin/PickleCompiler.java
```diff
@@ -97,10 +101,19 @@ private void compileRule(List<Pickle> pickles, Rule rule, List<Tag> parentTags,
...
-            steps.addAll(pickleSteps(backgroundSteps));
+        if (!scenario.getSteps().isEmpty()) {
+            List<Step> allSteps = new ArrayList<>();
+            allSteps.addAll(backgroundSteps);
+            allSteps.addAll(scenario.getSteps());
 
-        steps.addAll(pickleSteps(scenario.getSteps()));
+            StepKeywordType lastKeywordType = StepKeywordType.UNKNOWN;
+            for (Step step : allSteps) {
+                if (step.getKeywordType() != StepKeywordType.CONJUNCTION)
```
ciaranmcnulty, line 111: This logic is repeated 3 times in the class, which makes me think it may have a more natural home. Is the pickle the right place to resolve this, or is it the job of the consuming cucumber (which would do this in one place)?

gherkin/java/src/main/java/io/cucumber/gherkin/Token.java
ciaranmcnulty: this would make sense to be moved up next to the matchedKeyword property too

I do wonder if Keyword should be a small object at this stage (with the String+type), it might make things neater

gherkin/java/src/main/java/io/cucumber/gherkin/TokenMatcher.java
```diff
@@ -166,29 +173,51 @@ private boolean match_DocStringSeparator(Token token, String separator, boolean
...
+        Map<StepKeywordType, List<String>> keywords = currentDialect.getStepKeywordsByType();
+
+        String keyword = null;
+        StepKeywordType keywordType = StepKeywordType.UNKNOWN;
+        for (Entry<StepKeywordType,List<String>> entry : keywords.entrySet()) {
+            List<String> translations = entry.getValue();
+
+            if (keyword != null) { // we have a keyword already
+                if (translations.contains(keyword)) {
+                    keywordType = StepKeywordType.UNKNOWN;
```
mpkorstanje, line 193: `keywordType` is already set in the `else` branch. Why would it have to be set?

gherkin/java/src/main/java/io/cucumber/gherkin/TokenMatcher.java
```diff
@@ -166,29 +173,51 @@ private boolean match_DocStringSeparator(Token token, String separator, boolean
...
-        for (String keyword : keywords) {
-            if (token.line.startsWith(keyword)) {
-                String stepText = token.line.getRestTrimmed(keyword.length());
-                setTokenMatched(token, TokenType.StepLine, stepText, keyword, null, null);
-                return true;
+        Map<StepKeywordType, List<String>> keywords = currentDialect.getStepKeywordsByType();
+
+        String keyword = null;
+        StepKeywordType keywordType = StepKeywordType.UNKNOWN;
+        for (Entry<StepKeywordType,List<String>> entry : keywords.entrySet()) {
```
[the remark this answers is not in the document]
ehuelsmann: The code generates the following mapping:

| step keyword type | prior keyword type | keyword appears in given/when/then translations | output (keywordType)
|--------|-----------|--|--|
| CONJUNCTION | none (initial) | once | UNKNOWN (using `lastKeywordType`)
| CONJUNCTION | `lastKeywordType` | once | `lastKeywordType`
| CONTEXT | (not applicable) | once | CONTEXT
| ACTION  | (not applicable) | once | ACTION
| OUTCOME  | (not applicable) | once | OUTCOME
| CONTEX/ACTION/OUTCOME/CONJUNCTION | (not applicable) | multiple (e.g. given+when) | _**UNKNOWN**_

The last row is the one that relates to your question below. Because there are multiple mappings of a single keyword translation, we cannot determine the exact relation _translated keyword_ -> _Given/When/Then/But/And_; hence the conclusion `unknown`.

gherkin/java/src/main/java/io/cucumber/gherkin/GherkinDialect.java
ehuelsmann: Thinking about your second suggestion, I think that'll allow me to restore the original iteration over all keywords as it was originally coded, retrieving the keyword type through this function call. However, the function itself would need to do roughly the same as the inner loop right now (or it can do this in advance when selecting the dialect): It'll need to determine whether a keyword translation associates with more than one keyword (Given/When/Then/But/And).

gherkin/java/src/main/java/io/cucumber/gherkin/PickleCompiler.java
ehuelsmann: That's correct. I've thought about abstracting this out into a single place, but that's not so easy because of the fact that the "lastKeywordType" state needs to be maintained during the `compileScenario` and `compileScenarioOutline` functions.
I don't think this belongs in the consumers - even though they would indeed do this in one place - because the pickles are supposed to be directly executable, that is, it has all the necessary substitutions applied. While writing this code, I imagined this to be one of the substitutions the consumer needs to be able to rely on to have taken place.

gherkin/java/src/main/java/io/cucumber/gherkin/Token.java
ehuelsmann: That's definitely an option (making Keyword a small object); I decided not to do that, because not all keywords are step keywords (Scenario/Rule/Example/...), but foremost because I was trying to keep the API changes (or the changes in general) as small as possible.

gherkin/java/src/main/java/io/cucumber/gherkin/TokenMatcher.java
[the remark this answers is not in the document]
ehuelsmann: My last push seriously simplifies the loop. I hope that's an improvement.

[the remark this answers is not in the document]
mpkorstanje: I feel I have to attach a debugger to understand what is going on. But not having done so yet, my comments:

I'd expect to see to see the resolution of multiple mappings reflected in the code rather then implicitly brushed over. For example:

```
List<Entry<StepKeywordType, String> candidates = keywords.entrySet()
   .filter( ... some condition .... )
   .map( ... some transformation .... )   
   .collect(....);
   
if (candidates.size() != 1)
   return unknown
else
  return candidates.get(0).....
```

Though this also makes me wonder why a `Set<StepKeywordType,List<String>>` is used rather then a `Set<String,List<StepKeywordType>>`. Then the resolution of the keyword type is a lookup rather then an algorithm.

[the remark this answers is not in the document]
ehuelsmann: Heh. It seems our comments just crossed each other. The new code no longer uses the problematic pattern in this place of the code. I think that makes this bit more readable. A similar pattern is now in `GherkinDialect` though, which precomputes the `Map` that @ciaranmcnulty suggested to use in the `match_StepLine` function. There, I've added a comment a moment ago to indicate why the override on the type is happening.

ehuelsmann:
Had to force-push after rebasing on `main`, resolving conflicts due to the Java JSON changes.

gherkin/java/src/main/java/io/cucumber/gherkin/GherkinDialect.java
ehuelsmann: Implemented something along these lines and it simplifies `match_StepLine()` indeed.

gherkin/java/src/main/java/io/cucumber/gherkin/TokenMatcher.java
ehuelsmann: (discussion continued in the other comment)

gherkin/java/src/main/java/io/cucumber/gherkin/Token.java
ehuelsmann: Yep. Did the rename as you suggested.

gherkin/java/src/main/java/io/cucumber/gherkin/PickleCompiler.java
ciaranmcnulty: Maybe this should be something that's done in `compile` after `compileFeature` but before the pickles are returned?

gherkin/java/src/main/java/io/cucumber/gherkin/GherkinDialect.java
```diff
@@ -75,6 +105,10 @@ public List<String> getStepKeywords() {
         return stepKeywords;
     }
 
+    public Map<String, StepKeywordType> getStepKeywordsTypes() {
```
ciaranmcnulty, line 108: Because the only caller immediate calls `.get(keyword)` I feel this should be something more like:

```
public StepKeywordType getStepKeywordType(String keyword)
```

I'm a little unsure about the UNKNOWN result - it might be better to return a list of possibilities? It depends how you want to consume it in cucumber

messages/jsonschema/GherkinDocument.json
```diff
@@ -347,8 +348,20 @@
           "description": "The location of the steps' `keyword`"
         },
         "keyword": {
+          "description": "The actual keyword as it appeared in the source.",
           "type": "string"
         },
+        "keywordType": {
+          "description": "The test phase signalled by the keyword: Contex definition (Given), Action performance (When), Outcome assertion (Then). Other keywords signal Continuation (And and But) from a prior keyword. Please note that all translations which a dialect maps to multiple keywords (`*` is in this category for all dialects), map to 'Unknown'.",
+          "type": "string",
+          "enum": [
```
ciaranmcnulty, line 357: other enums have the 'default' value in the first position. This would make more sense with Unknown first IMO, as we could use that rather than null as a sane default

messages/jsonschema/Pickle.json
```diff
@@ -43,6 +44,16 @@
...
           "type": "string"
         },
+        "type": {
+          "description": "The context in which the step was specified: context (Given), action (When) or outcome (Then).\n\nNote that the keywords `But` and `And` inherit their meaning from prior steps and the `*` 'keyword' doesn't have specific meaning (hence Unknown)",
+          "type": "string",
+          "enum": [
+            "Context",
+            "Action",
+            "Outcome",
+            "Unknown"
```
ciaranmcnulty, line 54: See comment about unknown being first to make It 'default'

gherkin/java/src/main/java/io/cucumber/gherkin/PickleCompiler.java
```diff
@@ -181,7 +203,17 @@ private PickleDocString pickleDocString(DocString docString, List<TableCell> var
         );
     }
 
-    private PickleStep pickleStep(Step step, List<TableCell> variableCells, TableRow valuesRow) {
+
+    private static final Map<StepKeywordType, PickleStepType> pickleStepTypeFromKeywordType = new HashMap<>();
+
+    static {
```
ciaranmcnulty, line 209: Rather than here, look at how ew handle two other 'similar' enums that we need to cast between. It makes sense to me to handle the same way 

https://github.com/cucumber/common/blob/main/gherkin/java/src/main/java/io/cucumber/gherkin/Parser.java#L75

messages/jsonschema/GherkinDocument.json
ehuelsmann: Changed. (And changed the typo on `Contex` above too as you noted in the Slack channel.)

messages/jsonschema/Pickle.json
ehuelsmann: Changed too.

gherkin/java/src/main/java/io/cucumber/gherkin/GherkinDialect.java
ehuelsmann: Yes. I'm happy to change the code to return all the possibilitiies; I was looking at the use of the Dialect in the code base. It's really a very low in the stack of abstractions and a building block nobody saw fit to use for anything but i18n-ing the token matcher. which is why I coded the output to the currently-known use-case. And since this call will happen on every matched keyword, I was thinking it's rather inefficient to conclude every time that a keyword maps to multiple types, only to conclude *then* that the keyword type must be set to UNKNOWN.

ehuelsmann: This is now implemented as a list of reverse mappings.

gherkin/java/src/main/java/io/cucumber/gherkin/PickleCompiler.java
ehuelsmann: I verified that works. However, it hard-codes the ordering assumption between `keywordType` and `type` enum values, without that being explicit in the code base elsewhere. The way it's coded now, that ordering isn't required, reducing the likelihood of breaking the code base by changing the enum order in the schema.

ehuelsmann: Concluding for now that I'd rather keep it as it is for now, also because the cast is part of the enum declaration, which for `PickleStepType` is the result of code generation -- in contrast to the code you reference above.

gherkin/java/src/main/java/io/cucumber/gherkin/PickleCompiler.java
ehuelsmann: I looked at the options to do that. Although the pattern does repeat itself three times at an abstract level; on an implementation level, it's really two of one type and one of another; for now, I'll leave it as it is, because the deduplication is rather minimal and a repeated pattern will remain (2&1->1&1).

gherkin/java/src/main/java/io/cucumber/gherkin/TokenMatcher.java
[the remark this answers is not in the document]
ehuelsmann: This is now implemented as you suggested.

ehuelsmann:
Forced pushed to base this on the `messages-retain-keyword` branch, which factors out of this branch, the change solely to `messages`.

aslakhellesoy:
Updating the go part requires modifying [dialects_builtin.go.jq](https://github.com/cucumber/common/blob/main/gherkin/go/dialects_builtin.go.jq) - which is *really* hard to understand. I'm trying to grok it, but it might take some time...

aslakhellesoy:
I've figured out how to update the `jq` code generator, but *only* if we can make a step keyword correspond to a *single* keyword type. The current implementation assumes a step keyword can correspond to *multiple* keyword types.

This is because the following languages uses some keywords in multiple places:

* `en-old` (Old English) uses `Tha`, `Þa` and `Ða` for both `given` and `when`
* `ne` (Nepalese) uses `अनी` for both `then` and `and`
* `uz` (Uzbek) uses `Агар` for both `given` and `when`

Without being a native speaker of any of them, I think it might be a mistake to have this duplication. In order to move this forward I propose that we modify the 3 languages in question to not have duplicated keywords:

* `en-old` `given` becomes `þæt`, `þæs`, `þæm` ([Old English Translator](https://www.oldenglishtranslator.co.uk/))
* `ne` `and` becomes `अनि` ([Google Translate](https://translate.google.com/?sl=en&tl=ne&text=and&op=translate))
* `uz` `and` becomes `Va`, `Esa`, `Ham` ([Google Translate](https://translate.google.com/?sl=en&tl=uz&text=and&op=translate))

It's of course possible that this is wrong. In that case, the community will let us know. These languages will probably not be used a lot, so hopefully it will not affect many users. In fact, we're not *removing* any keywords, so everything will probably just work as before.

ehuelsmann:
@aslakhellesoy thanks for looking into that. If we agree that the mapping needs to be unique for all keywords other than `* ` (as you're proposing above), then the code on the other implementations can be (a bit) simpler too. So, I'm definitely :+1: on that. Maybe we should add a check to the test-suite to enforce it?

aslakhellesoy:
> Maybe we should add a check to the test-suite to enforce it?

The Go code won't compile if keywords are duplicated, so I think that's a good enough protection for now.

gasparnagy:
@ehuelsmann The suggestion of @aslakhellesoy about single keyword to type match makes the whole thing much simpler, because in this case, the step keyword type does not have to be represented on the token (the step keyword type can be queried any time based on the keyword as now this is a singular match), but only in the pickle.

I added the code to make the .NET pass (I hope), but my suggestion is that we completely remove the step type from the token and also from the token tests (test data files).

aurelien-reeves:
PHP is green 🎉

ehuelsmann:
> @ehuelsmann The suggestion of @aslakhellesoy about single keyword to type match makes the whole thing much simpler, because in this case, the step keyword type does not have to be represented on the token (the step keyword type can be queried any time based on the keyword as now this is a singular match), but only in the pickle.

@gasparnagy, thanks for your review. The lookup of the keyword can be simplified now that we have 1-to-1 correspondence, indeed. However, I considered what would happen without the `keywordType` in the `Step` object. If we would go with that design, we'd need access to the `Dialect` everywhere that a lookup is required. Currently, the dialect is only available to the parser, at least in the languages that I changed the code on. That's what made me decide to have the keywordType in the Step object.

> I added the code to make the .NET pass (I hope), but my suggestion is that we completely remove the step type from the token and also from the token tests (test data files).

Hmm. How about the need to access the Dialect then?

gasparnagy:
@ehuelsmann my suggestion was not to remove it from the "Step", but to remove it from the token. 

I don't know about the other platforms, but the token in the .NET implementation has a reference to the dialect. With that the keyword type can be calculated for the "Step" in the AST builder.

ehuelsmann:
@gasparnagy Sorry for the misunderstanding. Indeed, I now see that the token has the dialect in the languages that I ported this to (checked Perl and Java). The only reason to keep it as it is, then, is because at token match time, we're already finding out which keyword it is, or at least, it's really easy to: in Java, all keywords used to be lumped into a single array followed by an "exists in array" check for the potential step keyword. By going through each of the sets separately, instead of lumping them into one, we already know what the keyword type is and don't need to do the lookup later.
I did consider adding a method to the dialect to look up a keyword and return its keyword type. However, that doesn't work well with the JS implementation, where the dialect is a simple value map (js object), without any methods.

aurelien-reeves:
I don't know why your commits do not trigger the CI @gasparnagy 🤔

aurelien-reeves:
Thanks a lot @brasmusson for the python implem 🤩

Still dunno why dotnet is failing 🤔👀

gasparnagy:
@aurelien-reeves I look at the dotnet one today. It is strange because if I make the .NET tests pass, the ones that check the result with make/jq fail, but if I try to satisfy those, the others start to fail. Probably some rsync problem. So far I tried to save the efforst of setting up a work make-comatible work env locally, but I think I can't do it further without it... :(

gasparnagy:
@ehuelsmann I understand your argument about the performance and convenience thing. I don't know. I was never good at such a cleanness vs convenience/performance decisions, so I leave it up to you. 

My original vision was that the "token" is a generic concern. It contains fields that are generally available of most token types, like generic "keyword", "text", or "items". So far we did not have any specific field that is only valid for a single token type, like the "keyword type" would be (because that makes only sense for step tokens). 

If we really would like to represent this at the token level, then probably the clean solution would be to split the "step line" token type to "context step line", "action step line", etc. But that would be a bigger change and it would also require to change the grammar and all the token matchers... I don't know...

aurelien-reeves:
Oh my, it seems it is green 🤩🎉

I guess next steps are:
- update the branch to merge incoming commits from main
- set it "ready for review"
- having a last call for discussion/objection
- merge it :)

Well, do we have some documentation to update beside the changelog?

gherkin/php/src/PickleCompiler.php
```diff
@@ -27,9 +27,19 @@
...
 {
+    /** @var array<string, PickleStep\Type|null> */
+    private array $pickleStepTypeFromKeyword = [];
+
+    private Step\KeywordType $lastKeywordType = Step\KeywordType::UNKNOWN;
+
     public function __construct(
         private readonly IdGenerator $idGenerator,
     ) {
+        $this->pickleStepTypeFromKeyword[Step\KeywordType::UNKNOWN->name] = PickleStep\Type::UNKNOWN;
```
ciaranmcnulty, line 38: This can either be a single array assignment:

```php
$this->pickleStepTypeFromKeyword = [
   Step\KeywordType::UNKNOWN->name => PickleStep\Type::UNKNOWN,
// ... etc
];
```

or the same literal array moved to where the property is initialised (currently as `[]`)

gherkin/php/src/PickleCompiler.php
```diff
@@ -184,10 +195,17 @@ private function pickleStep(Step $step, array $variableCells = [], ?TableRow $va
             $argument = null;
         }
 
+        $this->lastKeywordType =
+            $step->keywordType === Step\KeywordType::CONJUNCTION
+            ? $this->lastKeywordType
+            : $step->keywordType ?? Step\KeywordType::UNKNOWN
```
ciaranmcnulty, line 201: Maybe parentheses here for clarity? I can never remember the order of operations

gherkin/php/src/TokenMatcher.php
```diff
@@ -6,13 +6,16 @@
...
 use Cucumber\Gherkin\Parser\TokenMatcherInterface;
+use Cucumber\Messages\Step\KeywordType;
 
 final class TokenMatcher implements TokenMatcherInterface
 {
     private const LANGUAGE_PATTERN = "/^\\s*#\\s*language\\s*:\\s*([a-zA-Z\\-_]+)\\s*$/";
     private GherkinDialect $currentDialect;
     private int $indentToRemove = 0;
     private ?string $activeDocStringSeparator = null;
+    /** @var array<array-key, array<array-key, KeywordType>> */
```
ciaranmcnulty, line 17: We should be specific here

I believe it's an `array<string, list<KeywordTypes>`?

ciaranmcnulty (commented):
PHP implementation looks good, some minor comments

gherkin/php/src/PickleCompiler.php
aurelien-reeves: That is what I tried first, but I did not manage to make psalm happy with this actually

gherkin/php/src/TokenMatcher.php
```diff
@@ -211,12 +221,41 @@ public function match_Language(Token $token): bool
...
+        }
+    }
+
+    private function getKeywordType(string $keyword): KeywordType
+    {
+        if (!array_key_exists($keyword, $this->keywordTypes) || count($this->keywordTypes[$keyword]) !== 1) {
+            return KeywordType::UNKNOWN;
+        }
+
+        return $this->keywordTypes[$keyword][0];
```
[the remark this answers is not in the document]
aurelien-reeves: I've added a comment to avoid any ambiguity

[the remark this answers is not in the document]
ciaranmcnulty: Actually the PHP version we support is high enough we can use https://www.php.net/manual/en/function.array-key-first.php

ehuelsmann:
> Well, do we have some documentation to update beside the changelog?

All that's going on here really is the implementation of a change in `messages`, which was documented in a prior PR. I'd expect all we need to do is to do the changelog item -- all other documentation is in the messages definitions (and there, we have some debt to repay, but that was true far before this PR).

gherkin/php/src/TokenMatcher.php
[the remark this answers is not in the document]
ciaranmcnulty: Sigh ignore me we can use `current()` rather than `[0]` to get the 'next' value irrespective of key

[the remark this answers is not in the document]
aurelien-reeves: I must say that I prefer `[0]` as it is explicit we want the first item in the array. And anyone, even junior developer, should be comfortable with that notation.

`current` implies an iteration, which there is not here. And using `current` without `reset` to get the first item does not looks safe enough imo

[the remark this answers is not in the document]
ciaranmcnulty: Yeah my point earlier was that without it being defined as a `list` we had no guarantees that `0` is the correct key

i.e. `['foo'=>'bar']` has `count(...) === 1`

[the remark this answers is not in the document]
aurelien-reeves: However I have just pushed a commit which removes the `array_key_exists` before the `count` on the array, as `count` will properly returns `0`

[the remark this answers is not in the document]
ciaranmcnulty: If the key doesn't exist there will be a warning (because accessing an unset key) and a fatal error (because then trying to count `null`)

https://3v4l.org/HUONa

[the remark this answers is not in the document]
ciaranmcnulty: Here's another option:

```php
count($this->keywordTypes[$keyword] ?? [])==1
```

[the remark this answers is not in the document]
aurelien-reeves: > Yeah my point earlier was that without it being defined as a `list` we had no guarantees that `0` is the correct key
> 
> i.e. `['foo'=>'bar']` has `count(...) === 1`

I have updated to type annotation of `$this->keywordTypes` to make it an explicit `list` actually



> If the key doesn't exist there will be a warning (because accessing an unset key) and a fatal error (because then trying to count `null`)
> 
> https://3v4l.org/HUONa

My bad. I have misread the doc, and found a bad online php interpreter :'(
I revert this right now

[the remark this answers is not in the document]
aurelien-reeves: > Here's another option:
> 
> ```
> count($this->keywordTypes[$keyword] ?? [])==1
> ```

Nice one indeed 👍

I suggest the following to make it absolutely clear what we actually want to achieve:
```php
    private function getKeywordType(string $keyword): KeywordType
    {
        if (!array_key_exists($keyword, $this->keywordTypes)) {
            return KeywordType::UNKNOWN;
        }

        // We want $this->keywordTypes[$keyword] having **only 1 entry**
        if (count($this->keywordTypes[$keyword]) !== 1) {
            return KeywordType::UNKNOWN;
        }

        return $this->keywordTypes[$keyword][0];
    }
``` 

This can be less performant and redundant, but it removes any doubt on what the code is actually doing

aurelien-reeves:
Last call for discussion and/or objection

Are we good? Can we merge? Are we still waiting for something else?

aslakhellesoy (approved):
LGTM!

aurelien-reeves (approved):

PR #1751, opened 2021-09-13
Fix fake-cucumber and the compatibility kit with pending steps

aurelien-reeves:
## Summary

The "pending" feature says that cucumber execution should fail with pending steps, but the resulting ndjson in the CCK was "success: true".

## Details

Fix fake-cucumber to fail whith pending step, and regenerate a proper ndjson for the compatibility kit

## Motivation and Context

The CCK validation of cucumber-js is failing because of that issue

## How Has This Been Tested?

the resulting ndjson is validating the update of fake-cucumber

## Types of changes

## Checklist:


davidjgoss (approved):
Whoops - thanks for sorting this!

PR #1756, opened 2021-09-16
Convert messages into a hybrid cjs/esm package

aurelien-reeves:
## Summary

This is the continuation of the conversion of JavaScript packages into hybrid cjs/esm packages.

Refs. https://github.com/cucumber/common/pull/1743

## Checklist:

- [x] default.mk and Makefile requires some changes for the "clean" target to avoid deleting `dist/cjs/package.json`


aurelien-reeves:
@aslakhellesoy I think this is done. Could you review it please?

message-streams/javascript/.gitignore
```diff
@@ -1,4 +1,7 @@
-dist/
+dist/*
```
aslakhellesoy, line 1: Rather than adding `dist/cjs/package.json` to git, can we use the approach used here?

https://github.com/cucumber/cucumber-expressions/blob/0875a402cc7efa2d01c1031c96cd2cc448f94b78/javascript/package.json#L21

message-streams/javascript/package.json
```diff
@@ -2,17 +2,34 @@
...
-  "types": "dist/src/index.d.ts",
+  "type": "module",
+  "main": "dist/cjs/src/index.js",
+  "types": "dist/cjs/src/index.d.ts",
+  "files": [
+    "dist/cjs",
+    "dist/esm"
+  ],
+  "module": "dist/esm/src/index.js",
+  "jsnext:main": "dist/esm/src/index.js",
```
aslakhellesoy, line 13: What is `jsnext:main` for?

message-streams/javascript/.gitignore
aurelien-reeves: Yeah, I saw what you did with cucumber-expressions.
Will do 👌

message-streams/javascript/package.json
aurelien-reeves: That is some legacy, it was before the "module" entry-point.
I remove it. I guess this is useless now as "module" is supported since at least node 12.

aurelien-reeves:
@aslakhellesoy one last review? 😊

aslakhellesoy (approved):
Looks good to me. Great work!!

PR #1767, opened 2021-09-24
react: Import correct default styles for Children component

### LINKED ISSUE #1766: [@cucumber/react] Customising Children component's class does not work
**Describe the bug**
Using the CustomRendering component to override classes on the Children component does nothing. The default class is also missing.

**To Reproduce**
1. Pass a custom class to the support field of `CustomRendering` for the `Children` component.
2. The class won't be in the DOM.

**Expected behavior**
Passing a custom class should override the default.

**Context & Motivation**
We're trying to keep @cucumber/react up-to-date on CucumberStudio.

**Additional context**
I'll make a pull request to implement the fix as we figured out where the problem is with @aslakhellesoy.

romaingweb:
## Summary

The Children component imported the styles for Anchor, which didn't load any default classes and prevented overriding the classes because of the way `mixinStyles` is implemented here: https://github.com/cucumber/common/blob/23404c7dee20931ee77817855cf760f0bd91e2a8/react/javascript/src/components/customise/CustomRendering.tsx#L9

## Motivation and Context

Fixes #1766

## How Has This Been Tested?

The change has been tested locally. I don't know if we want to add tests around this part of the custom rendering.

## Types of changes

## Checklist:


davidjgoss (approved):
Thanks @romaingweb!

PR #1778, opened 2021-10-05
tag-expressions: support escape backslashes in tag expressions

### LINKED ISSUE #1776: Tag Expressions cannot handle tags containing backslashes
**Describe the bug**
Javascript, Java, and Go implementations of Tag Expressions cannnot handle tags containing backslashes.
Backslash in Tag Expressions is escape character for "(" and ")" (#285), but there is no way to handle the literal backslash.

**Expected behavior**
Tag Expressions should be able to handle tags containing backslashes.

**Current behavior**
Tag Expressions has no way to handle tags containing backslashes.
On javascript implementation,
```javascript
const { default: parse } = require('@cucumber/tag-expressions');

// cannot handle tag "@foo\bar"
const exp1 = parse('@foo\\bar'); // returns Literal { value: '@foobar' }
exp1.evaluate(['@foo\\bar']); // returns false
exp1.evaluate(['@foobar']); // returns true

// "\\" escaping is not work
const exp2 = parse('@foo\\\\bar'); // returns Literal { value: '@foobar' }
exp2.evaluate(['@foo\\bar']); // returns false
exp2.evaluate(['@foobar']); // returns true
```
Java and Go implementations have same issue.

**Possible Solution**
Add support "\\\\" escape for backslash.
```javascript
const { default: parse } = require('@cucumber/tag-expressions');

// tag "@foo\bar"
const exp1 = parse('@foo\\\\bar'); // returns Literal { value: '@foo\\bar' }
exp1.evaluate(['@foo\\bar']); // returns true
```

**Your Environment**
 - Tag Expressions 4.0.0

**Related Issues**
Issue #285
PR #286

### LINKED ISSUE #1777: The Go implementation of TagExpressions escapes parentheses even if there are other characters between the backslash and the parentheses
**Describe the bug**
The Go implementation of TagExpressions escapes parentheses even if there are other characters between the backslash and the parentheses, such as `(@foo or @\bar)`

**Expected behavior**
Only the parentheses immediately after the backslash are escaped.
```go
expr1, err1 = Parse("(@foo or @\bar)") // success
expr1.Evaluate([]string{"@bar"}) // return true
expr2, err2 = Parse("@foo or @\bar)") // return error "Syntax error. Unmatched )"
```

**Current behavior**
The parentheses are escaped even if there are other characters between the backslash and the parentheses.
```go
expr1, err1 = Parse("(@foo or @\bar)") // should success, but return error "Syntax error. Unmatched ("
expr2, err2 = Parse("@foo or @\bar)") // should return error "Syntax error. Unmatched )", but success
expr1.Evaluate([]string{"@bar"}) // return false
expr1.Evaluate([]string{"@bar)"}) // return true
```

**Your Environment**
 - Tag Expressions 4.0.0

yusuke-noda:
## Summary

Add support "\\\\" escape for backslash in tag expressions.

## Details
Javascript, Java, and Go implementations of Tag Expressions cannnot handle tags containing backslashes, but Tag Expressions should be able to handle tags containing backslashes.

## Motivation and Context
Fixes #1776 
Fixes #1777

## How Has This Been Tested?
Added the following test cases:
- "\\" before "\\". `a\\ and b`
- "\\" before normal character. `\a and b`
- "\\" before space. `a\ and b`
- "\\" is the last character of tag expressions. `a and b\`
- "\\\\" before unescaped parentheses. `( a and b\\)`
- "\\\\" before escaped parentheses. `a\\\( and b\\\)`
- normal characters between "\\" and parentheses. `(a and \b)`

## Types of changes

## Checklist:

### DIFF
--- a/tag-expressions/go/parser.go
+++ b/tag-expressions/go/parser.go
@@ -112,6 +112,7 @@ func tokenize(expr string) []string {
 	for _, c := range expr {
 		if unicode.IsSpace(c) {
 			collectToken()
+			escaped = false
 			continue
 		}
 
@@ -119,7 +120,12 @@ func tokenize(expr string) []string {
 
 		switch ch {
 		case "\\":
-			escaped = true
+			if escaped {
+				token.WriteString(ch)
+				escaped = false
+			} else {
+				escaped = true
+			}
 		case "(", ")":
 			if escaped {
 				token.WriteString(ch)
@@ -130,6 +136,7 @@ func tokenize(expr string) []string {
 			}
 		default:
 			token.WriteString(ch)
+			escaped = false
 		}
 	}
 
@@ -192,7 +199,12 @@ func (l *literalExpr) Evaluate(variables []string) bool {
 
 func (l *literalExpr) ToString() string {
 	return strings.Replace(
-		strings.Replace(l.value, "(", "\\(", -1),
+		strings.Replace(
+			strings.Replace(l.value, "\\", "\\\\", -1),
+			"(",
+			"\\(",
+			-1,
+		),
 		")",
 		"\\)",
 		-1,
--- a/tag-expressions/go/parser_test.go
+++ b/tag-expressions/go/parser_test.go
@@ -42,6 +42,16 @@ func TestParseForValidCases(t *testing.T) {
 			given:    "not a\\(\\) or b and not c or not d or e and f",
 			expected: "( ( ( not ( a\\(\\) ) or ( b and not ( c ) ) ) or not ( d ) ) or ( e and f ) )",
 		},
+		{
+			name:     "test escaping backslash",
+			given:    "(a\\\\ and b\\\\\\( and c\\\\\\) and d\\\\)",
+			expected: "( ( ( a\\\\ and b\\\\\\( ) and c\\\\\\) ) and d\\\\ )",
+		},
+		{
+			name:     "test backslash",
+			given:    "(a\\ and \\b) and c\\",
+			expected: "( ( a and b ) and c )",
+		},
 	}
 
 	for _, tc := range cases {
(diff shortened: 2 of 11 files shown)


aurelien-reeves:
We may have to move that PR to https://github.com/cucumber/tag-expressions 🤔

tag-expressions/go/parser.go
```diff
@@ -112,14 +112,20 @@ func tokenize(expr string) []string {
...
 
 		switch ch {
 		case "\\":
-			escaped = true
+			if escaped {
+				token.WriteString(ch)
+				escaped = false
+			} else {
+				escaped = true
+			}
```
mpkorstanje, line 128: I would not be opposed to lifting this case out into a separate if-block so that the `escaped = false` in the `unicode.IsSpace(c) {` block is not needed.

mpkorstanje:
> We may have to move it to https://github.com/cucumber/tag-expressions thinking

@aurelien-reeves a "maybe" isn't actionable. Can you either provide clear directions or indicate that you are still looking for a solution?

mpkorstanje (commented):
In principle the changes look good to me.

@yusuke-noda are you participating in hacktober fest? I don't think we can merge this straight away but I can mark the PR as accepted for the purposes of the event.

aurelien-reeves:
> 
> 
> > We may have to move it to https://github.com/cucumber/tag-expressions thinking
> 
> @aurelien-reeves a "maybe" isn't actionable. Can you either provide clear directions or indicate that you are still looking for a solution?

That was the point of the "🤔" smiley. To start discussing about that because I am not sure of what should be done.

I think we should move that PR to https://github.com/cucumber/tag-expressions
What do you think?

mpkorstanje:
@mattwynne is there a technical reason that tag-expressions is still in located in common? Did the migration to https://github.com/cucumber/tag-expressions fail?

yusuke-noda:
> @yusuke-noda are you participating in hacktober fest? I don't think we can merge this straight away but I can mark the PR as accepted for the purposes of the event.

Thank you, but i am not participating in hacktober fest.

aslakhellesoy:
Let's merge this here. Then make a release.

When that's done there won't be any open issues/PRs for tag-expressions, and we can do the migration to the new repo again.

PR #1784, opened 2021-10-08
[compatibility-kit] Canonicalize NDJSON

ehuelsmann:
## Summary

Canonicalize compatibility-kit NDJSON test files to facilitate manipulation using Cucumber Messages.

## Details

While trying to modify the content of the NDJSON files with Cucumber Messages,
I run into the problem that the serialized messages are canonical NDJSON, but
the original is not, causing unnecessary and hard-to-review differences.

## Motivation and Context

In preparation of adding a field to some messages, which I expect to add by processing these messages through the messages library, I need the resulting diff later to be minimal for maximum reviewability.

## How Has This Been Tested?

By running the test framework.

## Types of changes

- [x] Refactoring (different code achieving the same functionality).

## Checklist:


aslakhellesoy:
We generate these messages whenever we feel they need to be regenerated:

```
cd compatibility-kit/javascript
GOLDEN=1 make clean
GOLDEN=1 make
```

So next time we do that, your changes will be lost.

What does it mean for NDJSON to be "canonical"? What command did you run to canonicalize them? (Again, I'm not sure what it means).

PR #1824, opened 2021-11-05
[create-meta][javascript] Generate ciDict.ts from ciDict.json

aurelien-reeves:
## Summary

For future conversion of this package into a hybrid esm/cjs one, we need
to convert ciDict.json into a typescript file.


aurelien-reeves:
Reported here: https://github.com/cucumber/create-meta/pull/5

PR #1858, opened 2022-01-05
Java optional codegen

aslakhellesoy:
## Summary

Use custom code generation for Java

## Details

Switch from using `jsonschema2pojo-maven-plugin` to using a custom code generator.

## Motivation and Context

In order to have better type safety for non-required fields, we want the generated Java code to use `Optional` in getters. See https://github.com/cucumber/ci-environment/pull/51 for context.

## Types of changes

## Checklist:


jamietanna:
Is this required? I see `useOptionalForGetters` is an option for the Maven/[Gradle](https://github.com/joelittlejohn/jsonschema2pojo/tree/master/jsonschema2pojo-gradle-plugin) plugin so it should mean we can avoid hand-writing stuff?

gherkin/java/src/main/java/io/cucumber/gherkin/Gherkin.java
```diff
@@ -93,10 +92,10 @@ private static String read(Reader reader) throws IOException {
         if (includeSource) {
             messages.add(envelope);
         }
-        if (envelope.getSource() != null) {
+        if (envelope.getSource().isPresent()) {
 
             Parser<GherkinDocument> parser = new Parser<>(new GherkinDocumentBuilder(idGenerator));
-            Source source = envelope.getSource();
+            Source source = envelope.getSource().get();
```
mpkorstanje, line 98: Consider using `.map` to map the source to a collection of messages.

gherkin/java/src/test/java/io/cucumber/gherkin/GherkinTest.java
```diff
@@ -32,9 +32,9 @@ public void use_this_in_readme() {
         boolean includeAst = true;
         boolean includePickles = true;
         Stream<Envelope> envelopeStream = Gherkin.fromPaths(paths, includeSource, includeAst, includePickles, idGenerator);
-        Stream<Envelope> pickleStream = envelopeStream.filter(envelope -> envelope.getPickle() != null);
+        Stream<Envelope> pickleStream = envelopeStream.filter(envelope -> envelope.getPickle().isPresent());
 
-        assertEquals("minimalistic", pickleStream.collect(Collectors.toList()).get(0).getPickle().getName());
+        assertEquals("minimalistic", pickleStream.collect(Collectors.toList()).get(0).getPickle().get().getName());
```
mpkorstanje, line 37: Consider using `Stream.findFirst()` instead of `Stream.collect`.

gherkin/java/src/main/java/io/cucumber/gherkin/Gherkin.java
aslakhellesoy: Can you paste an example? I'm still grappling with Optional and couldn't figure this one out. Sorry for being thick.

aslakhellesoy:
The custom code generator I implemented would do `requireNonNull` checks in getters, setters and the argument constructor (not the no-arg constructor though). This will fail earlier in case we forget to set some required fields.

I could add a `#validate()` method as well which would do the same, and we could call this in the code that puts the messages on the event bus.

Wortwhile restoring the custom code generator for those purposes? I think so!

gherkin/java/src/main/java/io/cucumber/gherkin/Gherkin.java
```diff
@@ -92,45 +94,55 @@ private static String read(Reader reader) throws IOException {
...
+
+            if (includeGherkinDocument) {
+                gherkinDocument = parser.parse(data);
+                gherkinDocument.setUri(uri);
+                Envelope gherkinDocumentEnvelope = new Envelope();
+                gherkinDocumentEnvelope.setGherkinDocument(gherkinDocument);
+                messages.add(gherkinDocumentEnvelope);
+            }
+            if (includePickles) {
+                if (gherkinDocument == null) {
```
mpkorstanje, line 125: `gherkinDocument` is always null.

aslakhellesoy:
I've added back the custom code generator so that we can have better validation at runtime. @mpkorstanje ok to merge?

gherkin/java/src/main/java/io/cucumber/gherkin/pickles/PickleCompiler.java
```diff
@@ -44,10 +44,10 @@ public PickleCompiler(IdGenerator idGenerator) {
 
     public List<Pickle> compile(GherkinDocument gherkinDocument, String uri) {
         List<Pickle> pickles = new ArrayList<>();
-        Feature feature = gherkinDocument.getFeature();
-        if (feature == null) {
+        if (!gherkinDocument.getFeature().isPresent()) {
             return pickles;
         }
+        Feature feature = gherkinDocument.getFeature().get();
```
mpkorstanje, line 50: This is `.isPresent` followed by `.get` that can be replaced with `.map`. There are few more in this method.

messages/jsonschema/scripts/templates/java.java.erb
```diff
@@ -0,0 +1,96 @@
+<% @schemas.each do |key, schema| %>
+    public static class <%= class_name(key) %> {
+      <%- schema['properties'].each do |property_name, property| -%>
+        private <%= type_for(class_name(key), property_name, property) -%> <%= property_name %><%= property['items'] ? ' = new java.util.ArrayList<>()' : '' %>;
+      <%- end -%>
+
+        public <%= class_name(key) %>() {}
```
mpkorstanje, line 7: Consider removing the default constructor.

messages/jsonschema/scripts/templates/java.java.erb
```diff
@@ -0,0 +1,96 @@
...
+            <%- else -%>
+            this.<%= property_name %> = <%= property_name %>;
+            <%- end -%>
+          <%- end -%>
+        }
+
+      <%- schema['properties'].each do |property_name, property| -%>
+        <%- if (schema['required'] || []).index(property_name) -%>
+        public <%= type_for(class_name(key), property_name, property) -%> get<%= capitalize(property_name) %>() {
+            return java.util.Objects.requireNonNull(<%= property_name %>, "<%= class_name(key) %>.<%= property_name %> cannot be null");
```
mpkorstanje, line 26: Getters typically don't have a null check. Ideally null checks are handled in the constructor and setters.

messages/jsonschema/scripts/templates/java.java.erb
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        public void set<%= capitalize(property_name) %>(<%= type_for(class_name(key), property_name, property) -%> <%= property_name %>) {
+            this.<%= property_name %> = java.util.Objects.requireNonNull(<%= property_name %>, "<%= class_name(key) %>.<%= property_name %> cannot be null");
```
mpkorstanje, line 30: Are setters are required?

messages/jsonschema/scripts/templates/java.java.erb
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+            <%= class_name(key) %> that = (<%= class_name(key) %>) o;
+            return <%- schema['properties'].each_with_index do |(property_name, property), index| %>
+                <%- if (schema['required'] || []).index(property_name) -%>
+                <%= property_name -%>.equals(that.<%= property_name -%>)<%= index < schema['properties'].length-1 ? ' && ' : ';' -%>
```
mpkorstanje, line 69: With the default constructor present this would result in null pointers.

mpkorstanje:
> @mpkorstanje ok to merge?

I can't quite put my finger on it but I get the impression you're still in two minds about the exact way to handle the null checks. I'm currently expecting a single constructor and no setters (essentially value objects). 

If you're running into trouble with Jackson due to a lack of default constructors and absence of setters, you should provide the https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#parameters option to maven. That will add sufficient information for Jackson to deserialize to value objects.

aslakhellesoy:
I should have made it more clear why I went with a dual empty/arg constructor approach. As you know, I’m a big fan of constructor injection (picocontainer).

However, I stumbled upon one some code in the pickle compiler where it seemed a constructor injection approach would be ugly. And I thought there might be some places in Cucumber-JVM too, but I didn’t check.

This is the only reason I kept empty constructors. And the only reason I added `validate()`, setters and validation in getters is because of empty constructors. It wasn’t because of Jackson.

If we remove empty constructors we can also remove `validate()`, setters and validation in getters.

It would be a more purist implementation, but I think it might come at the cost of some complex code where complex objects are built. But
maybe I am wrong and that it can be done cleanly with functional code.

Do you want me to do that?

mpkorstanje:
> However, I stumbled upon one some code in the pickle compiler where it seemed a constructor injection approach would be ugly. And I thought there might be some places in Cucumber-JVM too, but I didn’t check.

Mmh. That's a valid reason. ASTs can be a pain when not mutable. 

Though I think it can be done. Do you have an example of a problem?

aslakhellesoy:
IIRC it was this: https://github.com/cucumber/common/blob/b235cf267d1b2049f6bcc16dbdff9f5dddb4e450/gherkin/java/src/main/java/io/cucumber/gherkin/pickles/PickleCompiler.java#L184

Looking at it now it doesn’t seem that hard to do. Let’s give it a try!

aslakhellesoy:
@mpkorstanje Another problem is this:

```java
            Envelope envelope = new Envelope();
            envelope.setSource(new Source("hello.feature", "Feature: Hello", SourceMediaType.TEXT_X_CUCUMBER_GHERKIN_PLAIN));
```

If we don't allow empty constructors, we have to create the `Envelope` like this:

```java
Envelope envelope = new Envelope(
        null,
        null,
        null,
        null,
        null,
        null,
        null,
        new Source("hello.feature", "Feature: Hello", SourceMediaType.TEXT_X_CUCUMBER_GHERKIN_PLAIN),
        null,
        null,
        null,
        null,
        null,
        null,
        null,
        null,
        null
);
```

You ok with that?

aslakhellesoy:
One idea - if *all* of the properties are non-required (such as `Envelope`), generate a 1-arg constructor for each property. WDYT?

mpkorstanje:
The pattern for the envelope isn't that all fields are not required, but rather exactly-1 field is required. So the constructor would be:

```
private final Source source;
private final Other other;

public Envelopse(Source source) {
  this.source = requireNonNull(source)
  this.other = null;
}
```

But I don't know if the json schema can capture that. 

I also think that Envelope might be the only example that works this way. So I think it would be better for now to treat it as own special case rather then come up with generic rules.

aslakhellesoy:
JSON Schema has a [oneOf](http://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.10.2.1.3) keyword.

I think it would be a lot of work to support this in our custom code generators, so I've opted for a simpler solution in 5daf24a7bd2c364213d7cb9c52fbfe135d49169d.

WDYT?

messages/jsonschema/scripts/templates/java.java.erb
aslakhellesoy: Only classes without required fields have a public empty constructor now.

messages/jsonschema/scripts/templates/java.java.erb
aslakhellesoy: Fixed

messages/jsonschema/scripts/templates/java.java.erb
aslakhellesoy: Setters are gone now

messages/jsonschema/scripts/templates/java.java.erb
aslakhellesoy: Fields that are nullable are compared with `java.util.Objects.equals`. Non-nullable fields cannot be null because they are verified in `.validate()` (called in the constructor).

messages/jsonschema/scripts/templates/java.enum.java.erb
```diff
@@ -0,0 +1,39 @@
...
+        public String toString() {
+            return this.value;
+        }
+
+        public String value() {
+            return this.value;
+        }
+
+        public static <%= enum[:name] %> fromValue(String value) {
+            <%= enum[:name] %> constant = CONSTANTS.get(value);
```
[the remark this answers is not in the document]
aslakhellesoy: Good idea. Fixed.

aslakhellesoy:
> Consider overloading `Envelope.from(PropertyName propertyName)` with different parameter types instead of naming the method differently for each property. `Envelope.fromPropertyName(new PropertyName(....))` is very verbose.

Done! Been doing too much TypeScript lately...

aslakhellesoy:
I think it would be cleaner to return unmodifiable lists, to make the objects truly immutable. Agree?

aslakhellesoy:
I made them unmodifiable :-)

messages/java/src/main/java/io/cucumber/messages/Messages.java
```diff
@@ -2284,7 +2284,7 @@ public ParameterType(
             String id
         ) {
             this.name = name;
-            this.regularExpressions = regularExpressions;
+            this.regularExpressions = regularExpressions == null ? null : java.util.Collections.unmodifiableList(regularExpressions);
```
mpkorstanje, line 2287: Consider using `Collections.emptyList()` instead of `null`. 

Also can there be a parameter type without a regular expression?

mpkorstanje: Also consider using `Collections.unmodifiableList(new ArrayListt(regularExpressions))` to ensure that the the collection in `ParameterType` can not be modffified by accident by any one who still holds a reference to `regularExpressions`.

mpkorstanje:
> I made them unmodifiable :-)

Great. It's not the same as immutable though. Nice little gotcha from way back then.

messages/java/src/main/java/io/cucumber/messages/Messages.java
aslakhellesoy: What the! I always thought `unmodifiableList` made a copy rather than holding a reference. But it makes sense - someone might want to get updates made elsewhere without being allowed to update it. Fixed.

aslakhellesoy: > Consider using `Collections.emptyList()` instead of `null`.
> 
> Also can there be a parameter type without a regular expression?

The `ParameterType.regularExpressions` field is not nullable. This is verified in `validate()`. 

I added a few [minItems])(https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.6.4.2) in 29f85f7380c37d511bdbec1b54ede5e75597b31c and updated the code generator to verify the number of elements.

aslakhellesoy: > Consider using `Collections.emptyList()` instead of `null`

I don't think we should override what the user passes in. If they want an empty list, pass an empty list. Passing null when a list is required should be an error (validated later in `validate()`)

gherkin/java/src/main/java/io/cucumber/gherkin/Gherkin.java
```diff
@@ -50,9 +52,7 @@ private Gherkin(List<String> paths, List<Envelope> envelopes, boolean includeSou
     }
 
     public static Envelope makeSourceEnvelope(String data, String uri) {
```
mpkorstanje, line 54: This method is only used in tests. Consider removing it to reduce the exposed API.

messages/jsonschema/scripts/templates/java.java.erb
```diff
@@ -0,0 +1,111 @@
...
+        }
+        <%- else -%>
+
+        public java.util.Optional<<%= type_for(class_name(key), property_name, property) -%>> get<%= capitalize(property_name) %>() {
+            return java.util.Optional.ofNullable(<%= property_name %>);
+        }
+        <%- end -%>
+      <%- end -%>
+
+        private void validate() {
```
mpkorstanje, line 55: Don't forget to remove this.

mpkorstanje: If the constructors already have the `requireNonNull` for each required argument then what is there left to validate?

mpkorstanje:
> Good to merge @mpkorstanje ?

I'll have to take a look at the generated code. But hard to say right now just from the git diff.

mpkorstanje:
Ah! Thanks that helps a lot!

Consider:
 1. making all fields `final`. This will allow the compiler to enforce that all fields are set in a constructor.
 2. Use the pattern `this.field = requireNonNull(field);` instead of a separate validate method.
 3. And I think that since java 17 the message can be omitted from `requireNonNull` as java can see which argument was the problem. Not sure about that.

The combination of the first two make it easy to read the code and see what are the required and optional fields

aslakhellesoy:
If we make all fields `final` we cannot use the current implementation of `from`. Instead we would have to implement `from` like this:

```java
public static Envelope from(Source source) {
    return new Envelope(
            null,
            null,
            null,
            null,
            null,
            null,
            null,
            Objects.requireNonNull(source, "Envelope.source cannot be null"),
            null,
            null,
            null,
            null,
            null,
            null,
            null,
            null,
    );
}
```

That's fine with me - is this what you're suggesting? With this we can also remove the private empty constructors, as well as the `validate()` method since the validation no longer needs to be called in `from` - it can be inlined in the constructor.

mpkorstanje:
> Instead we would have to implement from like this:

No major objects. Though what was the reason we wanted to use `from` and not a 1-argument constructor?

messages/java/src/main/java/io/cucumber/messages/Messages.java
```diff
@@ -0,0 +1,4417 @@
...
+        }
+
+        public String value() {
+            return this.value;
+        }
+
+        public static SourceMediaType fromValue(String value) {
+            if ("TEXT_X_CUCUMBER_GHERKIN_PLAIN".equals(value)) return TEXT_X_CUCUMBER_GHERKIN_PLAIN;
+            if ("TEXT_X_CUCUMBER_GHERKIN_MARKDOWN".equals(value)) return TEXT_X_CUCUMBER_GHERKIN_MARKDOWN;
+            throw new IllegalArgumentException(value);
```
mpkorstanje: This implementation is equivalent to the `Enu.valueOf`. Did you mean to compare against `.value`? 

https://docs.oracle.com/javase/7/docs/api/java/lang/Enum.html#valueOf(java.lang.Class,%20java.lang.String)

aslakhellesoy: Yeah just noticed - fixed in 288ba0c06750542ce061bb1480c3d6ffef04a289 on #1879

aslakhellesoy:
> > Instead we would have to implement from like this:
> 
> No major objects. Though what was the reason we wanted to use `from` and not a 1-argument constructor?

So we don't have to create `Envelope` objects like this:

```java
return new Envelope(
            null,
            null,
            null,
            null,
            null,
            null,
            null,
            Objects.requireNonNull(source, "Envelope.source cannot be null"),
            null,
            null,
            null,
            null,
            null,
            null,
            null,
            null,
    )
```

aslakhellesoy:
Oh I see what you mean - might as well be a 1-arg constructor. I'll change to that in #1879

aslakhellesoy:
Closing in favour of #1879

PR #1879, opened 2022-01-17
Java optional codegen with final fields

aslakhellesoy:
This PR supersedes #1858 with suggestions made by @mpkorstanje in https://github.com/cucumber/common/pull/1858#issuecomment-1013131214

The code compiles, but the tests are failing because the classes don't have an empty constructor. I have enabled `<parameters>true</parameters>` for the compiler plugin as suggested in https://github.com/cucumber/common/pull/1858#issuecomment-1007508805, but it doesn't seem to help.


aslakhellesoy:
I've figured out the Jackson problems now.

messages/java/src/main/java/io/cucumber/messages/MessageToNdjsonWriter.java
```diff
@@ -15,7 +15,7 @@ public MessageToNdjsonWriter(OutputStream out) {
 
     @Override
     public void write(Object message) throws IOException {
-        JSON.writeValue(out, message);
+        Jackson.OBJECT_MAPPER.writeValue(out, message);
```
mpkorstanje, line 18: I wonder, could we remove Jackson and implement this in Cucumber JVM?

Maybe construct this class with a `Function<Object, String> serializer` and only use Jackson in tests.

mpkorstanje: Ah wait. Wrong function signature.

aslakhellesoy: Sure, that should be doable. Let me handle that in a separate PR.

mpkorstanje: `MessageToNdjsonWriter` should also implement closable.

mpkorstanje:
> Consider enabling the https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#parameters option to maven. That will add sufficient information for Jackson to deserialize to value objects.

aslakhellesoy:
> Consider enabling the https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#parameters option to maven. That will add sufficient information for Jackson to deserialize to value objects.

I've already done that.

mpkorstanje:
I'm not seeing it in this MR.

mpkorstanje:
Nevemrind. It's there. Dunno why I can't see it in the MR.

messages/java/pom.xml
```diff
@@ -73,25 +83,11 @@
...
+                <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
-                    <sourceDirectory>${basedir}/../jsonschema</sourceDirectory>
-                    <includes>*.json</includes>
-                    <annotationStyle>none</annotationStyle>
-                    <useLongIntegers>true</useLongIntegers>
-                    <includeConstructors>true</includeConstructors>
-                    <includeGeneratedAnnotation>false</includeGeneratedAnnotation>
-                    <targetPackage>io.cucumber.messages.types</targetPackage>
+                    <parameters>true</parameters>
```
aslakhellesoy, line 89: @mpkorstanje this is where `-parameters` is enabled

aslakhellesoy: The tests wouldn't have passed if I hadn't done it :-)

messages/java/src/main/java/io/cucumber/messages/MessageToNdjsonWriter.java
aslakhellesoy: What about the deserializer we'd have to add to `NdjsonToMessageIterable`? `BiFunction`?

mpkorstanje: Do we ever de-serialize outside of the tests?

aslakhellesoy: No production code uses deserialization, but we do use it in Gherkin's acceptance tests.

I think it would be nice to ship a general `NdjsonToMessageIterable` class with the library, but force users to construct it with a deserializer.

If we refactor the Gherkin acceptance tests to be parameterised tests a la cucumber expressions, we can do it all in JUnit, and we wouldn't need to depend on Jackson in Gherkin. Alternatively we move the Gherkin `Main` class over to `test`.

mpkorstanje: Ok. Sounds like a separate MR.

aslakhellesoy:
Hmmmm - the HTML formatter tests fail. Could it be that the shade plugin removes the parameter info from messages? If that's the case, maybe we should tackle the jackson decoupling in this PR after all?

messages/java/src/main/java/io/cucumber/messages/Messages.java
```diff
@@ -0,0 +1,4018 @@
...
+        public String toString() {
+            return "TestRunFinished{" +
+                "message=" + message +
+                ", success=" + success +
+                ", timestamp=" + timestamp +
+                '}';
+        }
+    }
+
+    public static class TestRunStarted {
```
mpkorstanje: Consider using `public static final` to reduce the api surface.

messages/java/src/main/java/io/cucumber/messages/Messages.java
```diff
@@ -0,0 +1,4018 @@
+package io.cucumber.messages;
+
+public class Messages {
```
mpkorstanje: Consider using `public final` to reduce the API service. And since this is a utility class it can also have a private constructor.

mpkorstanje:
```
Caused by: io.cucumber.messages.internal.com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `io.cucumber.messages.Messages$TestRunStarted` (no Creators, like default constructor, exist): cannot deserialize from Object value (no delegate- or property-based Creator)
 at [Source: (String)"{"testRunStarted":{"timestamp":{"nanos":0,"seconds":0}}}"; line: 1, column: 20] (through reference chain: io.cucumber.messages.Messages$Envelope["testRunStarted"])
```

At a glance I'd think the `ParameterNamesModule` would have to be added to the object mapper but I'm not sure.

mpkorstanje:
We'll need all those modules:

https://github.com/FasterXML/jackson-modules-java8#registering-modules

messages/java/src/main/java/io/cucumber/messages/Messages.java
aslakhellesoy: Fixed!

messages/java/src/main/java/io/cucumber/messages/Messages.java
aslakhellesoy: Fixed!

messages/java/pom.xml
```diff
@@ -44,19 +44,28 @@
     </dependencyManagement>
 
     <dependencies>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
+            <scope>compile</scope>
```
mpkorstanje, line 50: Are we using Jackson annotations anywhere?

messages/java/src/main/java/io/cucumber/messages/NdjsonToMessageIterable.java
```diff
@@ -1,43 +1,53 @@
...
 
 /**
  * Iterates over messages read from a stream. Client code should not depend on this class
  * directly, but rather on a {@code Iterable<Messages.Envelope>} object.
  * Tests can then use a {@code new ArrayList<Messages.Envelope>} which implements the same interface.
  */
-public final class NdjsonToMessageIterable implements Iterable<Envelope> {
-    private final BufferedReader input;
-    private Envelope next;
+public final class NdjsonToMessageIterable<T> implements Iterable<T>, AutoCloseable {
```
mpkorstanje, line 17: I think it's okay to keep this pinned to `Envelope`.

mpkorstanje:
>  Without the @JsonCreator constructor annotation I wasn't able to deserialize all messages.

Which ones were those?

messages/java/src/main/java/io/cucumber/messages/Messages.java
```diff
@@ -0,0 +1,4077 @@
...
+                ", url=" + url +
+                '}';
+        }
+    }
+
+    public static final class Duration {
+        private final Long seconds;
+        private final Long nanos;
+
+        @JsonCreator
```
mpkorstanje: This one shouldn't be needed if Jackson has the `ParameterNamesModule`.

messages/java/src/main/java/io/cucumber/messages/Messages.java
```diff
@@ -0,0 +1,4077 @@
...
+        public Duration(
+            Long seconds,
+            Long nanos
+        ) {
```
mpkorstanje: At a glance this meets the requirements on:

https://github.com/FasterXML/jackson-modules-java8/tree/master/parameter-names#usage-example

Not sure why it would need the annotation

mpkorstanje:
Ah. The problem is `Envelope` again. It has multiple constructors.

mpkorstanje:
So either:
 * we're back to using `Envelope.from` or;
 * we use a custom annotation introspector or;
 * something else.

aslakhellesoy:
Yes, I suggested a custom annotation introspector in https://github.com/cucumber/common/pull/1879#issuecomment-1015448240 - let's go for that.

But can we please do this in a separate PR? This one has been going on for so long now...

aslakhellesoy:
Using `Envelope.from` is not a solution. I ran into a problem with single-argument constructors (hence the new test for `TestRunStarted`. Jackson deserialization just refuses to work with single-arg constructors unless you put an annotation on it.

messages/java/src/main/java/io/cucumber/messages/Messages.java
aslakhellesoy: That one doesn't need it, but the code generator adds the annotation to all the constructors.

messages/java/src/main/java/io/cucumber/messages/NdjsonToMessageIterable.java
aslakhellesoy: In theory yes, but when I was debugging the internals of Jackson trying to figure out why it wouldn't deserialize a single argument constructor (see `handles_single_argument_constructors`) I wanted to have the tiniest object possible.

messages/java/src/test/java/io/cucumber/messages/NdjsonSerializationTest.java
```diff
@@ -11,28 +11,43 @@
...
+            }
+        });
     }
 
     @Test
     void writes_source_envelope() throws IOException {
         ByteArrayOutputStream output = new ByteArrayOutputStream();
-        MessageWriter writer = makeMessageWriter(output);
+        MessageWriter<Source> writer = makeMessageWriter(output);
         writer.write(new Source("uri", "data", SourceMediaType.TEXT_X_CUCUMBER_GHERKIN_PLAIN));
```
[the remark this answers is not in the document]
aslakhellesoy: I wanted to keep it consistent with the iterable implementation - see above.

messages/java/src/main/java/io/cucumber/messages/Messages.java
aslakhellesoy: Not this one in particular, but single-arg constructors do need it. If you remove the annotation from `TestRunStarted` you'll get a failing test.

mpkorstanje:
> Jackson deserialization just refuses to work with single-arg constructors unless you put an annotation on it.

Gimme a chance to debug that for a minute. There was a fix for that in V3 I think.

https://github.com/FasterXML/jackson-databind/issues/1498

mpkorstanje:
https://github.com/FasterXML/jackson-module-parameter-names/issues/21#issuecomment-111000035

Less cumbersome work around. Maybe. I don't have time to try this right now.

mpkorstanje:
@aslakhellesoy have a look at the `GherkinParser`. This is what I meant with using the builder pattern. The `parse` method that does the heavy lifting is now only 20 lines long and the code around is complexity free.

messages/java/src/main/java/io/cucumber/messages/NdjsonToMessageIterable.java
mpkorstanje: The problem with Jackson has been solved so I've removed it again. This is part of the public API and this would make it more complex the needed.

messages/java/Makefile
```diff
@@ -2,4 +2,24 @@ include default.mk
...
+	echo >> $@
+	echo "// Generated code" >> $@
+	echo "@SuppressWarnings(\"unused\")" >> $@
+	echo "public final class Messages {" >> $@
+	echo >> $@
+	echo "    private Messages() {}" >> $@
+	echo >> $@
+	ruby ../jsonschema/scripts/codegen.rb Java ../jsonschema java.java.erb >> $@
+	ruby ../jsonschema/scripts/codegen.rb Java ../jsonschema java.enum.java.erb >> $@
+	echo "}" >> $@
```
mpkorstanje, line 25: Would it be possible to get this header and footer into their own files too. Would make using `make` a bit easier. Make currently doesn't detect changes here.

gherkin/CHANGELOG.md
```diff
@@ -13,10 +13,14 @@ This document is formatted according to the principles of [Keep A CHANGELOG](htt
 
 ### Changed
 
+* [Java] the `Parser.parse` methods now require an `uri` parameter to be passed in.
```
mpkorstanje, line 16: TODO: This is out of date.

messages/java/Makefile
```diff
@@ -2,4 +2,24 @@ include default.mk
...
+	echo "import java.util.ArrayList;" >> $@
+	echo "import java.util.Objects;" >> $@
+	echo "import java.util.Optional;" >> $@
+	echo >> $@
+	echo "import static java.util.Collections.unmodifiableList;" >> $@
+	echo "import static java.util.Objects.requireNonNull;" >> $@
+	echo >> $@
+	echo "// Generated code" >> $@
+	echo "@SuppressWarnings(\"unused\")" >> $@
+	echo "public final class Messages {" >> $@
```
mpkorstanje, line 19: Is it a limitation of the code generator that all classes must be inside a single file? I think it would be better if they're all in the messages package.

mpkorstanje:
I've cleaned up the public API of gherkin, messages and the html formatter, added null checks.

messages/java/Makefile
aslakhellesoy: Yes, generating a single file is currently a limitation. What's the problem with a single file?

mpkorstanje: A few  reasons:

0. No real reason to use an inner class that acts like a package.
1. Editors don't like big files. 
2. The `import io.cucumber.messages.Messages.Something` is stuttering.
3. Smaller change compared to the previous API. So fewer imports to fix while migrating.
4. We're currently flip flopping between using inner classes (protobuf, this PR) and upper level classes ( as generated by the maven plugin). It would be nice to be consistent.

mpkorstanje:
I think we're good, but let's squash merge this. The commits are very noisy.

aslakhellesoy:
Wooot great work @mpkorstanje

PR #1884, opened 2022-01-21
messages: PHP implementation

ciaranmcnulty:
This is a PHP implementation of Cucumber Messages (with thanks to @aurelien-reeves)

## Todo before merge

- [x] README
- [x] Extract Makefile into default.mk templates 
- [x] hook up with CI build
- [x] Add PHP to docker image

## To make usable as a package

- [x] Create separate sub repo
- [ ] Cucumber user (Cukebot?) register with Packagist.org and set up GitHub hooks


aurelien-reeves:
Nice! Looks great! 🤩

messages/jsonschema/scripts/templates/php.php.erb
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Represents the <%= class_name(key) %> message in Cucumber's message protocol
+ * @see https://github.com/cucumber/common/tree/main/messages#readme
+ *
+<%=- format_description(schema['description'], indent_string: '') -%> */
+final class <%= class_name(key) %> implements JsonSerializable
+{
+    use JsonEncodingTrait;
+
+    private function __construct(
```
ciaranmcnulty, line 23: In retrospect this should be public, otherwise gherkin implementation won't be able to do much useful

aurelien-reeves:
I have created a subrepo: https://github.com/cucumber/messages-php
It will be updated when that PR will be merge on main

messages/jsonschema/scripts/templates/php.php.erb
aurelien-reeves: Indeed, in other implementation, we instantiate messages directly without de-serializing those from arrays or json.
It would be more efficient to do the same with php I suppose.

ciaranmcnulty: They're now able to be constructed (see README)

aurelien-reeves:
@ciaranmcnulty I've worked on adding php templates to the monorepo and messages/php to the CI.
I let you take a look on the makefiles if they looks fine

Regarding the CI, we may have to install php as part of the build image actually, as the codegen is written in ruby, we need both php and ruby installed

ciaranmcnulty:
@aurelien-reeves thanks for doing this!

Yes I see the issue. I could re-implement the file splitter in bash perhaps, to remove the need for PHP in the build script (the CS fixer could be removed)

aurelien-reeves:
> @aurelien-reeves thanks for doing this!
> 
> Yes I see the issue. I could re-implement the file splitter in bash perhaps, to remove the need for PHP in the build script (the CS fixer could be removed)

We still need php to run the phpunit tests so I'll add it to the build image 🙂

For cs fixer and all the tools of that kind, it's up to you. At the moment you are the owner of the php implementation 😁

ciaranmcnulty:
What's the split on other implementations between build + test? I would imagine we can use phpunit in the PHP image (and in future will want to with different PHP versions), but the code generation can happen in the main build image?

aurelien-reeves:
At the moment, the code generation actually happens on the developer computer and is pushed in the repo.
And yes, it is also built as part of the CI but only because of the target dependencies in the makefile.

Thus in the CI, the build and the tests are part of the same jobs too. Thus when it is needed, we use the cucumber-buid docker image which contains all development language we support.

This is not perfect, and this is done differently in polyglot repos. There is more decoupling between build and tests when it is needed. But as part of the monorepo, there is a single `make` command to prepare, build and test a package.

Does that answer your question? 😅

.templates/php/default.mk
```diff
@@ -0,0 +1,51 @@
...
+
+default: .tested
+.PHONY: default
+
+pre-release: update-version update-dependencies
+.PHONY: pre-release
+
+update-version:
+ifdef NEW_VERSION
+# TODO: something here?
```
ciaranmcnulty, line 20: Nothing is required here; Composer takes its versioning from the release tags so nothing in the repo needs the version number

.templates/php/default.mk
```diff
@@ -0,0 +1,51 @@
...
+endif
+.PHONY: update-version
+
+update-dependencies:
```
ciaranmcnulty, line 24: I imagine this should be `composer update`

We currently aren't committing `composer.lock` though so I'm not sure it'd be doing anything

.templates/php/default.mk
```diff
@@ -0,0 +1,51 @@
...
+.PHONY: update-dependencies
+
+publish:
+# TODO: how to publish?
```
ciaranmcnulty, line 28: Publishing = tagging a release, so the subrepo tool will be handling the publishing.

.templates/php/default.mk
aurelien-reeves: Great
However the `update-version` target still needs to be there to be compatible with other makefiles

.templates/php/default.mk
```diff
@@ -0,0 +1,51 @@
...
+post-release:
+# no-op
+.PHONY: post-release
+
+clean:
+	rm -rf vendor composer.lock
+.PHONY: clean
+
+.tested: .deps .codegen $(PHP_SOURCE_FILES)
+	vendor/bin/phpunit
```
ciaranmcnulty, line 40: This should also run psalm (static analysis tester)

.templates/php/default.mk
aurelien-reeves: Can be fine-tune at the first release of any php module

aurelien-reeves:
I have a commit ready to be pushed once the new build image is ready :D

aslakhellesoy:
Hi @ciaranmcnulty,

Aslak Hellesøy
Creator of Cucumber

PR #1888, opened 2022-01-26
Remove subrepo sync for html-formatter/javascript

aurelien-reeves:
In order to start moving out html-formatter from the monorepo, the synchronization of its javascript subrepo is removed. That way we can use the already-existing https://github.com/cucumber/html-formatter repository

### DIFF
--- a/html-formatter/javascript/.subrepo
+++ b/html-formatter/javascript/.subrepo
@@ -1,1 +1,0 @@
-cucumber/html-formatter


aurelien-reeves:
I am ready to start working on the migration
I already have a branch ready to be forced-pushed to main once this PR is merged and if there is no strong objection: https://github.com/cucumber/html-formatter

Issue #1724, opened 2021-09-02
Split monorepo up into several repos


aslakhellesoy:
**Is your feature request related to a problem? Please describe.**

There are several problems with the `cucumber/common` monorepo:

* Building all the 50+ packages in the repo takes ~1h (both locally and in CI)
* The build system is complex, brittle and hard to maintain
* Newcomers find the size of the repo and complexity of the build process intimidating and hard to navigate

**Describe the solution you'd like**

I want to split `cucumber/common` up into multiple repos:

* Polyglot repos (with `java`, `javascript`, `ruby` etc subdirectories)
  * `cucumber/cucumber-expressions`
  * `cucumber/tag-expressions`
  * `cucumber/gherkin`
  * etc.
* Single implementation repos
  * `cucumber/language-service`
  * `cucumber/language-server`
  * `cucumber/vscode`
  * `cucumber/monaco`
  * `cucumber/react`
  * etc.

The current build system has complex functionality that we'd have to replace:

* Consistent build configs
  * Extract to `cucumber/eslint`, `cucumber/tsconfig` etc. git repos.
* Release scripts (Update version numbers in descriptors/changelogs, package and publish)
  * Extract to bash scripts in a `cucumber/release` git repo and use from Docker when doing a release
* Update version numbers in dependent libraries after a release
  * Rely on WhiteSource Renovate to do this 

Also see discussion in [Slack](https://cucumberbdd.slack.com/archives/C62D0FK0E/p1630500343020600)

**Describe alternatives you've considered**

We could probably push ahead with #1720 and make the current monorepo serial build run in a few seconds (by leveraging a cache in the cloud), but the build process would still be complex and brittle. Newcomers would still be intimidated by size and complexity of this huge repo.

**Additional context**

In 2015 the Cucumber implementations had diverged and behaved inconsistently. Each release made them more inconsistent. To mitigate this we decided to bring all the Gherkin implementations into one repository, using a shared acceptance test suite.

This worked well, so we continued with the same approach for new libraries such as Cucumber Expressions and Tag Expressions - in the same repo.
 
Building and in particular _releasing_ libraries in 10 or so languages is complicated, so we built an "orchestration" build system with Make that makes the build process consistent across the increasing number of libraries.

Fast forward six years, and we have a build system with fangs, tentacles and worts. The build system wasn't designed with parallelism in mind, which is why it takes 1h.

# TODO

- [x] cucumber-expressions
- [ ] tag-expressions
- [ ] ...

[...]

PR #1896, opened 2022-02-09
Remove html-formatter from monorepo

aurelien-reeves:
## Summary

The html-formatter now lives in its own repository: https://github.com/cucumber/html-formatter

- [x] The compatibility kit still need some updates that will then be propagate to other packages which depends on the html-formatter through the CCK

## Motivation and Context

Part of #1724


aslakhellesoy (approved):
LGTM!

PR #1911, opened 2022-02-21
Gherkin PHP implementation

ciaranmcnulty:
This is passing the tests and parsing Gherkin the same as the other implementations.

In most places I've stuck as close as reasonable to the Java implementation, so have requested review from people recently working on Java too

Todo (pre-merge, need help)
- [x]  Check the Makefile setup is ok
- [x]  Add parallel job for gherkin-php

Todo (post-merge):
*  Separate sub-repo
*  Register package with Packagist


gherkin/php/composer.json
```diff
@@ -0,0 +1,23 @@
...
+    "type": "library",
+    "autoload": {
+        "psr-4": {
+            "Cucumber\\Gherkin\\": ["src/", "src-generated/"]
+        }
+    },
+    "require": {
+        "php": "^8.1",
+        "ext-mbstring": "*",
+        "cucumber/messages": "dev-main"
```
ciaranmcnulty, line 15: This needs to reflect the most recent release, in tagged versions. How does that hook in to the build process?

gherkin/php/composer.json
```diff
@@ -0,0 +1,29 @@
...
+    "description": "Gherkin parser",
+    "author": "Cucumber Limited <cukes@googlegroups.com>",
+    "license": "MIT",
+    "type": "library",
+    "autoload": {
+        "psr-4": {
+            "Cucumber\\Gherkin\\": ["src/", "src-generated/"]
+        }
+    },
+    "repositories": [
```
ciaranmcnulty, line 12: This needs to be removed for a release

gherkin/php/src/AstNode.php
```diff
@@ -0,0 +1,99 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Cucumber\Gherkin;
+
+use Cucumber\Gherkin\Parser\RuleType;
+use Cucumber\Gherkin\Parser\TokenType;
+use LogicException;
```
ciaranmcnulty, line 9: This is unused and can be removed

gherkin/php/Makefile
```diff
@@ -0,0 +1,68 @@
...
+	for file in build/Generated**; do mkdir -p src-generated/$$(head -n 1 $$file | sed 's/[^/]*.php$$//'); done
+	for file in build/Generated**; do tail -n +2 $$file > src-generated/$$(head -n 1 $$file); rm $$file; done
+	vendor/bin/php-cs-fixer --diff fix src-generated
+
+clean-build:
+	rm -rf build/classes.php
+	rm -rf src-generated/*
+
+clean:
+	rm -rf .compared acceptance
```
ciaranmcnulty, line 59: I'm not sure how to make this also call the parent `clean`?

gherkin/php/src/AstNode.php
```diff
@@ -0,0 +1,99 @@
...
+        return $items;
+    }
+
+    /**
+     * @template S of object
+     *
+     * @param class-string<S> $expectedType
+     * @param S|null $defaultValue
+     *
+     * @psalm-return S|null
```
ciaranmcnulty, line 57: I only returns null if $defaultValue is null, so this can probably be a conditional type annotation and remove some null checks elsewhere

gherkin/php/src/GherkinParser.php
```diff
@@ -0,0 +1,112 @@
...
+use Generator;
+
+final class GherkinParser
+{
+    private readonly PickleCompiler $pickleCompiler;
+    private readonly IdGenerator $idGenerator;
+
+    /**
+     * @param bool $predictableIds Ignored if IdGenerator is provided
+     * @param bool $includeSource
```
ciaranmcnulty, line 27: Redundant params here can be removed (but add some description of the method as it's public-facing)

gherkin/php/src/GherkinParser.php
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * @param iterable<Source> $sources
+     *
+     * @return Generator<Envelope>
+     */
+    public function parse(iterable $sources): Generator
+    {
+        foreach ($sources as $source) {
```
ciaranmcnulty, line 62: Should probably check the type of Source for users who don't validate the types

gherkin/php/src/Parser/TokenRuleTypesCastTrait.php
```diff
@@ -0,0 +1,27 @@
...
+        $tokenName = $tokenType->name;
+
+        if ($tokenName != TokenType::None->name) {
+            $tokenName = '_' . $tokenName;
+        }
+
+        $ruleType = constant(RuleType::class . '::' . $tokenName);
+
+        if (!$ruleType instanceof RuleType) {
+            throw new LogicException('Could not create RuleType from  TokenType::' . $tokenName);
```
ciaranmcnulty, line 22: remove this extra space

gherkin/php/src/Parser/TokenRuleTypesCastTrait.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Cucumber\Gherkin\Parser;
+
+use SebastianBergmann\Type\LogicException;
```
ciaranmcnulty, line 7: Wrong type of LogicException! won't be present on prod

gherkin/php/src/Parser/TokenScanner.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
```
ciaranmcnulty, line 3: Not needed on an interface

gherkin/php/src/ParserException/UnexpectedEofException.php
```diff
@@ -0,0 +1,27 @@
...
+namespace Cucumber\Gherkin\ParserException;
+
+use Cucumber\Gherkin\ParserException;
+use Cucumber\Gherkin\Token;
+
+final class UnexpectedEofException extends ParserException
+{
+    public function __construct(
+        public readonly Token $receivedToken,
+        /**
```
ciaranmcnulty, line 14: This should be `@param` instead for Psalm to see it

gherkin/php/src/PickleCompiler.php
```diff
@@ -0,0 +1,255 @@
...
+        $feature = $gherkinDocument->feature;
+        $language = $feature->language;
+
+        yield from $this->compileFeature($feature, $language, $uri);
+    }
+
+    /**
+     * @return Generator<Pickle>
+     */
+    private function compileFeature(Feature $feature, string $language, string $uri)
```
ciaranmcnulty, line 53: needs a return type

gherkin/php/src/StringGherkinLine.php
```diff
@@ -0,0 +1,146 @@
...
+        return StringUtils::symbolCount($this->trimmedLineText) === 0;
+    }
+
+    public function startsWith(string $string): bool
+    {
+        return StringUtils::startsWith($this->trimmedLineText, $string);
+    }
+
+    /** @return list<GherkinLineSpan> */
+    public function getTableCells(): array
```
ciaranmcnulty, line 66: Fairly sure this should be a regex, have another try and writing one

gherkin/php/src/StringTokenScanner.php
```diff
@@ -0,0 +1,37 @@
...
+final class StringTokenScanner implements TokenScanner
+{
+    private int $lineNumber = 0;
+
+    /** @var list<string> */
+    private array $lines;
+
+    public function __construct(string $source)
+    {
+        $this->lines = explode("\n", $source);
```
ciaranmcnulty, line 25: Rather than explode, we should store the string and consume the line each time read() is called

gherkin/php/src/StringUtils.php
```diff
@@ -0,0 +1,77 @@
...
+namespace Cucumber\Gherkin;
+
+/**
+ * Keeps common string operations in one place using correct unicode functions
+ * (and normalises naming with other implementations)
+ */
+final class StringUtils
+{
+    private const DEFAULT_ENCODING = 'utf-8';
+    private const WHITESPACE_PATTERN = "[ \\t\\n\\x0B\\f\\r\\x85\\xA0]+";
```
ciaranmcnulty, line 14: Add comment explaining this comes from the Java version...

gherkin/php/src/Token.php
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * @psalm-assert-if-true null $this->line
+     */
+    public function isEof(): bool
+    {
+        return is_null($this->line);
+    }
+
+    public function detach(): void
```
ciaranmcnulty, line 37: Find out why this is a no-op

gherkin/php/src/Token.php
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    public function getLocation(): Location
+    {
+        return $this->location;
+    }
+
+    public function getTokenValue(): string
+    {
+        return $this->isEof() ? 'EOF' : $this->line->getLineText(-1);
```
ciaranmcnulty, line 48: Make the magic string a constant

gherkin/php/src/Token.php
ciaranmcnulty: From Slack, we aren't sure what it does and if it can be removed, but @gasparnagy may know!

gherkin/php/Makefile
aurelien-reeves: I think it will call the "parent" clean target if you just add dependencies without altering the target, like with messages: https://github.com/cucumber/common/blob/main/messages/php/Makefile#L5

gherkin/php/composer.json
aurelien-reeves: Such thing is done with other implementation in the pre-release make target.
For example with ruby, we comment some lines in the Gemfile: https://github.com/cucumber/common/blob/main/gherkin/ruby/default.mk#L48
Or with Go, we remove some "replace" instructions: https://github.com/cucumber/common/blob/main/gherkin/go/default.mk#L86

aurelien-reeves:
Impressive work @ciaranmcnulty 😮

I had a first look
Reviewing it entirely will be difficult. I will give it a deeper look, but as long as relevant tests have been added and are passing, I guess we can give it a go

gherkin/php/src/Token.php
gasparnagy: I don't remember fully, but it does not seem to be in use. I remember that I did a token/line implementation that did nor require to split the feature file text into multiple strings (memory optimization) and for some reason it was important to that implementation to know when the token is not used anymore. But I can't really imagine why was that needed and my optimized implementation seem not to be used anymore, probably deleted. 

I guess the parser generator will generate this call and tells to token when it is not in use anymore, but feel free to not do anything there.

ciaranmcnulty: Thanks, removed it! We can always add back in :)

gherkin/php/composer.json
ciaranmcnulty: Great, I've added a target using `jq` to remove this block during `pre-release`

aurelien-reeves:
I have created the gherkin-php repository
You can now add the .subrepo file in gherkin/php

.templates/php/default.mk
```diff
@@ -33,20 +48,21 @@ post-release:
 .PHONY: post-release
```
aurelien-reeves, line 48: I guess here we should revert what has been done by `remove-local-repository`

ciaranmcnulty: ok, something like `restore-local-repository`

ciaranmcnulty:
I need to add a `pre-subrepo-push` and `post-subrepo-push` into the language templates, invoked from the main subprepo target, and then we can invoke the `remove-local-repository` and `restore-local-repository` in there

gherkin/php/src/GherkinDocumentBuilder.php
```diff
@@ -0,0 +1,371 @@
...
+    {
+        return $node->getitems(Step::class, RuleType::Step);
+    }
+
+    /** @return list<TableRow> */
+    private function getTableRows(AstNode $node): array
+    {
+        $rows = [];
+
+        foreach ($node->getTokens(TokenType::TableRow) as $token) {
```
theofidry, line 314: a lil' array_map here instead?

gherkin/php/src/GherkinDocumentBuilder.php
```diff
@@ -0,0 +1,371 @@
...
+            $rows[] = new TableRow($this->getLocation($token, 0), $this->getCells($token), $this->idGenerator->newId());
+        }
+
+        $this->ensureCellCount($rows);
+
+        return $rows;
+    }
+
+    /** @param list<TableRow> $rows */
+    private function ensureCellCount(array $rows): void
```
theofidry, line 324: let's make those methods static?

gherkin/php/src/Parser/ParserContext.php
```diff
@@ -0,0 +1,28 @@
...
+use Cucumber\Gherkin\Token;
+use Cucumber\Gherkin\TokenMatcher;
+
+final class ParserContext
+{
+    public function __construct(
+        public readonly TokenScanner $tokenScanner,
+        public readonly TokenMatcher $tokenMatcher,
+
+        /**
```
theofidry, line 17: I would move them to above the constructor "as usual"

gherkin/php/src/PickleCompiler.php
```diff
@@ -0,0 +1,255 @@
...
+     * @return list<PickleTag>
+     */
+    private function pickleTags(array $tags): array
+    {
+        return array_map(fn ($t) => $this->pickleTag($t), $tags);
+    }
+
+    private function pickleTag(Tag $tag): PickleTag
+    {
+        return new PIckleTag($tag->name, $tag->id);
```
theofidry, line 253:
```suggestion
        return new PickleTag($tag->name, $tag->id);
```

gherkin/php/src/StringGherkinLine.php
```diff
@@ -0,0 +1,141 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Cucumber\Gherkin;
+
+final class StringGherkinLine implements GherkinLine
+{
+    // TODO: set this to 0 when/if we change to 0-indexed columns
```
theofidry, line 9: TODO?

gherkin/php/src/StringUtils.php
```diff
@@ -0,0 +1,79 @@
...
+{
+    private const DEFAULT_ENCODING = 'utf-8';
+
+    // same patterns as the Java implementation
+    private const WHITESPACE_PATTERN = "[ \\t\\n\\x0B\\f\\r\\x85\\xA0]+";
+    private const WHITESPACE_PATTERN_NO_NEWLINE = "[ \\t\\x0B\\f\\r\\x85\\xA0]+";
+
+    public static function symbolCount(string $string): int
+    {
+        return mb_strlen($string, self::DEFAULT_ENCODING);
```
theofidry, line 21: to check if the mb polyfill is there or the extension required

gherkin/php/src/TokenFormatter.php
```diff
@@ -0,0 +1,26 @@
...
+
+        return sprintf(
+            "(%s:%s)%s:%s/%s/%s",
+            $token->getLocation()->line,
+            $token->getLocation()->column,
+            $token->matchedType?->name ??  '',
+            $token->matchedKeyword ?? '',
+            $token->matchedText ?? '',
+            $token->matchedItems === null ? ''
+                : join(',', array_map(fn ($o) => $o->column . ':' . $o->text, $token->matchedItems))
```
theofidry, line 23: `$o`?

gherkin/php/src/TokenMatcher.php
```diff
@@ -0,0 +1,229 @@
...
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    public function match_TableRow(Token $token): bool
+    {
+        if ($token->line?->startsWith(GherkinLanguageConstants::TABLE_CELL_SEPARATOR)) {
```
theofidry, line 139: might be worth inverting the condition here?

gherkin/php/src/TokenMatcher.php
```diff
@@ -0,0 +1,229 @@
...
+            // open
+            ? $this->_match_DocStringSeparator($token, GherkinLanguageConstants::DOCSTRING_SEPARATOR, true)
+            || $this->_match_DocStringSeparator($token, GherkinLanguageConstants::DOCSTRING_ALTERNATIVE_SEPARATOR, true)
+            // close
+            : $this->_match_DocStringSeparator($token, $this->activeDocStringSeparator, false);
+    }
+
+    private function _match_DocStringSeparator(Token $token, string $separator, bool $isOpen): bool
+    {
+        if ($token->line?->startsWith($separator)) {
```
theofidry, line 173: same here and a few other places down there

gherkin/php/tests/Parser/RuleTypeTest.php
```diff
@@ -0,0 +1,27 @@
...
+    /** @dataProvider tokenCaseProvider */
+    public function testItCanCastFromTokens(TokenType $tokenType): void
+    {
+        self::assertInstanceOf(RuleType::class, RuleType::cast($tokenType));
+    }
+
+    /**
+     * @return Generator<string,array{0:TokenType}>
+     */
+    public function tokenCaseProvider(): Generator
```
theofidry, line 21: I highly recommend to make PHPUnit data providers `static`. This avoids ever trying to use `$this` in it which would be a nasty surprise (the instance of the test case inside the provider is different than the one executing the test)

theofidry, line 21: I also tend to go with `iterable` rather than `Generator` as it lets you switch to `array` without further changes

gherkin/php/tests/ParserException/NoSuchLanguageExceptionTest.php
```diff
@@ -0,0 +1,18 @@
...
+namespace Cucumber\Gherkin\ParserException;
+
+use Cucumber\Gherkin\Location;
+use PHPUnit\Framework\TestCase;
+
+final class NoSuchLanguageExceptionTest extends TestCase
+{
+    public function testItHasSensibleMessage(): void
+    {
+        $e = new NoSuchLanguageException('en', new Location(1, 2));
```
theofidry, line 14:
```suggestion
        $exception = new NoSuchLanguageException('en', new Location(1, 2));
```

theofidry, line 14:
```suggestion
        $exception = new NoSuchLanguageException('en', new Location(1, 2));
```

gherkin/php/tests/ParserException/UnexpectedTokenExceptionTest.php
```diff
@@ -0,0 +1,62 @@
...
+{
+    public function testMessageSpecifiesWhichTokenWasUnexpected(): void
+    {
+        $line = $this->createMock(GherkinLine::class);
+        $line->method('getLineText')->willReturn('Baz');
+
+        $exception = new UnexpectedTokenException(
+            new Token($line, new Location(10, 10)),
+            ['Foo', 'Bar'],
+            'stateComment'
```
theofidry, line 22: I recommend the following setting for PHP-CS-Fixer:
```php
        'trailing_comma_in_multiline' => [
            'after_heredoc' => true,
            'elements' => [
                'arrays',
                'arguments',
                'parameters',
            ],
        ],
```

for adding trailing commas where possible

gherkin/php/src/StringUtils.php
```diff
@@ -0,0 +1,79 @@
...
+
+declare(strict_types=1);
+
+namespace Cucumber\Gherkin;
+
+/**
+ * Keeps common string operations in one place using correct unicode functions
+ * (and normalises naming with other implementations)
+ */
+final class StringUtils
```
theofidry, line 11: `Str`?

theofidry (approved):
👍 really cool!

Most if not all my comments are nitpicks. Please feel free to ignore/resolve them without having to justify yourself

.templates/php/default.mk
```diff
@@ -33,20 +48,21 @@ post-release:
...
+.deps: vendor .codegen
 	touch $@
 
 .codegen:
 	touch $@
 
-composer.lock: composer.json
+vendor: composer.json
 	composer install
-	touch $@
```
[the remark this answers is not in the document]
ciaranmcnulty: Because the make target is `vendor` and it's created by the commands that are run

gherkin/php/composer.json
```diff
@@ -0,0 +1,32 @@
...
+            "url": "../../messages/php"
+        }
+    ],
+    "require": {
+        "php": "^8.1",
+        "ext-mbstring": "*",
+        "cucumber/messages": "dev-main||^17.0"
+    },
+    "require-dev": {
+        "phpunit/phpunit": "^9.5",
```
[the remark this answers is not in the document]
ciaranmcnulty: It's on my list ;) Would like to talk to you actually about how to get coverage from the end-to-end black box tests (that run bin/gherkin)

gherkin/php/src-generated/Parser.php
```diff
@@ -0,0 +1,4047 @@
...
+            function () use ($context, $token) {
+                return $context->tokenMatcher->match_Other($token);
+            },
+            false
+        );
+    }
+
+    private function matchToken(int $state, Token $token, ParserContext $context): int
+    {
+        return match ($state) {
```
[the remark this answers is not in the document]
ciaranmcnulty: I did think about this yes, but [there's a condition for one of the states](https://github.com/cucumber/common/blob/d91fc0ff03fdb196247164609e8b2a900763e668/gherkin/php/gherkin-php.razor#L130-L134) and I don't want to move that from the codegen into the PHP code template

gherkin/php/tests/Parser/RuleTypeTest.php
ciaranmcnulty: Nothing depends on this, so switching to `array` later is trivial, no?

gherkin/php/tests/ParserException/UnexpectedTokenExceptionTest.php
ciaranmcnulty: Thanks, I've been trying to do that so it'll be good to use the rule 👍

gherkin/php/src/StringUtils.php
ciaranmcnulty: Is that a specific library you're recommending, or a naming suggestion? If it's the latter I've stuck to what it's called in Gherkin Java to make it more comprehensible for people working across languages

.templates/php/default.mk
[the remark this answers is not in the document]
ciaranmcnulty: It's not needed as the commands create the `vendor` folder

gherkin/php/src/GherkinDialectProvider.php
```diff
@@ -0,0 +1,65 @@
...
+        private readonly string $defaultDialectName = 'en'
+    ) {
+        try {
+            /**
+             * Here we force the type checker to assume the decoded JSON has the correct
+             * structure, rather than validating it. This is safe because it's not dynamic
+             *
+             * @var non-empty-array<non-empty-string, Dialect> $data
+             */
+            $data = json_decode(file_get_contents(self::JSON_PATH), true, flags: JSON_THROW_ON_ERROR);
```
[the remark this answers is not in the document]
ciaranmcnulty: As we maintain the JSON (shared with other cucumbers) I'm not too concerned about that because it'll break _everything_ if someone borks the JSON

gherkin/php/src/GherkinDocumentBuilder.php
```diff
@@ -0,0 +1,371 @@
...
+
+    public function startRule(RuleType $ruleType): void
+    {
+        array_push($this->stack, new AstNode($ruleType));
+    }
+
+    public function endRule(RuleType $ruleType): void
+    {
+        $node = array_pop($this->stack);
+        if ($transformedNode = $this->getTransformedNode($node)) {
```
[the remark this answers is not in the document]
ciaranmcnulty: Yeah probably, if I'm expecting people less familiar with PHP to want to change stuff 👍

gherkin/php/src/StringUtils.php
ciaranmcnulty: Yep it's in composer.json

gherkin/php/src/GherkinDialectProvider.php
```diff
@@ -0,0 +1,65 @@
...
+        return array_keys($this->DIALECTS);
+    }
+
+    /**
+     * @param non-empty-string $language
+     */
+    public function getDialect(string $language, ?Location $location): GherkinDialect
+    {
+        if (!isset($this->DIALECTS[$language])) {
+            throw new NoSuchLanguageException($language, $location);
```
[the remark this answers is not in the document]
ciaranmcnulty: I want naming to be the same with the Java exceptions, where there's a suffix

.templates/php/default.mk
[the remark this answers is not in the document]
theofidry: yes but sometimes the command does not correctly refreshes the timestamp of the directory hence the necessity of the touch since makefiles relies on the timestamp

gherkin/php/src/StringUtils.php
theofidry: the latter so 👍

gherkin/php/src/GherkinDocumentBuilder.php
```diff
@@ -0,0 +1,371 @@
...
+    public function __construct(
+        private readonly string $uri,
+        private readonly IdGenerator $idGenerator
+    ) {
+        $this->reset($uri);
+    }
+
+    public function build(Token $token): void
+    {
+        if (is_null($token->matchedType)) {
```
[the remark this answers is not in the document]
ciaranmcnulty: You want Yoda style? 😱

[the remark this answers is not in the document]
theofidry: i have the habit to so i dont care, its enforced by the CS tool anyway :P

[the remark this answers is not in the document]
theofidry: i have the habit to so i dont care, its enforced by the CS tool anyway :P

gherkin/php/src/GherkinDocumentBuilder.php
```diff
@@ -0,0 +1,371 @@
...
+            case RuleType::ExamplesDefinition:
+
+                $examplesNode = $node->getSingle(AstNode::class, RuleType::Examples);
+                if (is_null($examplesNode)) {
+                    return null;
+                }
+                $examplesLine = $examplesNode->getToken(TokenType::ExamplesLine);
+                /** @var list<TableRow>|null $rows */
+                $rows = $examplesNode->getSingleUntyped(RuleType::ExamplesTable);
+                $tableHeader = is_array($rows) && count($rows) ? $rows[0] : null;
```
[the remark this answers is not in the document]
Lewiscowles1986: Surely to be consistent with the other feedback you would use `0 === count($rows)` ?

Why yoda in some places and not others?

Also I think this inverted the logic. `&& count($countable)` is checking `!empty($countable)` whereas `0 === count($rows)` is checking if something is empty.

gherkin/php/src/GherkinDocumentBuilder.php
```diff
@@ -0,0 +1,371 @@
...
+
+    private function currentNode(): AstNode
+    {
+        return $this->stack[array_key_last($this->stack)];
+    }
+
+    /**
+     * @return object|string|list<object>|null
+     */
+    private function getTransformedNode(AstNode $node): object|string|array|null
```
Lewiscowles1986, line 100: At 189 lines. Would it be terrible to pull all the internals of each case, out into a named, possibly private methods? Not asking for the change, but trying to understand why you did not?

Was it the rule of three (these each, presumably only being used here?)

gherkin/php/src/GherkinDocumentBuilder.php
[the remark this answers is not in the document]
theofidry: @Lewiscowles1986 IMO the consistency should be enforced by PHP-CS-Fixer.

In PHP `empty` is to be avoided as `empty('0') === true`

gherkin/php/src/TokenMatcher.php
ciaranmcnulty: Maybe later; I'd want to change it in the Java version too and don't have the confidence yet!

gherkin/php/src/GherkinDocumentBuilder.php
ciaranmcnulty: This is a good point, I've split it out consequently can use a `match`

gherkin/php/src/GherkinDocumentBuilder.php
ciaranmcnulty: I don't think I will here, I'm not 100% on the pattern (can change later)

.templates/php/default.mk
```diff
@@ -7,6 +7,15 @@ PHP_SOURCE_FILES = $(shell find . -name "*.php")
...
 
+BERP_VERSION = 1.3.0
+BERP_GRAMMAR = gherkin.berp
+
+define berp-generate-parser =
+-! dotnet tool list --tool-path /usr/bin | grep "berp\s*$(BERP_VERSION)" && dotnet tool update Berp --version $(BERP_VERSION) --tool-path /usr/bin
+berp -g $(BERP_GRAMMAR) -t $< -o $@ --noBOM
+endef
+
+
```
aurelien-reeves, line 18: Could that be part of the `gherkin/php/Makefile` rather than in all `php/default.mk`?

gherkin/php/composer.json
```diff
@@ -0,0 +1,32 @@
...
+        "vimeo/psalm": "^4.20",
+        "friendsofphp/php-cs-fixer": "^3.5",
+        "psalm/plugin-phpunit": "^0.16.1"
+    },
+    "repositories": [
+        {
+            "type": "path",
+            "url": "../../messages/php"
+        }
+    ]
```
aurelien-reeves, line 31: What is the meaning of this?

messages/php/scripts/update-gherkin-dependency.php
```diff
@@ -0,0 +1,52 @@
...
+
+// like ^21.0 (same as >=21.0.0 <22.0.0)
+$newDependency = ',^'.$matches['major'].'.0';
+
+if (str_contains($dependencyString, $newDependency)) {
+    fwrite(STDERR, 'Nothing to update, already depends on ' . $newDependency . "\n");
+    $newDependency = '';
+}
+
+$newDependencyString = str_replace('dev-main||', '', $dependencyString) . $newDependency;
```
aurelien-reeves, line 48: If I understand well, once we have release `messages`, we update the dependency in gherkin to make it dependent on the version we just released, and remove the "dev-main" part.

Could it be possible to do that also in `pre-release` target for `gherkin/php`?

gherkin/php/composer.json
ciaranmcnulty: This means when working on `cucumber/common` and typing `composer install`, it will symlink the messages into the `vendor` folder rather than downloading the package.

This is only when working on a checkout of `common`, not if requiring `cucumber/gherkin` as a package in some other project

.templates/php/default.mk
ciaranmcnulty: It's in the default.mk for other languages too; I'm not clear why TBH

messages/php/scripts/update-gherkin-dependency.php
aurelien-reeves: Maybe not the version number, but at least removing the `dev-main` thing

And I do not suggest not to do it anymore in the `post-release` of `messages` ;)

ciaranmcnulty: you're right removing dev-main is better there 👍

ciaranmcnulty: ok removed that substitution here and added a new pre-release hook in the gherkin makefile

aurelien-reeves (approved):
👍

PR #1913, opened 2022-03-02
Remove gherkin-streams from the monorepo

aurelien-reeves:
## Summary

Removes the gherkin-streams from the monorepo as it is now located at [github.com/cucumber/gherkin-streams](https://github.com/cucumber/gherkin-streams)


fake-cucumber/javascript/.mocharc.local.json
```diff
@@ -0,0 +1,6 @@
+{
```
davidjgoss, line 1: So this is the same as `.mocharc.local.json` but without the require of tsconfig-paths - does it break without this? Does it also mean you have to build dependent modules before running tests now?

davidjgoss: (Not the end of the world if so, as these will presumably be coming out of the monorepo soon enough, but curious.)

aurelien-reeves: Yes, tsconfig-paths does break the build.

Actually if I recall correctly,we already need to build JavaScript modules before anything else

davidjgoss (approved):

aurelien-reeves:
Actually it seems tsconfig-paths was not required as part of the mocha configuration at least

Issue #1139, opened 2020-07-09
Ability to name the hooks


suswariadla:
Hi,
I don't see there is an ability to add names to the Before and After hooks rightnow. 
Can the team consider adding the ability to name Before and After hooks so that they appear in the json reporter.

davidjgoss:
@suswariadla would you mind elaborating on your use case a little? Interested to know the motivation and whether just `attach`ing some text would be sufficient.

suswariadla:
<img width="413" alt="Screen Shot 2020-08-14 at 10 16 18 AM" src="https://user-images.githubusercontent.com/47369252/90260010-eb827c00-de18-11ea-8bf0-f3aa2824f823.png">

Hey @davidjgoss  Thanks for Replying.
You see in the image, that is a shot from jenkins cucumber html reporter, how the statement looks for the Before hook. After taking a look in to json file which is the source for jenkins cucumber html reporter we found that the json file does not capture the Before hook name and infact the cucumberJs does not provide an option to define the hook name in its syntax.
So we request the flexibility to name the hooks in cucumberJs

aslakhellesoy:
I think this could be useful for reporting purposes. Would you be interested in helping us implement this @suswariadla?

Since the JSON format is now [deprecated in favour of messages](https://github.com/cucumber/cucumber/tree/master/messages#why-cucumber-messages), this is something that would have to be added to the message protocol.

We probably wouldn't update the existing native JSON formatters, but we could update the [standalone json formatter](https://github.com/cucumber/cucumber/tree/master/json-formatter) instead.

Cucumber implementations would have to extend the hook API. The syntax for various languages is already somewhat constrained because it's already used to pass [tag expressions](https://cucumber.io/docs/cucumber/api/#tag-expressions).

Proposed syntax:

## JavaScript

```javascript
Before({name: "Take out the rubbish"}, function () {
})
```

## Java

```java
@Before(name = "Take out the rubbish")
public void doSomething(){
    // Do something before each scenario
}
```

## Ruby

```ruby
Before(name: 'Take out the rubbish') do
end
```

suswariadla:
@aslakhellesoy Sure, I would be happy to implement it.

aslakhellesoy:
Thanks @suswariadla - much appreciated!

Here is a rough outline of the steps that should be taken:

[...]

PR #1914, opened 2022-03-02
add name property to hook schema

davidjgoss:
## Summary

- Add `name` as an optional field on the Hook schema
- Teach fake-cucumber to populate that field
- Add case to the compatibility kit

## Motivation and Context

First part of https://github.com/cucumber/common/issues/1139.

## How Has This Been Tested?

via the compatibility kit.

## Types of changes

## Checklist:


aslakhellesoy:
Rather than trying to make the JSON formatter work, I think we should delete it.

It was created as a fallback solution when we decided to remove the native JSON formatters. Now that we're keeping them, this tool is obsolete.

davidjgoss:
@aslakhellesoy well okay then! I'll raise a separate PR for that.

aurelien-reeves (approved):
Looks good 🥳

PR #1916, opened 2022-03-02
debt: delete json formatter

### LINKED ISSUE #1915: Remove standalone json formatter
We don't need it anymore - Cucumbers are keeping their native JSON formatters.

Its presence is currently holding up https://github.com/cucumber/common/pull/1914 - let's delete it.

davidjgoss:
## Summary

Delete `json-formatter` and its references and usages in the monorepo.

## Motivation and Context

Fixes #1915.

## How Has This Been Tested?

If the build passes, we're good.


aurelien-reeves (commented):
I would still wait a little bit before removing the standalone json-formatter:

- it may be used by users so we have to keep a safe place for the binaries, and announce and document the removal
- some other packages tests still rely on it to generate some test data

davidjgoss:
> it may be used by users so we have to keep a safe place for the binaries, and announce and document the removal

Good point, perhaps we should break it out to a repo and then archive that before deleting here?

> some other packages tests still rely on it to generate some test data

The only one I found was `json-to-messages` and I've refactored that dependency out; everything's green. Are there more I'm not seeing?

aurelien-reeves:
> Good point, perhaps we should break it out to a repo and then archive that before deleting here?
> 

Yes, I would go for something like this.

> The only one I found was `json-to-messages` and I've refactored that dependency out; everything's green. Are there more I'm not seeing?

Oh, well done! Sorry, I did not get that 😅
So, as long as the CI is green and not impacted anymore with this, I am ok with that point 👍

davidjgoss:
I've created a repo which has the contents of json-formatter including the history https://github.com/cucumber/json-formatter

It would probably need a little work to make it operational as a standalone package - is that worthwhile though?

aurelien-reeves:
> I've created a repo which has the contents of json-formatter including the history https://github.com/cucumber/json-formatter
> 
> It would probably need a little work to make it operational as a standalone package - is that worthwhile though?

Definitely, yes! Thank you! 🤩

badeball:
I'm a bit out of the loop regarding JSON formatting and was wondering if someone could help me catch up. I've authored / authoring a [plugin for Cypress](https://github.com/badeball/cypress-cucumber-preprocessor) to allow users to write Gherkin specs, a library which re-implements much of the behavior users are familiar with from Cucumber.

JSON reports are fairly popular and I've refactored my implementation to output messages, because that was significantly easier implementation wise, and I rely on users to use https://github.com/cucumber/json-formatter to further convert this output to JSON. Is this stand-alone formatter about to go away?

mattwynne:
Hey @badeball, thanks for getting in touch and thanks for all your hard work to bring Cucumber functionality to the Cypress community!

We originally had a vision whereby we would remove the language-specific JSON formatters from all the Cucumber implementations get people to move over to using the new NDJSON message protocol instead. We wrote this Go binary as part of that strategy so that there was a fall-back, but we got [quite a bit of pushback about this](https://github.com/cucumber/cucumber-js/issues/1250), and decided to keep the native JSON formatters for now. We've had similar issue to you with people getting false positives from their anti-virus scanners from the Go binary.

We weren't sure if anyone else was using this tool so as we work to [break up this monorepo](https://github.com/cucumber/common/issues/1724) we'd thought it was OK to just remove it and archive the code. Since we know there's at least one user we'll make sure we keep the code live, but it will start to live in this new stand-alone repo once this PR is merged.

davidjgoss:
Okay, thanks to @mattwynne's efforts on https://github.com/cucumber/json-formatter/pull/1 we now have the JSON formatter in its own repo with a green build, so this should be good to go.

mattwynne:
> Okay, thanks to @mattwynne's efforts on [cucumber/json-formatter#1](https://github.com/cucumber/json-formatter/pull/1) we now have the JSON formatter in its own repo with a green build, so this should be good to go.

I'd just like to finish the work on https://github.com/cucumber/json-formatter/pull/6 before we call this done, so that we can point everyone over to the new repo for downloading the binaries.

mattwynne:
OK, with https://github.com/cucumber/json-formatter/pull/6 and [binaries available to download](https://github.com/cucumber/json-formatter/releases/tag/v19.0.0) from that repo, I think we're good to go with this.

I wonder what we should do about the releases sitting on this repo. I've updated the [latest release](https://github.com/cucumber/common/releases/tag/json-formatter%2Fgo%2Fv19.0.0) to point to the new repo. Should we delete the assets from that release too, for clarity? What about older releases?

davidjgoss:
Going to merge this once the current run is green, as it keeps getting conflicted.

PR #1920, opened 2022-03-11
i18n-keywords: Update Portuguese on gherkin-languages.json

lucasthecodes:
## Summary
Expanding Portuguese "given" translations to enhance steps code reusability.
Also inspired on the French translations.

## Details
I have added four new translations into Portuguese for the "given" step, as further detailed in the section below. In summary: 
* "Dado que " helps in code reusability, having its correspondents in other translations;
* "Sabendo que " is inspired on the French translation, being equivalent to "Dado que ";
* "Seja " and "Sejam " are also inspired in French, being closer to the current translation.

## Motivation and Context
In practice most testers start their "given" steps in Portuguese with "Dado que" ("Given that"), what makes for more linguistically accurate statements. On the other hand, it breaks code reusabilty for different steps on different scenarios.
For example, I can have a step "I visit some page" originally written as a "when" step (to test for correct exhibition of elements when I access that page, for instance) that can be later used on a "given" step of a different scenario (from where I'll test some other functionality, say, "when I click a button"). Using the current translation ("Dado" or its declensions), that reuse is impossible without breaking the sentence. I've seen terrible workarounds for that (Portuguese grammar rip!). Adding the single new translation "Dado que" solves that.
It's worth noting the same approach has been already taken in other languages like French and Romanian. Also, as many people have already used the current translations and will keep doing so, they are not to be changed.

The other three proposed translations are inspired on the French example and suggested for completeness.  
"Sabendo que " (like the French "Sachant que " - "Knowing that") allows for the same kind of construct of "Dado que " ("Given that") but with slightly different uses, thus contributing to richer scenarios writing.
"Seja " (like the French "Soit ", meaning "Be " in the Subjunctive form - e.g.: "Seja um usuário autenticado" - "Be an authenticated user") can have a use close to the current Portuguese translations "Dado " and "Dada ".
Analogously, "Sejam " (also meaning "Be " in the Subjunctive form) would be a close equivalent to the current translations "Dados " and "Dadas ".  

The following sections don't seem to apply to this simple translation request, being then skipped.


aurelien-reeves:
Hi @lucasthecodes 

Thanks for your contribution!

I wanted to synchronize your changes as described here: https://github.com/cucumber/common/blob/main/gherkin/CONTRIBUTING.md#adding-or-updating-an-i18n-language

but I do not have permissions to push to your fork. Could you please give me permission to push to your fork or do the synchronization yourself?

aurelien-reeves:
@lucasthecodes I have run some tests, and actually I confirm they may be some issues.

First: the update you proposed put the `dado ` keyword before `dado que `. Acutally this won't work. The steps will still always require to be implemented with the `que` prefix: `Given('que ...', function () {`

Second: if we fix the order of the keyword to put the more long first: `dado que `, then `dado `, your initial suspicion is true: the step definitions in the end-user suites  will need to be updated from `Given('que ...', ` to `Given('...', ` (removing the `que` prefix).

My opinion is to properly add the `dado que ` keyword and the other ones and to inform end users through the changelog to do the update. This is an easy search&replace thing, and it would actually be a far better support for Portugese. But this is indeed a breaking change.

What is your though about that?

@aslakhellesoy @mattwynne @mpkorstanje @davidjgoss @luke-hill any though about this? (Would you need more context?)

lucasthecodes:
Thank you for your reply and tests, @aurelien-reeves .
The main purpose of the proposed changes is to expand the possibilities of scenarios creation and code reusability, not to somehow restrict what is already working.
Many companies have their test suites running with the current keywords and breaking those suites might create chaos everywhere. We don't want that.
I'll wait for any insights of the others you marked on your reply. Maybe some change on cucumber's code might allow this kind of change. If not, probably this PR is to be closed.

aslakhellesoy:
I think we should be consistent with the French dialect - listing the longest keywords first. `Dado que ` before `Dado `.

aslakhellesoy:
Slightly unrelated - I just realised that `Given`, `When`, `Then` keywords are not internationalised. Other Cucumber implementations allow this:

```typescript
Dado('que some condition', function () {
  //
})
```

See https://github.com/cucumber/cucumber-js/blob/7e1f8589cfc433c4bec470be500c8811add8c7b2/src/support_code_library_builder/index.ts#L121-L135

lucasthecodes:
Hello all, thanks for the insights.
As I stated before, I'm not proposing any breaking change. If that's the
only option, I'll close this PR.
As simple as this change can be, it'd affect code everywhere and I'm not to
be responsible for it.
Maybe your impression taking from French or other language's perspective is
that it's not much used and not too spread, the standard English words
being more common.
In Brazil things tend to be done in Portuguese, so this change would really
affect many sensitive work and create many problems.
Repeating myself, when I proposed it I thought Cucumber's keyword parser
would be able to handle that. My mistake.
Let me know if I can close this.
Thanks.

aurelien-reeves:
@aslakhellesoy could you please take a look at the conflict here?
I guess this is related to the gherkin java codegen PR. Removing `gherkin/java/src/main/resources/io/cucumber/gherkin/gherkin-languages.json` would fix the conflict easily, but I don't know if something else should be done beside that

lucasthecodes:
The message didn't go through.
Trying again...

aurelien-reeves:
Well, your PR was not only adding "Dado que" but plenty of other keyword that I guess would be welcome for Portuguese users

PR #1923, opened 2022-03-15
Remove SHA1 detail from Pickle ID description

ciaranmcnulty:
We no longer generate them this way, if we did at all (and in particular the guarantee about when it changes is wrong)


aurelien-reeves:
Nice catch.
I think we'll need to regenerate messages as some come with some documentation coming from those descriptions.
I guess we need to execute `make` from within `./messages`, and commit the resulting files

ciaranmcnulty:
Oh good point I'll do that later when less lazy

aurelien-reeves (approved):
👍
Looks good

PR #1926, opened 2022-03-18
Messages PHP - Remove dependency on UUID library

ciaranmcnulty:
We only have one major dependency but it pulls in a lot. This removes it but still uses that library in the tests to check we're generating valid IDs


messages/php/src/Id/FallbackUuidGenerator.php
```diff
@@ -0,0 +1,33 @@
...
+
+declare(strict_types=1);
+
+namespace Cucumber\Messages\Id;
+
+final class FallbackUuidGenerator implements IdGenerator
+{
+    /**
+     * Inspired by
+     * @see https://github.com/clue/reactphp-clickhouse
```
mpkorstanje, line 11: Would be better to link the UUIDv4 spec here.

messages/php/src/Id/FallbackUuidGenerator.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Cucumber\Messages\Id;
+
+final class FallbackUuidGenerator implements IdGenerator
```
mpkorstanje, line 7: Does it need to be a fallback?

messages/php/src/Id/FallbackUuidGenerator.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    /**
+     * Inspired by
+     * @see https://github.com/clue/reactphp-clickhouse
+     */
+    public function newId(): string
+    {
+        $bytes = random_bytes(16);
```
mpkorstanje, line 15: What is the quality of this random source?

messages/php/src/Id/FallbackUuidGenerator.php
```diff
@@ -0,0 +1,33 @@
...
+
+        /**
+         * set MSBs of octet 6 to 0100... to indicate v4
+         *
+         * @see https://datatracker.ietf.org/doc/html/rfc4122#section-4.1.3
+         */
+        $bytes[6] = chr(ord($bytes[6]) & 0b0000_1111 | 0b0100_0000);
```
mpkorstanje, line 22: This `chr` and `ord` looks expensive. I think that with two random longs you can achieve the same.

mpkorstanje: Nevermind, PHP doesn't have well defined numerical sizes.

I imagine there is a faster way but I don't know enough to suggest one.

messages/php/src/Id/FallbackUuidGenerator.php
Adirelle: According to [PHP manual](https://www.php.net/manual/en/function.random-bytes.php): 
"Generates an arbitrary length string of cryptographic random bytes that are suitable for cryptographic use, such as when generating salts, keys or initialization vectors."

messages/php/src/Id/FallbackUuidGenerator.php
ciaranmcnulty: I literally stole the code with permission so wanted to link back

maybe I should co-authored-by instead? link to spec is a good idea 👍

messages/php/src/Id/FallbackUuidGenerator.php
ciaranmcnulty: It's slightly platform dependent bit is meant to be "for reading CS random bytes from the system CSPRNG."

The original RFC is here https://wiki.php.net/rfc/csrandombytes

messages/php/src/Id/FallbackUuidGenerator.php
ciaranmcnulty: Yes you can inspect PHP_INT_SIZE (and it'll be 32 or 64 unless it's some crazy exotic platform) but that complicates the algo

there are BigInt extensions that you can use on a project, but for a library you can't really require too many extensions as they aren't pulled in by package management

I'll re-address this part of the code and see if it can be simplified

messages/php/src/Id/FallbackUuidGenerator.php
ciaranmcnulty: So TL;DR it could be poor quality if the system doesn't have a high quality random source, or if the person compiling the PHP runtime somehow disabled the available sources.

messages/php/src/Id/FallbackUuidGenerator.php
ciaranmcnulty: True maybe could just replace

messages/php/src/Id/FallbackUuidGenerator.php
ciaranmcnulty: I'll just move this to the commit message ;)

ciaranmcnulty: It's drifted enough that we don't need to credit; removed

messages/php/src/Id/FallbackUuidGenerator.php
ciaranmcnulty: I've been doing PHP for over 20 years and found out today you can use bitwise operators directly on strings 🤯

Dynamic typing rocks

I've reworked this bit but GH has lost the ref because the filename changed

messages/php/src/Id/FallbackUuidGenerator.php
ciaranmcnulty: No longer falls back

messages/php/src/Id/UuidIdGenerator.php
```diff
@@ -4,17 +4,41 @@
...
+         */
+        $bytes[6] = $bytes[6]
+            & "\x0F" // 0 the first 4 bits
+            | "\x40" // raise 2nd bit to 1
+        ;
+
+        return preg_replace(
+            '/^(.{8})(.{4})(.{4})(.{4})(.{12})$/',
+            '\\1-\\2-\\3-\\4-\\5',
+            bin2hex($bytes),
```
mpkorstanje, line 41: Might be faster to join substrings instead of a regex.

ciaranmcnulty: I guess I can benchmark 👍  Regex are  often unreasonably fast in PHP

ciaranmcnulty: Regex 2X faster https://3v4l.org/L6Hqb

ciaranmcnulty: Er, no significant diff

https://3v4l.org/f82j9

(ignore deleted comment)

mpkorstanje: That's because you've put `$bytes = bin2hex(random_bytes(16));` inside the loop. 

If you move that out, regex is actually faster.

mpkorstanje: Very surprising but looks like the implementation of `preg_replace` can more efficiently build the string then php can.

ciaranmcnulty: Now I need to google if I can bin2hex as part of the regex!

ciaranmcnulty: Or maybe stop microoptimising :)

mpkorstanje: It's worth spending some time on it, it gets called quite often.

mpkorstanje: I do think the current performance is reasonable enough though.

messages/php/src/Id/FallbackUuidGenerator.php
mpkorstanje: Since Cucumber is a testing framework, I think we're okay then. At worst the messages output will be unusable then.

mpkorstanje (approved):

messages/php/src/Id/UuidIdGenerator.php
ciaranmcnulty: At some point I'll get some proper benchmarking and see where bottlenecks are (separately)

PR #1927, opened 2022-03-21
Remove message-streams from the monorepo

aurelien-reeves:
## Summary

`message-streams` now lives in its own repo here: https://github.com/cucumber/message-streams/

## Motivation and Context

Refs. #1724 

## How Has This Been Tested?

Thanks to the CI


aurelien-reeves:
Reviewers: I would like to draw your attention on the gherkin changelog, as there may be kinda undesired changed with `gherkin-streams`

davidjgoss (approved):
> Reviewers: I would like to draw your attention on the gherkin changelog, as there may be kinda undesired changed with gherkin-streams

This seems like the right change to me. `gherkin-streams` lists `message-streams` as a peer dependency - although I do think we should add an `Installation` section to the readme there and explicitly call out the peer dependencies.

Also I think the plan is for gherkin to be decoupled from messages, right? So this should be short-lived anyway.

aurelien-reeves:
@davidjgoss something like that: https://github.com/cucumber/gherkin-streams/pull/12?

PR #1952, opened 2022-04-14
[messages - elixir] Convert proto-based messages to json schema based messages

WannesFransen1994:
## Summary

Remove protobuf dependency and create modules based on json schema (manual parser). No validation of data in messages, that's done through acceptance testing.

## Checklist:

- [x] Clean dependencies
- [x] Generate modules in generated folder
- [x] Adjust Make file & do integrations in CICD
- [x] See if compatibility testing kit is relevant
- [x] Adjust gherkin library as well and verify with acceptance testing there
- [x] Provide meaningful debug messages if applicable


WannesFransen1994:
Attempted a makefile adjustment, but seems that there's a problem with the messages generation. (based on the log)
```
==> cucumber_messages
Compiling 3 files (.ex)
Generated cucumber_messages app

```

WannesFransen1994:
Not that familiar with the whole monorepo build, but I've managed to make it work with adjusting the Makefile. Would love some pointers / reviews on the integration, specifically for
1. [manually copying the json schema files](https://github.com/cucumber/common/blob/176bd7e0989283fa098a296c000bc788c4245108/messages/elixir/Makefile#L11)
2. [instead of adjusting the general gitignore - which caused an rsync conflict-, I've adjusted an ignore file here](https://github.com/cucumber/common/blob/176bd7e0989283fa098a296c000bc788c4245108/messages/elixir/priv/json_schemas/.gitignore#L3)
3. The CircleCI config integration file
4. Instead of running the compiled binary with every file and verifying the output, I do this in the [test suite](https://github.com/cucumber/common/blob/176bd7e0989283fa098a296c000bc788c4245108/gherkin/elixir/test/cucumber_gherkin_good_testdata_test.exs#L7) (since starting an beam VM every time is expensive). Saves a lot of time though (would be even more if we ran this in parallel, though not sure it is worth the effort)

aurelien-reeves (changes requested):
We pair today with @WannesFransen1994 to review the PR.
We noticed the following:

- [ ] Please join the generated code for messages as part of the repo
- [ ] We usually use the CCK as part of the messages high-level validation to make sure they serialize and deserialize as expected (you can take inspiration from ruby, perl, php, java or js implementation for that)
- [ ] If you want to comply with request from Erik, you can copy the good/bad test data using make but keeping those ignored in the repo

Beside that: great job @WannesFransen1994! 👍Thanks a lot!

aurelien-reeves:
PR now available in the dedicated repo here: https://github.com/cucumber/messages/pull/29

PR #1959, opened 2022-04-19
Gherkin java codegen

### LINKED ISSUE #1957: gherkin/java: Use generated code instead of reading gherkin-languages.json
### 🤔 What's the problem you're trying to solve?

Reading `gherkin-languages.json` requires the inclusion of a json library. To avoid poluting the class path this library has to be shaded. This makes in place upgrades (e.g. for security reasons) harder.

### ✨ What's your proposed solution?

Instead of reading `gherkin-languages.json` generate the code for a list of `GherkinDialect` objects. See for example https://github.com/cucumber/ci-environment/blob/cfedf95c90052b2db812a56851daf1a6f7d1f690/java/src/main/groovy/io/cucumber/cienvironment/CiEnvironments.gsp#L7-L13

aslakhellesoy:
## Summary

Generate GherkinDialects.java at build time, using Groovy as a template engine and the root `gherkin-languages.json` as data for the template. This fixes #1957.

## Details

With this change the library no longer needs to parse JSON at runtime. This makes the library:

* more performant (not really important)
* smaller memory footprint (not really important)
* more secure and less maintenance needed, because 3rd party JSON parsers tend to have CVEs

With this change there is no need to copy the `gherkin-languages.json`  file either.

## Motivation and Context

See #1957

## Types of changes

- [x] Breaking change.

## Checklist:


gherkin/java/src/main/groovy/io/cucumber/gherkin/GherkinDialects.gsp
```diff
@@ -0,0 +1,32 @@
...
+
+import static java.util.Arrays.asList;
+import static java.util.Collections.unmodifiableList;
+
+final class GherkinDialects {
+    static final Map<String, GherkinDialect> DIALECTS = new HashMap<>();
+
+    static {
+<% dialects.eachWithIndex {name, dialect, index -> %>
+        DIALECTS.put("<%= name %>", new GherkinDialect(
```
mpkorstanje, line 14: The map should be unmodifiable too. 

```
Map<String, GherkinDialect> dialects  = new HashMap<>();

...

DIALECTS = unmodifyableMap(dialects);
```

gherkin/java/src/main/java/io/cucumber/gherkin/GherkinDialectProvider.java
```diff
@@ -44,18 +24,16 @@ public GherkinDialect getDefaultDialect() {
 
     public GherkinDialect getDialect(String language, Location location) {
         requireNonNull(language);
+        GherkinDialect dialect = GherkinDialects.DIALECTS.get(language);
 
-        JsonValue languageObject = DIALECTS.get(language);
-        if (languageObject == null) {
+        if (dialect == null) {
             throw new ParserException.NoSuchLanguageException(language, location);
```
mpkorstanje, line 30: Would be nice if we could throw this parser exception elsewhere and have this method return `Optional<GherkinDialect>`.

gherkin/java/src/main/java/io/cucumber/gherkin/GherkinDialectProvider.java
```diff
@@ -44,18 +24,16 @@ public GherkinDialect getDefaultDialect() {
...
 
-        return new GherkinDialect(language, languageObject.asObject());
+        return dialect;
     }
 
     public List<String> getLanguages() {
-        List<String> languages = new ArrayList<>(DIALECTS.names());
-        sort(languages);
-        return unmodifiableList(languages);
+        return unmodifiableList(asList("en", "no"));
```
mpkorstanje, line 37: Should this be `GherkinDialects.DIALECTS.keySet()`?

gherkin/java/src/main/java/io/cucumber/gherkin/GherkinDialect.java
```diff
@@ -1,95 +1,109 @@
...
-        List<String> result = new ArrayList<>();
-        for (JsonValue jsonValue : array) {
-            result.add(jsonValue.asString());
-        }
-        return result;
+        return featureKeywords;
     }
 
     public String getName() {
-        return keywords.getString("name", null);
```
mpkorstanje, line 34: Was this field nullable?

gherkin/java/src/main/java/io/cucumber/gherkin/GherkinDialect.java
```diff
@@ -1,95 +1,109 @@
...
+        return name;
     }
 
     public String getNativeName() {
-        return keywords.getString("native", null);
```
mpkorstanje, line 38: Was this field nullable?

gherkin/java/src/main/java/io/cucumber/gherkin/GherkinDialectProvider.java
mpkorstanje: And if we're breaking things, may well change the return type to a set.

gherkin/java/src/main/groovy/io/cucumber/gherkin/GherkinDialects.gsp
```diff
@@ -0,0 +1,32 @@
+package io.cucumber.gherkin;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static java.util.Arrays.asList;
+import static java.util.Collections.unmodifiableList;
+
+final class GherkinDialects {
+    static final Map<String, GherkinDialect> DIALECTS = new HashMap<>();
```
mpkorstanje, line 10: Consider a `LinkedHashSet` to preserve the order from `gherkin-languages.json`. This may be useful for `GherkinDialectProvider.getLanguages()`

gherkin/java/src/main/java/io/cucumber/gherkin/GherkinDialectProvider.java
aslakhellesoy: Ah yes - leftover from the first increment

gherkin/java/src/main/java/io/cucumber/gherkin/GherkinDialect.java
aslakhellesoy: No, all the entries in the JSON file have a `name` property.

gherkin/java/src/main/java/io/cucumber/gherkin/GherkinDialect.java
aslakhellesoy: No, same as above

gherkin/java/src/main/groovy/io/cucumber/gherkin/GherkinDialects.gsp
aslakhellesoy: Ok

gherkin/java/src/main/java/io/cucumber/gherkin/GherkinDialectProvider.java
aslakhellesoy: Ok

gherkin/java/src/main/groovy/io/cucumber/gherkin/GherkinDialects.gsp
aslakhellesoy: Ok

aslakhellesoy:
Thanks for the feedback @mpkorstanje - fixed now I think. Good to merge?

gherkin/java/src/main/java/io/cucumber/gherkin/GherkinDialectProvider.java
```diff
@@ -39,23 +19,15 @@ public GherkinDialectProvider() {
...
-        return new GherkinDialect(language, languageObject.asObject());
+        return Optional.ofNullable(GherkinDialects.DIALECTS.get(language));
     }
 
-    public List<String> getLanguages() {
-        List<String> languages = new ArrayList<>(DIALECTS.names());
-        sort(languages);
-        return unmodifiableList(languages);
+    public Set<String> getLanguages() {
+        return unmodifiableSet(GherkinDialects.DIALECTS.keySet());
```
mpkorstanje, line 31: Already unmodifiable because `DIALECTS` is.

mpkorstanje (approved):
LGTM. The one nit pick aside.

mpkorstanje:
Cheers btw.

ciaranmcnulty:
Looks good 👍

PR #1984, opened 2022-05-18
CCK: Add implementation of named hook in ruby

aurelien-reeves:
## Summary

About to release cucumber-ruby 8.0.0.
After upgrading all the gems, cucumber-ruby does not pass the CCK due to missing named hooks support.
It is actually pretty easy to add, so, let's add it in the CCK first, then on cucumber-ruby.

Here, we just need to add a name to the "Before" hook in the ruby step definitions

### DIFF
--- a/compatibility-kit/CHANGELOG.md
+++ b/compatibility-kit/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+* [Ruby] Add implementation for named hooks
+
 ## [9.2.0] - 2022-04-01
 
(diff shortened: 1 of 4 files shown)


luke-hill:
No no it's fine. I just couldn't remember the discussion / why. Definitely not against this.

aurelien-reeves:
TBH there have not been any discussion regarding supporting those named hooks in cucumber-ruby. I was about to release cucumber-ruby, and noticed a cck failed test after upgrading the gems (gherkin, messages, the cck). So I worked on adding that support  in cucumber-ruby

But of course, that would have passed through a PR, then the discussion would have started there :)

PR #1725, opened 2021-09-03
Gherkin-Utils - Java - Address issue #1662 - add a Java pretty print gherkin module

nddipiazza:
## Summary

Ported the TypeScript Gherkin pretty print utility to Java.

## Details

Make a java port of the javascript pretty printing library available in gherkin-utils.

## Motivation and Context

A huge amount of the Gherkin users out there are java users through cucumber-jvm project. We need a pretty printer for Gherkin.

## How Has This Been Tested?

Took most of the unit tests from the typescript project.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue).
- [x] New feature (non-breaking change which adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).

## Checklist:

- [x] The change has been ported to Java.
- [x] I've added tests for my code.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have updated the CHANGELOG accordingly.

PR #1985, opened 2022-05-18
Nddipiazza address issue 1662

aslakhellesoy:
This supersedes #1725


jamietanna (changes requested):
A few features that may need output tweaking.

<details>

<summary><code>examples_comment.feature</code></summary>

Before:

```gherkin
Feature: Comment before an Examples

  Scenario Outline: with examples
    Given the <value> minimalism
    # then something happens

    Examples:
      | value |
      | 1     |
      | 2     |
```

After:

```gherkin
Feature: Comment before an Examples

  Scenario Outline: with examples
    Given the <value> minimalism

    # then something happens
    Examples:
      | value |
      | 1     |
      | 2     |
```

</details>


<details>

<summary></summary>

Before:

```gherkin
# This is a feature file for a new thing that we're delivering, and may not quite work yet
Feature: Some feature

  Scenario: The thing
    # TODO: add in given
    When something happens
    Then an HTTP 200 is returned
    # TODO validate more!
```

After:

```gherkin
# This is a feature file for a new thing that we're delivering, and may not quite work yet
Feature: Some feature

  Scenario: The thing
    # TODO: add in given
    When something happens
    Then an HTTP 200 is returned
# TODO validate more!
```

</details>

I unfortunately no longer have access to some of the more complex Gherkin features that I've used in the past, as they have some interesting uses of comments, but hopefully with these we can see some areas to tweak.

PR #1998, opened 2022-06-01
[Perl] Align Makefile and default.mk more with Ruby

ehuelsmann:
By using the same build targets, where possible, and the same
division of concerns between 'pre-release', 'test', 'clean' and
'publish', the release process be more comprehensible for non-Perl
contributors.

### DIFF
--- a/.templates/perl/default.mk
+++ b/.templates/perl/default.mk
@@ -3,7 +3,11 @@
 #     source scripts/functions.sh && rsync_files
 #
 SHELL := /usr/bin/env bash
-ALPINE := $(shell which apk 2> /dev/null)
+PERL_SOURCE_FILES = $(shell find lib -name '*.pm') $(shell test -d bin && find bin -type f) $(shell find t -name '*.t')
+LIBNAME := $(shell head -10 dist.ini | grep 'name.*=' | sed -e 's/name.*= *//')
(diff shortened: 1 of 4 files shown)


.templates/perl/default.mk
```diff
@@ -42,28 +68,31 @@ else
...
-.PHONY: predistribution
 
-pre-release: update-version
-.PHONY: pre-release
+publish: dist
+ifeq ($(IS_TESTDATA),-testdata)
+	# no-op
+else
+ifneq (,$(DISTINI))
+	cpan-upload $(DIST)
```
aurelien-reeves, line 77: That requires using the new version of the build image?

If so, maybe we should use that new version as part of that PR?

aurelien-reeves (commented):
As far as I can tell, it looks good

We just have to make sure to use the proper build image

.templates/perl/default.mk
aurelien-reeves: (FYI: I've just released a new version of the cucumber build image: 0.13.0)

ehuelsmann:
> That requires using the new version of the build image?
>
> If so, maybe we should use that new version as part of that PR?
>

Exactly! And seems like the right thing to do, to include the upgrade of
the image, indeed.

>

PR #2001, opened 2022-06-08
Perl less dzil

aurelien-reeves:
Fix the issue with  `make clean` with perl I guess?
@ehuelsmann ^


ehuelsmann:
To answer your question from Slack: yes, this fixes the Perl "make clean" issue.

PR #2010, opened 2022-06-17
[Messages][JavaScript] Add schemas to npm package

aurelien-reeves:
## Summary

The dedicated repo for the CCK requires access to messages json schemas. An easy way to do so is to publish the schemas with the messages NPM package.

Actually that may be a nice addition to the package itself even beside our own primary need. We may later consider publishing those schemas with other implementations of messages.

## Details

We copy the schemas as part of the `.codegen` make target


mattwynne:
To release this, we still need to modify `package.json` to include these files in the published package.

mattwynne:
This is released in [messages/v19.1.0](https://github.com/cucumber/common/releases/tag/messages%2Fv19.1.0)

aurelien-reeves:
> To release this, we still need to modify `package.json` to include these files in the published package.

We actually did not do that part 😶

ciaranmcnulty:
Late to this but FWIW why not have the schemas in their own subrepo with their own package.json? That way if I wanted them in PHP I could add a composer.json... etc

aurelien-reeves:
@ciaranmcnulty that was on the table indeed (refs. https://github.com/cucumber/compatibility-kit/issues/5)

We've picked the option to embed those as part of the @cucumber/messages npm package in order to move fast with this and to comply with the original need easily and quickly.

Be we agree that:
- having those schemas as part of all messages implementation
- or having dedicated packages for the schema in different implementations

could be nice things to have "later".

And as I write that comment, I realize that those schemas could actually also be part of the CCK 🤔 .

Do you know that we have a CCK for javascript and for ruby, and that we could also have a CCK for php, or any other languages? ☺️

mattwynne:
> And as I write that comment, I realize that those schemas could actually also be part of the CCK 🤔 .

As I see it, there are three levels of abstraction that we could express in the dependencies:

1. The schemas themselves - an abstract representation of the data structures we use for the message protocol
2. The message DTOs - packages that provide concrete implementations of those data structures in a given programming language
3. The CCK - a tool that uses those DTOs and schemas to validate the behaviour of Cucumber implementations against our expectations of the message protocol.

I see the CCK as something analogous to a [contract test](https://blog.thecodewhisperer.com/permalink/getting-started-with-contract-tests) for the message protocol in that it validates an implementation of the message protocol not just in terms of message structure, but message content and sequence, in different scenarios.

Whether we need to separate the schemas out to another package now or in the long term to better represent these abstractions, I'm not sure.

aurelien-reeves:
> > > To release this, we still need to modify `package.json` to include these files in the published package.
> > 
> > 
> > We actually did not do that part 😶
> 
> Gah! So we need a 19.1.1 release to fix that?

Fixed in 19.1.2

19.1.1 did not worked as expected either

PR #2014, opened 2022-06-20
Gherkin PHP - Fix messages constraint

ciaranmcnulty:
The post-release script must have an error in it :/ it's only meant to add new minor versions


ciaranmcnulty:
https://github.com/cucumber/common/blob/main/messages/php/scripts/update-gherkin-dependency.php#L45 this is just dumb

PR #2027, opened 2022-06-28
Remove ./compatibility-kit/

aurelien-reeves:
Refs. [github.com/](https://github.com/cucumber/compatibility-kit/)
Refs. #2005

PR #2028, opened 2022-06-28
Remove json-to-messages

aurelien-reeves:
json-to-messages acceptance testing was broken since a long time without anyone noticed.

It seems is it not used that much.

We decided at a community meeting (2022-0616) to remove it:

> - Monorepo exit
>   - [Matt / Aurellien] Do we still need json-to-messages?
>     - Decision: pull it out and archive it but not fix CI etc.

It has been exfiltrated and archived here: https://github.com/cucumber/json-to-messages


aurelien-reeves:
Merge https://github.com/cucumber/common/pull/2027 first

PR #2033, opened 2022-07-01
messages: extend javascript message type with workerId

lamartire:
## Summary

Provide optional `workerId` property to `Message` type

## Details

`workerId` property, which can be useful for some cases, when you need to capture additional testing details. The changes required only for `javascript` message type because most of another platforms don't have problem with thread ID identifying.

Then I want to provide additional PR for `cucumber-js` where I plan to use the property.

## Motivation and Context

`allure-cucumberjs` needs to have access to thread or worker id for better reports.

## Types of changes

## Checklist:

### DIFF
--- a/messages/CHANGELOG.md
+++ b/messages/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+* [Javascript] Add `workerId` to `Message` type
+  ([PR#2033](https://github.com/cucumber/common/pull/2033))
+
 ### Changed
 
 ### Deprecated
--- a/messages/javascript/src/messages.ts
+++ b/messages/javascript/src/messages.ts
@@ -80,6 +80,9 @@ export class Envelope {
 
   @Type(() => UndefinedParameterType)
   undefinedParameterType?: UndefinedParameterType
+
+  @Type(() => String)
+  workerId?: string
 }
 
 export class GherkinDocument {


baev:
In Allure we have a Timeline graph that visualises how tests were executed across hosts and threads:

Screenshot 2022-07-01 at 21 20 57

So we are looking for ways to implement it for cucumber-js since it now has parallel mode.

lamartire:
I think, we can extend the type right in the `cucumber-js` repo. It that way, we'll keep original interface and provide what we need in the library

lamartire:
Like this:

```js
Message & { workerId?: string }
```

aurelien-reeves:
> I think, we can extend the type right in the `cucumber-js` repo. It that way, we'll keep original interface and provide what we need in the library

I think so, yes 👌

Thanks for your understanding 🙂

How do you integrate cucumber-js with allure? Do you use a custom formatter?

lamartire:
> Ok, thanks
> 
> Adding the workerId as part of the formatter is not possible? (this is a legit question!)

It would be possible only if the messages contain `workerId`. At this moment we don't have access to any worker env-variables then, we can't know which worker it was.

I opened a PR to `cucumber-js` and it adds the logic with batch of documentation:

```javascript
const { Formatter } = require('@cucumber/cucumber')
class ExampleFormatter extends Formatter {
  constructor(options) {
		options.eventBroadcaster.on('envelope', (envelope) => {
			if (envelope.workerId) {
        console.log(`the envelope was sent from a worker with id ${envelope.workerId}`)
      } else {
        console.log('the envelope was sent outside of any worker')
      }
		})
    super(options)
  }
}
module.exports = ExampleFormatter
```

PR #2034, opened 2022-07-03
PHP Gherkin Parser: v24 is not compatible with v18 of messages

dantleech:
## Summary

Updated `composer.json` to require at least v19 of cubumber PHP messages

## Details

When using v24 with lowest dependencies:

```
cucumber/gherkin                      v24.0.0     Gherkin parser
cucumber/messages                     v18.0.0     JSON schema-based messages for Cucumber's inter-process communication
```

We get the following error:

```
PHP Fatal error:  Uncaught Error: Class "Cucumber\Messages\Step\KeywordType" not found in /home/daniel/www/dantleech/gherkin-lint/vendor/cucumber/gherkin/src/GherkinParser.php:38
```

With v19.0.0 of cucumber/messages it works.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue).
- [ ] New feature (non-breaking change which adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).

PR #2035, opened 2022-07-04
PHP: Check lowest versions of dependencies in CI

ciaranmcnulty:
This is mostly because of #2034 - we need to check that everything still works with the 'lowest' versions of the dependencies that are allowed by composer.json

There aren't any external dependencies yet so currently this will only detect when Gherkin doesn't work with older versions of Messages any more, or when we've written tests that don't work with older versions of the test tools

This is expected to fail until #2034 is merged

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -479,6 +479,7 @@ jobs:
           name: gherkin/php
           command: |
             cd gherkin/php
+            composer config --unset repo.0
             make
 
   gherkin-php-bc:
@@ -492,6 +493,7 @@ jobs:
           name: gherkin/php
           command: |
             cd gherkin/php
+            composer config --unset repo.0
             make
 
 ###


aurelien-reeves:
Do you confirm that as long as #2034 is not merged, the job `gherkin-php-bc` should fail?

ciaranmcnulty:
ok I need help with make :/

I have this in the Makefile:
```
composer update ${COMPOSER_FLAGS:-}
```

the new jobs are running:
```
composer update
```

aurelien-reeves:
(I copy/paste my answer on slack)

It looks like using variables in a makefile is not the same as with bash

refs. https://www.gnu.org/software/make/manual/html_node/Using-Variables.html

So, two possibilities:

using ifdef as we do in gherkin/go/Makefile (or its counterpart ifndef like in gherkin/javascript/Makefile)

or using $(VAR) instead of ${VAR} but here I don't know how to process unset variables

aurelien-reeves:
Actually you should be able the braces too (`${}`). I think here what is not working is the `:-` that is a bash feature and seems not working with make

ciaranmcnulty:
> Do you confirm that as long as https://github.com/cucumber/common/pull/2034 is not merged, the job gherkin-php-bc should fail?

Yes. I suggest we merge that one (it looks fine to me) and then rebase this PR

ciaranmcnulty:
Grargh

Ok this doesn't work because we point to the 'local' messages with some config in composer.json, so `--prefer-lowest` hasn't pulled in the older package

It's probably still worth merging for the other minor fixes but I'll try and see a clean way to remove the config (in just that job)

ciaranmcnulty:
ok it's 'successfully' failing now

aurelien-reeves:
I guess this should now pass!?

Once everything is ready, let me know for a last review :)

ciaranmcnulty:
Yes I think it can merge if passed (it failed on gherkin-php-bc just before rebase)

.circleci/config.yml
```diff
@@ -466,6 +479,21 @@ jobs:
           name: gherkin/php
           command: |
             cd gherkin/php
+            composer config --unset repo.0
```
aurelien-reeves, line 482: Could you add some comments here to explain why we need to do this?

PR #2095, opened 2022-10-10
gherkin/php: Fix psalm issues

mpkorstanje:
## Summary

Running `make` for `gherkin/php` inside the docker container results in complaints about:

```
$cd gherkin/php
$ make
...
vendor/bin/php-cs-fixer --dry-run --diff fix
Loaded config default from "/app/gherkin/php/.php-cs-fixer.php".
Using cache file ".php-cs-fixer.cache".

Checked all files in 0.018 seconds, 14.000 MB memory used
vendor/bin/psalm --no-cache
Target PHP version: 8.1 (inferred from composer.json)
Scanning files...
Analyzing files...

░░░░░░░░░░░░░░░░░░░░░░░░E░░░░░░░░░░░░░░░░░░░░░░░░░░░

ERROR: InvalidReturnType - src/AstNode.php:56:22 - The declared return type '(S:fn-cucumber\gherkin\astnode::getsingle as object)|null' for Cucumber\Gherkin\AstNode::getSingle is incorrect, got '(S:fn-cucumber\gherkin\astnode::getsingle as object)|(TGeneratedFromParam2:fn-cucumber\gherkin\astnode::getsingle as null|object)' (see https://psalm.dev/011)
     * @psalm-return ($defaultValue is null ? S|null : S )


ERROR: InvalidReturnStatement - src/AstNode.php:62:16 - The inferred type '(S:fn-cucumber\gherkin\astnode::getsingle as object)|(TGeneratedFromParam2:fn-cucumber\gherkin\astnode::getsingle as null|object)' does not match the declared return type '(S:fn-cucumber\gherkin\astnode::getsingle as object)|null' for Cucumber\Gherkin\AstNode::getSingle (see https://psalm.dev/128)
        return $items[0] ?? $defaultValue;


```


mpkorstanje:
@ciaranmcnulty could you try running `vendor/bin/php-cs-fixer --dry-run --diff fix` from inside the docker container? I guess there is a difference between CI but I can't immediately spot it.

gherkin/php/src/AstNode.php
```diff
@@ -62,8 +62,14 @@ public function getSingle(string $expectedType, RuleType $ruleType, ?object $def
         return $items[0] ?? $defaultValue;
     }
 
-    /** needed for non-object return */
-    public function getSingleUntyped(RuleType $ruleType, mixed $defaultValue = null): mixed
+    /**
+     * needed for non-object return
+     *
+     * @param null|object $defaultValue
```
ciaranmcnulty, line 68: This isn't correct, this method is here because not everything in PHP is an object (it's the non-object version of getSingle) so `mixed` is the correct type here

Additionally the use of `S` as a template below isn't meaningful as `S` isn't defined anywhere

I think this whole docblock can be reverted as it's not causing the issue

ciaranmcnulty (changes requested):
I think e should the changes to `getSingleUntyped`

ciaranmcnulty:
I've pushed `abaa3db` to show how to silence these errors - the types look correct to me and the fact it's only showing in some environments makes me think it's a bug somewhere

I've raised https://github.com/vimeo/psalm/issues/8572 and reached out to the psalm devs

TBH probably the easiest route forwards is to apply `abaa3db` directly onto `main` in a new branch

